PR #1, opened 2016-03-03
Sighup fix

lijog:


krader1961:
Besides the obvious style issues (e.g., the extraneous parens in statements like `return (0);`) this looks like a reasonable change that probably fixes the issue without introducing a new bug. Or at least a new bug no more serious than the one it is trying to fix. I'm wondering why it hasn't been merged or rejected.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -1209,6 +1209,61 @@ int job_kill(register struct process *pw,register int sig)
 }
 
 /*
+ * Similar to job_kill, but dedicated to SIGHUP handling when session is
+ * being disconnected.
+ */
+int
+job_hup(struct process *pw, int sig)
```
siteshwar: We can remove the ```sig``` argument if it's dedicated to ```SIGHUP``` handling.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -1209,6 +1209,61 @@ int job_kill(register struct process *pw,register int sig)
...
+	if(pw->p_cojob) {
+		job_lock();
+		r = cokill(pw->p_cojob->coshell,pw->p_cojob,sig);
+		job_unlock();
+		return (r);
+	}
+#endif /* SHOPT_COSHELL */
+
+	job_lock();
+	if (pw->p_pgrp != 0) {
```
siteshwar: This has been tested above. This condition can not be ```false```.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -1209,6 +1209,61 @@ int job_kill(register struct process *pw,register int sig)
...
+		 * that the p_pgrp is still the valid process group that we
+		 * made, ie pid may have been recycled and same p_pgrp may
+		 * have been assigned to unrelated processes.
+		 */
+		if (palive) {
+			if (killpg(pw->p_pgrp, SIGHUP) >= 0)
+				job_unstop(pw);
+		}
+	}
+	for (; pw != NULL && pw->p_pgrp == 0; pw = pw->p_nxtproc) {
```
siteshwar: This loop will never execute as we return from this function if ``` pw->p_pgrp == 0``` above.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -1209,6 +1209,61 @@ int job_kill(register struct process *pw,register int sig)
...
+ * being disconnected.
+ */
+int
+job_hup(struct process *pw, int sig)
+{
+	struct process	*px;
+	pid_t	pid;
+	int	r;
+
+	if (pw->p_pgrp == 0 || (pw->p_flag & P_DISOWN))
```
siteshwar: I am curious how ```SIGHUP``` will be handled for processes in current process group if we return from here.

krader1961:
After further reflection I recommend rejecting this change. The fact that the current code is sending a signal to a process that is no longer a child process, other than in response to an explicit `kill` command, implies a serious bug in job management. It is not at all obvious this change does anything more than paper over that bug.

krader1961:
FWIW, This change is from https://github.com/oracle/solaris-userland/blob/master/components/ksh93/patches/260-22964338.patch.

siteshwar:
For now I am closing this pull request. As @krader1961 mentioned we need to have a deeper look at the job management code.

PR #18, opened 2016-05-26
Add class alias for typeset -T

dannyweldon:
I think this would be useful as the word class is much more intuitive and elegant for defining types.

### DIFF
--- a/src/cmd/ksh93/data/aliases.c
+++ b/src/cmd/ksh93/data/aliases.c
@@ -34,6 +34,7 @@ const struct shtable2 shtab_aliases[] =
 #endif /* SHOPT_FS_3D */
 	"autoload",	NV_NOFREE,		"typeset -fu",
 	"bool",		NV_NOFREE|BLT_DCL,	"_Bool",
+	"class",	NV_NOFREE|BLT_DCL,	"typeset -T",
 	"command",	NV_NOFREE,		"command ",
 	"compound",	NV_NOFREE|BLT_DCL,	"typeset -C",
 	"fc",		NV_NOFREE,		"hist",


siteshwar:
As I understand, this introduces new alias in ksh. I might consider this patch for development branch, but it won't be a good idea to add it in the beta release.

dannyweldon:
> As I understand, this introduces new alias in ksh. I might consider this patch for development branch, but it won't be a good idea to add it in the beta release.

Yes, agreed.
Plus, if accepted into dev, I should add something to the man page as well.

krader1961:
FWIW, This is the type of change I prefer to see discussed in an issue before a PR is created since it is likely to be controversial. I'm not opposed to the enhancement since we already have aliases like `compound` for `typeset -C`. But it isn't obvious to me that the new alias is sufficiently useful to warrant usurping another word.

Also, I don't see a "development" branch in existence and there isn't a *CONTRIBUTING.md* file or similar document to explain the how this project evolves, @siteshwar, so your comment is slightly baffling to someone new to the project. :smile:

siteshwar:
I am thinking if we should merge current beta branch into master and do the development in master. Given the current status of beta branch, I don't see a release for it in near future.

krader1961:
> I am thinking if we should merge current beta branch into master ...

I am too new to this project to have a strong opinion on such matters. Some projects use the "master" branch for bleeding edge development and cherry-pick changes from it into other branches for specific purposes such as a new release. But that is by no means the only change management model. The situation is complicated when you consider how to handle a new minor release at the same time you are working on a major release.

Whatever principles this project adopts with respect to release management, and associated handling of pull-requests, it needs to be documented. Because it is not clear to me why the "alpha" or "beta" branches exist. Or why there are three pull-requests more than a year old. Note that this is the condition I found [fish-shell](https://github.com/fish-shell/fish-shell) in when I started contributing in October 2015. So, this project is not at all unusual in this respect. But keeping ksh a viable alternative to bash, zsh, etc., is going to require active management. Including timely handling of pull-requests.

The alternative is to freeze the implementation (no new features) and only accept bug fixes.

dannyweldon:
Regarding adding the "class" alias, I have quite a few other ideas, which I will raise in a separate issue.  But if any of these new ideas or keywords are considered worth pursuing after being discussed, I recommend putting in a note into the man page straight away that these keywords should be avoided as they may become ksh reserved words in the future.  Of course, I still think the "class" alias is a great idea, but people need to be given advanced warning, and perhaps we can enable some features with a new set option, so any existing scripts don't break, or even add a warning for the use of a potential future reserved word.  I vaguely recall that ksh now does give warnings about some things.

I should point out that the changes I am thinking about should not affect POSIX compatibility and are nothing like that other shell starting with 'z'.  But I'll leave that for the other issue.

@siteshwar  You may cancel this PR if you like, as it is trivial to recreate, plus it would also need a man page update,, unless you want to keep it open to continue discussions on the git work flow.

I think we could tag the current master branch to be 2008-08-11 and merge the beta branch once we are sure it compiles on OSX and BSD, and tag that with 2014-12-24.  Or we could update this file to a more recent date and tag the commit to master accordingly:

    ksh93/include/version.h

Unfortunately, it predates semver, but I did some investigation a while back and I think that they have followed this sort of convention:

    2012-08-01
    2012-08-01.alpha
    2012-08-01.beta
    2012-08-01.rc1

krader1961:
> perhaps we can enable some features with a new set option

/me runs away 😸 Just kidding. I do, however, agree quite strongly with the core guideline of the fish-shell project: options that configure the behavior of the shell is the root of all evil. It's why [I gave up using zsh](http://www.skepticism.us/2015/10/its-time-to-replace-zsh-with-a-saner-shell-because-unsetopt-multifuncdef-breaks-tab-completion/). If we had something like the concept of python modules so that we could implement the python equivalent of `from __future__ import print_function` to enable the `print()` builtin function only for the module containing that import such configurable behavior would be okay. But as long as the options are global in scope they cause more problems than they solve.

siteshwar:
As @krader1961 mentioned it would be best to discuss the reasons for this change in an issue. For now I will close this pull request.

PR #19, opened 2016-07-19
Patches to compile on FreeBSD (2017-10-14)

saper:
RELEASE notes for src/cmd/ksh93/RELEASE

```
16-07-17  --- Release ksh93v- ---
16-07-17  A fix to avoid "non-void function 'ls' should return a value" error.
16-07-17  Some patches for compilation on FreeBSD have been added.
```

Credit goes to Robert S. Krzaczek


dannyweldon:
Do we know if this still builds okay on Linux with these changes?  It looks like travis has not built it because it only builds commits made after travis has been enabled for the repo.

saper:
I have amended the commit and pushed again, let's see...

krader1961:
FYI, This plus PR #76 allows everything but `dlls` and `pax` to be built on macOS Sierra (10.12.6) I recommend merging it.

siteshwar:
I will have to clean up and rearrange the commits before I merge this one.

saper:
@siteshwar thank you, yes, this is a quick dump of what "works for me" to share, but it should be cleaned up

krader1961:
I should have looked more closely at this before I made my changes to fix building ksh93 on macOS (a BSD derivative). At this point the only part of this PR not already merged simply changes the version number. Something I don't think should be necessary. We shouldn't have to manually update the version every time a commit is merged. It's worthwhile to have the version reflect the commit the build is based on but that should be done dynamically at build time by extracting the git head commit ID or tag.

Sorry this languished so long, @saper.

saper:
I have bumped the version number to create my own ksh93 release, since the upstream seemed pretty much dead. I found it useful to add some comments in the release notes.

PR #45, opened 2017-06-30
Pass correct file descriptor to sh_iosave()

siteshwar:
Resolves: rhbz#1421650


dannyweldon:
@kdudka Can you elaborate?

kdudka:
First off, there is duplicated code: ```tname?fname:(trunc?Empty:0)``` could be at a single place.  Then both the long lines at end of those if/else blocks only differ in the ```fd``` vs. ```fn``` identifiers while the actual if condition was ```(fd == fn)```.  The only coincidence making it useful was the ```fd = r;``` assignment in a nested condition.  That took me a while to realize...

Issue #56, opened 2017-07-11
Incorrect exit message when exiting


siteshwar:
```ksh``` gives incorrect exit message if exit code is greater than 256. ksh uses ```256+signal number``` to show error message codes for signals. However below command should not give an error message :

```
$ exit 257
Hangup
```

siteshwar:
Closed by https://github.com/att/ast/pull/57

PR #57, opened 2017-07-11
Use posix exit code if last command exits due to a signal

siteshwar:
Resolves: #56

### DIFF
--- a/src/cmd/ksh93/sh/fault.c
+++ b/src/cmd/ksh93/sh/fault.c
@@ -671,7 +671,7 @@ void sh_done(void *ptr, register int sig)
 	sfsync((Sfio_t*)sfstdin);
 	sfsync((Sfio_t*)shp->outpool);
 	sfsync((Sfio_t*)sfstdout);
-	if(savxit&SH_EXITSIG)
+	if((savxit&SH_EXITMASK) == shp->lastsig)
 		sig = savxit&SH_EXITMASK;
 	if(sig)
 	{
@@ -698,6 +698,12 @@ void sh_done(void *ptr, register int sig)
 #endif /* SHOPT_KIA */
 	if(shp->pwdfd >=0)
 		close(shp->pwdfd);
+
+	/* Return POSIX exit code if last process exits due to signal */
+	if (savxit & SH_EXITSIG) {
+		exit(128 + (savxit&SH_EXITMASK));
+	}
+
 	exit(savxit&SH_EXITMASK);
 }
 


src/cmd/ksh93/sh/fault.c
```diff
@@ -698,6 +698,12 @@ void sh_done(void *ptr, register int sig)
 #endif /* SHOPT_KIA */
 	if(shp->pwdfd >=0)
 		close(shp->pwdfd);
+
+	// Return POSIX exit code if last process exits due to signal
```
kdudka: Please use ```/* ... */``` comments to stay consistent with the rest of ksh source code.

dannyweldon:
Also, it would be good to run the tests to see if the tests `tests/signal.sh` and `tests/sigchld.sh` pass:

```
bin/package use
bin/package test ksh93
```

kdudka:
@dannyweldon Those tests seem to to pass in my testing environment.

@siteshwar Please update the commit message such that it matches the actual changes in code.  This changes the way how the shell interpreter terminates in certain cases.

dannyweldon:
Hmm.  How hard would it be to get travis to do code quality checking to catch the comment style?
Though, I was thinking about a (optional) builtin that used a bit of C++ (because it calls a C++ library); would that be a no-no?
Stuff to think about.

kdudka:
@dannyweldon The C++ style comments could be detected by grep.  I will submit a PR for it as time permits...

kdudka:
@siteshwar Please update the title of this PR accordingly if you can.  I am not approved to do so.

siteshwar:
I had updated it already.

kdudka:
My mistake.  Github was fetching new comments but did not redraw the title until I hit refresh.

DavidMorano:
dannyweldon wrote:
> Though, I was thinking about a (optional) builtin that used a bit of C++ (because it calls a 
> C++ library); would that be a no-no?

I have written many built-ins that use C++.
In general there is no problem with this.
Just make sure that the shared library that the builtin command is in, has
'libstdc++' as a dependency and any other C++ exception handing libraries as may be
required.  Also, just for complete disclosure, any C++ builtins
finally (after some custom library layers) call the AST SFIO interfaces.
This is needed in order for the KSH I/O to work properly with the input and
output of the builtin command.  A construct such as:

A=$( builtin_cmd )

is an example.

I have also written many multi-threaded built-ins.
In fact since my common libraries are internally multithreaded, probably ALL of
my built-ins are really multi-threaded also. I also have threads running in the background
of the single threaded KSH itself. even when no builtins proper are being executed.
(These background threads still get |joined()| by use of the exit handlers in any 
loaded shared library objects which created them through their registered exit handlers,
which also get called on |dlclose()|.)  Why background threads?  :-) :-)
Well, the short answer is "the sky is the limit!"  Use your imagination!
One reason (among many) is that they serve as standing servers that are
also right in the same address space as KSH and any loaded built-ins
that may use them as such (through an appropriate client interface).
Another reason, and perhaps even more powerful, is that each running
KSH instance can serve requests from"outside" of the KSH, to either
receive information or to provide information to and from one or more outside "clients".
Not for the faint of heart, but all doable none-the-less!

The only (major) problem I encountered was that I could not use the AST supplied
memory allocator (VMALLOC) even though it was configured for multi-thread
safe operation.  Even ignoring background threads, the VMALLOC underneath even
just builtins alone leads to crashes in a multi-threaded environment.
I do not really know why VMALLOC is so problematic except to guess that it is
not really as multi-thread safe (when in supposed multi-thread safe mode) as it
is supposed to be.  I did not take a lot of time to figure out what was not working with
VMALLOC, so I never did find the exact reason for its failure.  I got around this problem 
by disabling the entire inclusion of VMALLOC and instead let the default system MALLOC
 interface be used (which, of course, is properly multi-threa safe).

Someday, maybe in the (long) future, every aspect of KSH and ASTLIB
and any other library which KSH calls, should be made completely multi-thread safe by
default. In my opinion, this should have been done several decades ago, perhaps at least by 
about 1990-1992 or so (about the time when the POSIX thread interface was well established
and widely implemented). In general, every program that calls a library is maybe already
secretly multithreaded without the so-called |main()| even knowing about it. 
Again, just personally, all code everywhere should have been made permanently
multi-thread safe decades ago!

kdudka:
@DavidMorano Sorry but I see no relation with this PR and your comment.  Please open a separate issue and tell us which built-ins you want to introduce and why they need to be built-ins in the first place.

dannyweldon:
@DavidMorano thanks for this tip, I came across this exact issue:

> finally (after some custom library layers) call the AST SFIO interfaces.
> This is needed in order for the KSH I/O to work properly with the input and
> output of the builtin command. A construct such as:
>
> A=$( builtin_cmd )
>
> is an example.

PR #59, opened 2017-07-14
.travis.yml: add check for line comments

kdudka:
... which are not used in ksh source code and should not be introduced
by accepted PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
     - docker pull ${OS_TYPE}
 
 script:
+    - find \( -name \*.c -o -name \*.h \) -exec sh -c '! grep --color "^[[:space:]]*//" $@' find {} \+ && echo No line comments found...
     - docker run -v $TRAVIS_BUILD_DIR:/source ${OS_TYPE} bash -c "set -e;
         ${INSTALL_REQUIREMENTS};
         cd /source;
--- a/src/cmd/builtin/builtin.c
+++ b/src/cmd/builtin/builtin.c
@@ -26,7 +26,7 @@
 
 #include <cmd.h>
 
-// added to prevent the linker from optimizing this whole file out
+/* added to prevent the linker from optimizing this whole file out */
 int _foo_bar (void) {
     sfprintf (sfstderr, "");
 }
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -14,6 +14,7 @@ before_install:
     - docker pull ${OS_TYPE}
 
 script:
+    - find \( -name \*.c -o -name \*.h \) -exec sh -c '! grep --color "^ *//" $@' find {} \+ && echo No line comments found...
```
dannyweldon: The grep won't catch comments indented by tabs because AST code uses tabs, not spaces.
Perhaps this: ! grep --color "^[[:space:]]*//" $@'
But that brings up another point.  Should we be also checking that tabs have been used?

kdudka:
On Saturday, July 15, 2017 18:59:36 Danny Weldon wrote:
> > @@ -14,6 +14,7 @@ before_install:
>      - docker pull ${OS_TYPE}
> 
>  script:
> +    - find \( -name \*.c -o -name \*.h \) -exec sh -c '! grep --color "^
> *//" $@' find {} \+ && echo No line comments found...
> 
> The grep won't catch comments indented by tabs because AST code uses tabs,
> not spaces. Perhaps this: ! grep --color "^[[:space:]]*//" $@'

Good point.  I have updated the PR per your suggestion.

> But that brings up another point.  Should we be also checking that tabs have
> been used?

There are currently 114490 lines in *.c and *.h files that begin with space.
So we would have to check only for newly introduced lines that begin with 
space.  And even in that case, some uses of them may be valid.  So in the end, 
it would mean too much work for no real benefit.

dannyweldon:
> There are currently 114490 lines in *.c and *.h files that begin with space.
> So we would have to check only for newly introduced lines that begin with
> space. And even in that case, some uses of them may be valid. So in the end,
> it would mean too much work for no real benefit.

Okay, agreed.  The only other consideration is if there is any C++ code
allowably contributed, their header files could contain C++ comments.
But we can cross that bridge if it comes.  nmake can build C++ files.

kdudka:
On Monday, July 17, 2017 02:07:47 Danny Weldon wrote:
> Okay, agreed.  The only other consideration is if there is any C++ code
> allowably contributed, their header files could contain C++ comments.
> But we can cross that bridge if it comes.  nmake can build C++ files.

It would only cause problems if the C++ code was added to *.c or *.h files, 
which is not very common.  Other file suffixes are not affected by this check.

Thanks for review!

PR #60, opened 2017-07-17
Fix crash while creating cross reference database

siteshwar:


kdudka:
I can confirm the changes make the crash go away on the following example:
```
% touch empty.sh
% bin/ksh -R xxx empty.sh
```

But I do not understand the reasoning behind the changes.  Could you please summarize?

siteshwar:
If ```shp``` variable gets initialized to ```NULL``` value then I used the same values to initalize it that were being used in older version of ```ksh``` (look into ```master``` branch). 

```if``` condition was reordered to avoid crash if ```table``` is ```NULL```.

File: src/cmd/ksh93/sh/name.c [hunk 1]
```diff
@@ -3916,7 +3920,7 @@ char *nv_name(register Namval_t *np)
 			return((*fp->disc->namef)(np,fp));
 		}
 	}
-	if(!(table=shp->last_table) || *np->nvname=='.' || table==shp->namespace || np==table)
+	if( *np->nvname=='.' || np==table || (shp && (!(table=shp->last_table) || table==shp->namespace)) )
```
kdudka: The ```shp &&``` condition seems redundant.  ```sh_getinterp()``` cannot return ```NULL``` as I understand it.

siteshwar: Right! I mixed up things while testing.

Issue #42, opened 2017-06-18
build process -- Food for thought


marcastel:
## Should we change/adapt the build process?
(with the objective to make it easily maintainable and accessible to a greater number)

### Small knowledgeable community

I know nobody amongst my fellow developers that has insight knowledge on the build process beyond `bin/package make`, let alone to maintain it, but simply to customise it. Neither do I.

That said, they know how to maintain and tweak GNU Autotools and CMake toolchains.

_Note_: Do not jump to conclusions here, I am not trying to promote Autotools or CMake; quite the contrary. I simply want to emphasise the lack of _openess_ of the build process' logic and toolchain.

### Scarse documentation

Embedded usage documentation in any C or Korn shell script is a fantastic asset of the AST developments, and certainly immensely underused among users of AST packages.... except for the AST developers who have consistently added usage information to all their utilities.

Nonetheless that documentation does not suffice for a newbie to get his head round the build toolchain and gain sufficient insight information to act alone without calling out for help.

Today I am confronted with a failing build on a platform, which is certainly not exotic (MacOS), and I find myself spending hours trying to understand where the errors occur.

Unless told otherwise, I have no supporting information to help me get through my build failures. And calling out for help won't be of great help because ( I presume) only a few have invested significant time in understanding the guts of the toolchain. Questions will take time to be answered, if ever answered.

### Build tool

When all goes well, the AST build toolchain seems to beat flat out the other tools mentioned above. It has (apparently) no dependencies, allows for all the GNU Autoconfigure probing without the M4 hell, and nicely lays out its build _products_.

_Opinion_: GNU Autotools are a fantastic suite. But they have a major inconvenience: M4. Opaque and to a certain extend clumsy. Probably a good compromise for portability 30 to 40 years ago. But no longer the ad hoc tool for today; pre-processing could be done the AST way :-)

Could the AST build toolchain be system agnostic and a possible replacement for GNU Autotools or CMake on other projects? A toolchain written in portable POSIX shell targeting any _raw_ (POSIX) UNIX or Linux.

Whilst this was probably a driver in its conception, we see, going through the source files that it depend on `bash` here, `lynx` or `wget` there, etc. So it is **not** agnostic and doesn't build on a **raw** system; it requires GNUish capabilities. Hence it targets UNIX/GNU or Linux/GNU platforms.

Note: for the reasoning let us ignore for now that we probably need `gcc` to avoid proprietary compilers (where such compilers still exist).

[...]

PR #76, opened 2017-10-14
Fix how overlapped buffers are handled

krader1961:
Trying to build this project on macOS (a BSD variant) resulted in two
`assert()` failures when running `nmake`. The problem is that there are
at least two places which pass overlapping buffers to functions which are explicitly
defined to have undefined behavior when the buffers overlap.

See issue #42.

### DIFF
--- a/src/cmd/nmake/variable.c
+++ b/src/cmd/nmake/variable.c
@@ -577,7 +577,11 @@ resetvar(register Var_t* p, char* v, int append)
 		p->value = newof(p->value, char, n + 1, 0);
 		p->length = n;
 	}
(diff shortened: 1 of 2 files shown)


krader1961:
So why did the Travis CI build fail? Looking at the log I don't see an obvious reason. Also, the fact the build output was so large that I had to download the "raw" log is disturbing. That shouldn't be necessary.

dannyweldon:
I think it failed because of the use of line comments, which are checked for using grep in the .travis.ci file.  You can see the error in the fedora build log at line 542.  I thought it was supposed to terminate the build on error??

PR #80, opened 2017-10-24
Remove invalid use of `memccpy()`

### LINKED ISSUE #78: overlapping buffers and `memccpy()`
The change I authored, commit de001193661b7389d8c201552d36de078cac5f4d, which was recently merged to fix building on BSD based systems like macOS only works when using `gcc` to do the compilation. If you instead use the `clang` compiler provided by Apple it fails because the `BSD` preprocessor symbol isn't defined. You also need to blacklist `__APPLE__`.

But here's the thing, the code before my change began with this preprocessor directive:

```
#if _lib_memccpy && !__ia64 /* these guys may never get it right */
```

The problem is that assertion is wrong. It implies that the libc implementation of `memccpy()` on IA64 is broken. Which is incorrect. The problem is the AST `sfputr()` function is depending on what is explicitly undefined behavior in the face of overlapping source and destination buffers. Every implementation is free to optimize the operation by assuming the buffers do not overlap but in so doing produce incorrect results, including aborting the program, if the buffers do overlap.

I was too timid with my previous change. I should not have added another blacklisted architecture. Instead the use of `memccpy()` should simply be removed. If you look at the two alternatives using `memccpy()` simply complicates the code and is unlikely to be measurably, let alone noticeably, faster.

krader1961:
The behavior of the `memccpy()` function is undefined when the buffers
overlap which is sometimes true for the `sfputr()` function. So always
use an explicit loop to copy the data.

Fixes #78

### DIFF
--- a/src/lib/libast/sfio/sfputr.c
+++ b/src/lib/libast/sfio/sfputr.c
@@ -46,7 +46,7 @@ int		rc;	/* record separator.	*/
 
 	SFLOCK(f,0);
 
-	f->val = sn = -1; ss = (char*)s; 
+	f->val = sn = -1; ss = (char*)s;
 	for(w = 0; (*s || rc >= 0); )
 	{	/* need to communicate string size to exception handler */
 		if((f->flags&SF_STRING) && f->next >= f->endb )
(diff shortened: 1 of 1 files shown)


krader1961:
You might argue that we should still use `memccpy()` where it is known to handle overlapping buffers (e.g., glibc on x86). But that's dangerous. It would probably never happen but theoretically a future libc release could change the implementation such that overlapping buffers are no longer handled. More importantly, just because glibc on x86 handles this situation that is no guarantee that glibc on another architecture (e.g., ppc) will do so. Thus you can't use a blacklist like the code currently does. You have to use a whitelist. And as I noted in issue #78 the added complexity isn't justified.

PR #83, opened 2017-10-27
Correctly handle multibyte characters when KEYBD trap is set

siteshwar:
Resolves: rhbz#1503922

### DIFF
--- a/src/cmd/ksh93/edit/edit.c
+++ b/src/cmd/ksh93/edit/edit.c
@@ -1048,7 +1048,10 @@ int ed_getchar(Edit_t *ep,int mode)
 		/* The while is necessary for reads of partial multbyte chars */
 		*ep->e_vi_insert = (mode==-2);
 		if((n=ed_read(ep,ep->e_fd,readin,-LOOKAHEAD,0)) > 0)
+		{
+			/* putstack is pushing negated characters on the stack */
 			n = putstack(ep,readin,n,1);
+		}
 		*ep->e_vi_insert = 0;
 	}
 	if(ep->e_lookahead)
@@ -1089,13 +1092,17 @@ int ed_getchar(Edit_t *ep,int mode)
 							break;
 					}
 				}
-				if(n=keytrap(ep,readin,n,LOOKAHEAD-n,mode))
-				{
-					putstack(ep,readin,n,0);
-					c = ep->e_lbuf[--ep->e_lookahead];
-				}
 				else
+				{
+					/* unnegate the character */
+					c = -c;
+				}
+
+				if(!keytrap(ep,readin,n,LOOKAHEAD-n,mode))
+				{
 					c = ed_getchar(ep,mode);
+				}
+
 				ep->e_keytrap = 0;
 			}
 			else


kdudka:
After staring at the code for 15 minutes, I am not able to explain myself what the bug is and how the patch will fix it.  Could you please explain it briefly and include the explanation as a comment into the patch?

siteshwar:
[This call](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1051) to putstack pushes negated characters on the stack . See the description of [putstack](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L938).

First character was again put on the stack by [this line](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1094) and next character was read [here](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1095). It was causing issues with multibyte characters and I could not find a reason why it's written this way. The last read character is in ```c``` variable so there is no need to push it back to the stack and reread it.I have removed the code under [this block](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1093-L1096). Also, if [this uncondition](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1068) is not true we should unnegate the character and it's returned back in last line.

siteshwar:
So it looks like there is an interface to change input buffer [inside ```keytrap``` function](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1516-L1519) but it's not obvious how it works.

Also [```n``` is always initialized to ```1```](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1067). It's passed as an [argument to ```keytrap()```](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1092) function and  will cause corruption with multibyte characters because ```keytrap``` function modifies input buffer to [terminate with ```NULL```](https://github.com/att/ast/blob/9c887749d471713e798772867cafb442de59c0f4/src/cmd/ksh93/edit/edit.c#L1507) based on ```insize```.

krader1961:
Like @kdudka I can't figure out what the problem is or how this change fixes it. As best I can tell the core issue seems to be that the ksh code assumes the char type is signed (or perhaps unsigned). Which is an invalid assumption. Rather than papering over that invalid assumption in this specific block of code it would be better to address the invalid assumption. I'm especially bothered by the

```
/* unnegate the character */
c = -c
```

portion of the change since it seems to assume the `char` type is signed rather than unsigned.

krader1961:
What do you want to do with this PR, @siteshwar? It's more than a month old. Is this something from the Fedora patches you are upstreaming? I can take a closer look but as I noted in my previous comment the `c = -c` statement is suspicious since it seems to assume the `char` type is signed. Which is an invalid assumption.

kdudka:
IIRC, we agreed with @siteshwar that the proposed patch would not work properly.  Has the PR been updated since then?

In any case, we should have at least some smoke test for a keytrap function modifying the input buffer before touching this code.

kdudka:
On Thursday, December 7, 2017 6:28:50 AM CET Kurtis Rader wrote:
> Is this something from the Fedora patches you are upstreaming? I can take a
> closer look but as I noted in my previous comment the `c = -c` statement is
> suspicious since it seems to assume the `char` type is signed. Which is an
> invalid assumption.

I am afraid that this assumption is made on multiple places in ksh code.  
Perhaps we could use -fsigned-char on compilers that support it to satisfy
the assumption?

siteshwar:
@krader1961 It's not related to fedora patches.

I am going to close this pull request and will open a separate issue for it.

krader1961:
Okay, SGTM. Regarding signed versus unsigned char: It has been a long time since I've used a C compiler which defaults to unsigned char. So it may be sufficient to simply add a compile time check that chars are signed and abort the build if they aren't. If and when someone reports that check is failing we can decide what to do (e.g., add compiler flags, etc.).

kdudka:
On Thursday, December 7, 2017 4:13:03 PM CET Kurtis Rader wrote:
> Okay, SGTM. Regarding signed versus unsigned char: It has been a long time
> since I've used a C compiler which defaults to unsigned char. So it may be
> sufficient to simply add a compile time check that chars are signed and
> abort the build if they aren't. If and when someone reports that check is
> failing we can decide what to do (e.g., add compiler flags, etc.).

It is not only about compiler but also the target architecture.  Even GCC 
defaults to unsigned char on s390x, which is an architecture for which we 
regularly build RHEL-6 and RHEL-7 packages of ksh.  Nothing exotic really.

krader1961:
> Even GCC defaults to unsigned char on s390x...

Yes, this isn't really a compiler issue per se it's a question of whether it is more appropriate (or efficient) for the underlying machine architecture to treat chars as signed or unsigned. And upon further reflection this shouldn't matter if the code is written correctly. It's well defined what happens when you negate an unsigned char in C. Still, it is bad form to do so IMO. I guess someone will need to spend the time to figure out if the code is actually doing the right thing on architectures like s390x where the EBCDIC char set has code points with the high bit set.

PR #86, opened 2017-10-30
Allow building with `clang` rather than `gcc`

krader1961:
This hack allows using clang to build the project. You need to

    export USE_CLANG=yes
    export CC=clang

before running `bin/package make`. I didn't implement a cleaner solution
because most of the active contributors agree we want to replace Nmake
with Meson or Cmake ASAP. Hence this hack to support using clang and
(soon) oclint. I've tested this on macOS, FreeBSD, and Ubuntu 16.10.


siteshwar:
LGTM. It works fine on Linux with ```clang```.

PR #90, opened 2017-11-02
Assume `MAP_ANON` is valid

### LINKED ISSUE #84: Remove feature test to check if mmap should be used
This [iffe feature test](https://github.com/att/ast/blob/0fa99a9db97f0b3db5e4f2363c461c3dd4f50f3e/src/lib/libast/features/mmap#L206) results in non-determinstic builds, because it [compares loop execution times at build time](https://github.com/att/ast/blob/0fa99a9db97f0b3db5e4f2363c461c3dd4f50f3e/src/lib/libast/features/mmap#L302). I believe it can be removed and we can choose to always use ```mmap```.

krader1961:
One of the feature tests involving `mmap()` is dubious since it involves
highly variable timing data that can lead to non-deterministic builds.
Several other feature tests are no longer relevant since, for example, I
can't think of a single OS we care about which supports `mmap()` but not
the `MAP_ANON` flag.

Finally, there is no good reason to even use `mmap()` and implement our
own memory allocator. These days we should be using the OS provided libc
`malloc()` family of functions. And the value of the performance speedup
of using it in the SFIO code is dubious for a command shell. However, I
found that when I disabled it completely it results in more test failures
so this change is limited to the MAP_ANON related configuration.

Fixes #84

### DIFF
--- a/src/lib/libast/features/mmap
+++ b/src/lib/libast/features/mmap
@@ -16,14 +16,7 @@ tst	lib_mmap note{ standard mmap interface that works }end execute{
 
 	#define Failed(file)	(remove(file),1)
 
-	int
-	#if _STD_
-	main(int argc, char** argv)
-	#else
-	main(argc,argv)
-	int     argc;
-	char**  argv;
-	#endif
+	int main(int argc, char** argv)
 	{
 		caddr_t		mm;
 		char		*file;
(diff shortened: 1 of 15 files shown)


siteshwar:
LGTM

PR #91, opened 2017-11-02
Disable hijacking some standard POSIX functions

siteshwar:


siteshwar:
I have left some references to ast_map.h inside build scripts and sources. But none of the hijacked functions should be used now.

krader1961:
LGTM. But in light of the recent discussion on the ast-developers mailing list and issues like #42 this is probably an opportune time to decide how we want to handle deviations from the old way of doing things and put a stake in the ground. We've talked about three options:

1) Clone the directories needed to build ksh93 into a distinct subtree that is part of the current `beta` branch and might become the `master` branch.

2) Create a branch for these types of changes leaving the existing `beta` (and `master`) branch for work on the legacy code base where things are done the old way (e.g., using Nmake) and which supports all the legacy platforms.

3) Create a branch that represents the legacy code base and build process, drop the `alpha` and `beta` branches and do all work of this nature, including changing the build tools and dropping everything unrelated to ksh93, in the `master` branch.

There are obviously variations on each of those options. Even though I proposed option 1 I am no longer advocating for it. Given how git handles branches I think that approach causes more confusion than benefit.

Of the other two options I prefer 3. That, however, is going to be politically and emotionally sensitive. Which I can understand had I written large chunks of this project's code. Which means option 2 is probably the best choice today. If ksh93 (or whatever it is called) is still a viable shell, undergoing active development, in a couple of years things can be reshuffled and the `master` branch switched to track the new code base.

siteshwar:
I will vote for 3. Unfortunately, as you said the discussion gets political and emotional here. I would like to hear what the old timers have to say about it.

siteshwar:
The beta branch has already lost it's meaning and it will go away as soon as we agree on the restructuring of the branches. Until then I think it's fine to merge new commits into it.

PR #92, opened 2017-11-02
Remove feature test for botched return value from get_groups()

siteshwar:


src/lib/libast/Mamfile
```diff
@@ -266,11 +262,6 @@ prev std/endian.h implicit
 done FEATURE/sys generated
 prev FEATURE/lib implicit
 prev FEATURE/common implicit
-done features/botch.c
```
krader1961: I admit I know nothing about Nmake and don't intend to learn about its behavior. But I am curious why the preceding lines (e.g., `prev FEATURE/common implicit`) should not also be removed based on a naive reading of this config file. Other than that this LGTM. But per my comment on your previous change merging this should be predicated on deciding how we want to handle such deviations from the existing code base.

siteshwar: I tried to remove the full block but it leads to strange compiler warnings that are not worth investing time to fix. I believe there is an issue with specifying dependencies in Mamfiles.

kdudka:
This should be safe if it really does what I think it does.  System ```getgroups()``` should return correct value on modern systems...

PR #93, opened 2017-11-02
Remove a 'hack' generated by feature test

siteshwar:
It looks like it is generated to hijack standard functions, so it should be safe to remove.


krader1961:
LGTM but per my comment on similar changes I think we need to decide on how we want to handle potentially non backward compatible changes.

The people arguing that this project should continue to use Nmake and attempt to support dead operating systems (e.g., DYNIX/3, Solaris) are basically arguing that ksh be frozen. I think that doing so is the death of ksh.

PR #94, opened 2017-11-03
Remove feature test for sys_nerr and sys_errlist

siteshwar:


krader1961:
LGTM. We should probably start a TODO file or perhaps just open issues for next steps. For example, we can safely assume `strerror()` is available. There shouldn't be any need for the AST implementation (which seems to be the only reason for testing that `sys_errlist` is available).

siteshwar:
Yes, I noticed it. And it should be removed too.

PR #95, opened 2017-11-03
Disable uwin related feature tests and compatibilty functions

siteshwar:


krader1961:
LGTM. Builds on macOS. Somewhat surprisingly it reduced the build time only a few seconds.

PR #97, opened 2017-11-04
Disable failing tests on Linux

siteshwar:
I have disabled some tests that fail on Linux and marked them to be fixed later.

PR #98, opened 2017-11-04
Enable running ksh tests on Travis

siteshwar:


siteshwar:
There are tests failing on Linux. I have disabled some of them through #97, however some tests are failing only on Travis. We need to check if those tests can be fixed, or we can just disable them only for Travis.

siteshwar:
There are some issues with setting permissions on files while running tests in Docker. It might be a bug in Docker.

siteshwar:
Nevermind. It's happening because we are running tests as ```root``` user.

siteshwar:
I have modified the travis script to run as normal user, however there are still different tests failing on Fedora and openSUSE.

siteshwar:
I finally got around failing tests on docker. Changes in this pull request and #97 should enable running ksh tests on Travis.

.travis.yml
```diff
@@ -15,11 +15,24 @@ before_install:
...
         export CCFLAGS='-fno-strict-aliasing -Wno-unknown-pragmas -Wno-missing-braces -Wno-unused-result -Wno-return-type -Wno-int-to-pointer-cast -Wno-parentheses -Wno-unused -Wno-unused-but-set-variable -Wno-cpp -P';
         export CC=gcc;
         ./bin/package;
-        ./bin/package make -S 2>&1 | tee out.log ;
-        grep -q 'error:' out.log && exit 1 || exit 0;"
+        ./bin/package make ast-ksh 2>&1 | tee out.log;
+        grep -q 'error:' out.log && exit 1;
+        ln -s /usr/bin/cat arch/$(sh bin/package)/bin/;
+        ln -s /usr/bin/tee arch/$(sh bin/package)/bin/;
+        ln -s /usr/bin/chmod arch/$(sh bin/package)/bin/;
```
krader1961: I also noticed that these commands aren't built by `bin/package make ast-ksh` when someone else suggested I do that when I pointed out it's silly to have to build the entire AST project just to build and test ksh. Apparently that person has never tried to run the unit tests. Not to mention there is obviously missing dependencies in the definition of the `ast-ksh` target. These symlinks are fine for now but obviously we'll want to remove this hack once we've changed the build tool chain.

siteshwar: I agree that we should remove these symlinks later.

.travis.yml
```diff
@@ -15,11 +15,24 @@ before_install:
...
+        ulimit -n 1024;
+        ./bin/package test ksh93 2>&1 | tee test.log;
+        grep -q 'failed' test.log && exit 1 || exit 0;"
```
krader1961: This seems fragile. I wouldn't expect the word `failed` to appear in another context in the test output but you never know. Perhaps something like `grep -q '^test [^ ]* failed'` to help ensure it only matches entries like this one (from running the tests on my Ubuntu server):

```
test pty failed with exit code 1 [ 20 tests 1 error ]
```

krader1961:
Other than making the final grep pattern more restrictive this LGTM.

siteshwar:
The tests in ```signal.sh``` seem a bit unreliable and can cause the build process to hang. I think we should  put a note about it in the build log to avoid unncessary debugging.

dannyweldon:
The only thing is that it has increased the build time significantly now which may be another issue.

PR #99, opened 2017-11-05
Assume math.h functions are available

### LINKED ISSUE #88: Many math functions should be removed
When I found this project and was pondering whether to contribute I wrote this [blog article](http://www.skepticism.us/2017/10/ksh93-has-unexpected-undocumented-support-for-math-functions/) about the presence of math functions that have no legitimate purpose in a CLI shell.

I admit this issue is mostly driven by a desire to reduce the build time to something more reasonable and consistent with the time required to build other shells. More fundamentally why should `ksh` support more than one method for rounding a floating point value to an integer? Let alone support obscure libc math functions like `fma()` which does this according to `man fma` on my system:

> The fma() functions compute (x*y)+z, rounded as one ternary operation: they compute the value (as if) to infinite precision and round once to the result format, according to the current rounding mode.

Supporting those types of functions is especially confusing since they often depend on concepts like [NaN](https://en.wikipedia.org/wiki/NaN) which make no sense in the context of a command line shell such as ksh.

I propose we eliminate any math function not relevant to a shell script.

krader1961:
Bypass the incredibly inefficient and expensive (59 seconds on my
system) process of figuring out which math functions are available. It
does this by using a handcrafted version of the C module the
./src/cmd/ksh93/features/math.sh script creates if it can be built and
linked without error. That should be the case for any modern UNIX system
we care about. Thus avoiding the expensive probing for each function
individually.

Fixes #88


dannyweldon:
Thanks Kurtis.  LGTM.

siteshwar:
I have enabled ```ksh``` tests to run on Travis now. It would be good to rebase this branch with ```beta``` and let the tests run through Travis.

krader1961:
Merged as commit 26512cffd926e9d74f99de385258c16cb88ef2fb.

PR #101, opened 2017-11-05
Disable some tests only on openSUSE

siteshwar:


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -731,15 +731,20 @@ fi
...
-echo "TODO: Skipping test - 'cd with no arguments fails if HOME is unset'. It should be fixed later."
-#$SHELL <<- \EOF
-#	home=$HOME
-#	unset HOME
-#	cd 2> /dev/null
-#	[[ $(pwd) == "$home" ]]
-#EOF
-#[[ $? == 0 ]] || err_exit 'cd with no arguments fails if HOME is unset'
+# Below test fails on openSUSE
+if $(grep -q openSUSE /etc/os-release)
```
krader1961: Shouldn't this just be `if grep -q OpenSUSE /etc/os-release`?

src/cmd/ksh93/tests/shtests
```diff
@@ -323,6 +323,14 @@ do	if	[[ ! -r $i ]]
 		[[ $utf8 == 0 || $i == $setslocale ]] || locales+=" C.UTF-8"
 		for lang in $locales
 		do	o=$u
+
+			# This test only fails on openSUSE on Travis
+			if [[ $i == "builtins.sh" ]] && [[ $lang == "C.UTF-8" ]] && [[ $IS_RUNNING_TRAVIS ]] && $(grep -q openSUSE /etc/os-release)
```
krader1961: Same thing here: the `$( )` form seems wrong since we want to test the exit status, not whether `grep -q` wrote anything to stdout (which it wouldn't).

siteshwar:
Right! I have updated the pull request.

krader1961:
LGTM

siteshwar:
It caused some strange failures in CI. I have removed it from beta branch for now.

krader1961:
> It caused some strange failures in CI. I have removed it from beta branch for now.

Have you enabled Travis to build your fork? I recommend doing so as it lets you know if your change has a problem before you create a PR and others notice. When I was contributing to fish-shell it's one of the things I added to their *CONTRIBUTING.md* document. I think I'll strip out the fish specific items from that doc and install a copy here.

siteshwar:
Yes, I have enabled it on my fork. But I actually did not run the tests on travis because they were working fine my local system. It's possible that some of the tests are unreliable and behave strangely under resource constrained environments.

PR #102, opened 2017-11-05
Remove feature test and compatbility code for eaccess() function

siteshwar:


krader1961:
This change confuses me.

Note that `eaccess()` (or `euidaccess()` for which the former is just an alias) exists on Linux but not BSD (including macOS). This change does not remove (or replace) all uses in the AST code. Which surprises me since after applying this change I can still build ksh93 and all the other AST code (which is not all of it) on macOS. It would appear that removing this compatibility function doesn't actually break anything I care about (i.e., building ksh93). But exactly why that is true is not clear.

This change is unlike PR #105 which removes all references to the `strmode()` function.

siteshwar:
I presumed eaccess is [available on BSD](https://www.freebsd.org/cgi/man.cgi?query=eaccess&sektion=2).

krader1961:
Yes, `eaccess()` is available on FreeBSD, but it's not on macOS. But macOS does provide `faccessat()` with the `AT_EACCESS` flag. It looks like things still work on macOS with this change because of this block of code in *src/lib/libast/port/intercept.c*:

```
int
ast_eaccess(const char* path, int mode)
{
        return ast_faccessat(LOCAL(pwd), path, mode, AT_EACCESS);
}
```

So, LGTM.

Issue #40, opened 2017-05-24
typeset -S within functions


marcastel:
Weird behaviour when using static function variables (`typeset -S`)

Consider the following script:
```{.sh}
function alpha {
    integer -S count=0
    (( ++ count ))
    print -n " $count"
}

function beta {
    integer count=0
    (( ++ count ))
    print -n " $count"
}

print -n "Alpha:"; alpha; alpha; alpha; alpha; alpha; print
print -n "Beta: "; beta;  beta;  beta;  beta;  beta;  print
```

My understanding is that since the static variable is declared within a non-POSIX function its scope is that function's scope. Consequently the expected output should be:

```{.sh}
Alpha: 1 2 3 4 5
Beta:  1 1 1 1 1
```

While the output of the `alpha()` function is consistent across tests. The behaviour of `beta()` is not consistent. I get various outputs, sometimes correct, but mostly incorrect. This has been tested with 93u+ on Linux and macOS.

Sample outputs:
``` 
Beta:  0 0 0 0 0
Beta:  0 0 0 1 0
Beta:  0 0 1 0 0
Beta:  1 1 0 1 1
```

I have not detected the pattern. Consequently I can reproduce the error (almost systematically), but not the exact output.

I tried the following without success:
- Change the loading order of functions
- Declare a variable with same name in the global scope (to eventually force a local scoping)
- Replace the `integer` alias by its `typeset` equivalent

jelmd:
At least on Solaris 11.3 SRU 24.4 it works as expected, i.e. embedded into a `for (( I=0; I < 20; I++ )); do  ... done` it produces expected results, i.e.:
```
Alpha: 1 2 3 4 5
Beta:  1 1 1 1 1
Alpha: 6 7 8 9 10
Beta:  1 1 1 1 1
Alpha: 11 12 13 14 15
Beta:  1 1 1 1 1
Alpha: 16 17 18 19 20
Beta:  1 1 1 1 1
...
```

However, on Ubuntu 16.04 I see similar misbehavior as you:
```
Alpha: 1 2 3 4 5
Beta:  1 0 0 0 0
Alpha: 1 2 3 4 5
Beta:  0 0 1 1 0
Alpha: 1 2 3 4 5
Beta:  0 0 1 1 0
Alpha: 1 2 3 4 5
Beta:  0 0 1 1 0
...
```

Perhaps it is worth to check the patches applied to the Solaris version: https://github.com/oracle/solaris-userland/tree/master/components/ksh93/

[...]

PR #103, opened 2017-11-05
Trim down feature tests for file modes

siteshwar:


src/lib/libast/features/mode.c
```diff
@@ -27,192 +27,22 @@
...
-	int	n;
-	int	idperm;
-	int	idtype;
-
-	idperm = idtype = 1;
 #ifndef S_ITYPE
 #ifdef	S_IFMT
 	printf("#define S_ITYPE(m)	((m)&S_IFMT)\n");
 #else
 	printf("#define S_ITYPE(m)	((m)&~S_IPERM)\n");
```
krader1961: Ugh! I've never heard of the `S_IPERM` or `S_ITYPE` symbol. It's not defined on macOS or Linux. I can't find them mentioned in the APUE book either. So either this feature test can be completely eliminated or there are platforms we care about which have this unusual configuration. If the latter a comment to that effect would be a good idea.

siteshwar: None of these macros are provided by default and there is some ast code that depend on them. I plan to clean it up later.

krader1961: SGTM, ship it.

jelmd: Ehmm, if it is not ready, it should never be shipped/commited! Wondering, where you learned to do such insane things ...

krader1961: @jelmd, Your comments are not constructive and their tone is condescending. AFAICT you haven't actually contributed any changes. On the contrary, you've made it clear you won't be contributing. In issue #40 I suggested that you or @marcastel could contribute by helping port some of the Solaris patches to ksh93 to this project. You snidely replied:

> It was just a hint, to possibly safe you some time. If you prefer troubleshooting it by your own and provide your own fix, just go ahead.

So the next time you start to write a comment might think about what I've said if you want anyone to pay attention to what you have to say.

jelmd:
@krader1961: Not sure, how you come to this conclusion. I've already shown, how one can build ksh only, provided related files and made suggestions, how to improve things, how things should/could be done and also mentioned, what IMHO should  not be done. I also gave a hint, where more patches are, which might help to get problems solved. You explained, why one should not use them/it is not worth to look at it, and now you demand, that other people should do it. I think, something doesn't match here.

PR #104, opened 2017-11-06
Remove feature tests and some code to intercept functions provided by dirent.h

siteshwar:


krader1961:
LGTM. I can't imagine that any of these compatibility checks are still needed for any platform we care about.

PR #105, opened 2017-11-06
Remove the `strmode()` function

krader1961:
A recent cleanup by @siteshwar revealed that AST defines a `strmode()`
function that conflicts with the function of the same name provided on
BSD distros (including macOS). That function isn't actually used by any
of the AST code so just remove it.


jelmd:
Ehmmm, you know that you are breaking with this an API again?

krader1961:
@jelmd, Yes, I know that. Most of the recent changes either do, or might, break external uses of the AST library. As you can see from the comment for this change it is just the simplest solution for the breakage caused by an earlier change. These changes are not meant for general use.  The work being done is to facilitate making it easier to maintain the ksh code.

jelmd:
Hmmm, did you ask before, whether one is using libast e.g. in its own utils, cares about is breakage?
IMHO just "making it easier to maintain ... code" is not an acceptable reason to break anything.

krader1961:
@jelmd, The intent is to create a branch (whether `master` or `ast-full` is still being discussed) that includes all of the open sourced AST project code. And which omits changes like this one which will be limited to a branch where only ksh93 development occurs. Anyone who wants to build against that version of the project can then do so without worrying about changes to the API. Of course that branch will be more or less static as of the time this project was open sourced since I predict there will be very few changes made to that branch (whatever it is named).

kdudka:
It might be a stupid question.  Do any open source projects (not included in this git repo) use libast directly?

jelmd:
@krader1961: wasn't what I was asking for, but anyway, thanx for the answer.

@kdudka: Not sure, whether its use is restricted to open source. However, because documented it is a public interface and thus at least a minimal effort should be made, to find out, whether someone gets hurt by this cut. At least [here](https://alanhargreaves.wordpress.com/category/dtrace/) it seems, that it is used a separate lib and thus could be potentially used by others as well ...

krader1961:
@kdudka, I actually spent some time doing google searches to try and answer that question. It's hard because there appear to be at least two projects named "libast" (with varying capitalization) which are being used by other projects and which have packages which can be installed on various distros. But as best I can tell the number using the AT&T AST "libast" is so close to zero that we don't care.

Note that the example [linked to](https://alanhargreaves.wordpress.com/category/dtrace/) by @jelmd is for someone debugging why ksh93 was causing problems. The `sh_mactrim()` function at the top of the stack in that article is from *src/cmd/ksh93/sh/macro.c*. So that is not an example of an independent open source project using libast.

kdudka:
Then I would not invest much effort to keep the libast ABI/API compatible beyond this git repo.  Keeping ksh behavior backward compatible is way more important in my opinion.

krader1961:
Merged as b7a7930e148b7d2774fa11ddf1adb031ab743d18

PR #109, opened 2017-11-08
Remove socket related feature tests and intercept functions

siteshwar:


krader1961:
LGTM. Builds on macOS and FreeBSD. Two merge conflicts in *src/lib/libast/Mamfile* with your other recent PR that has an obvious fix.

krader1961:
Also, it won't materially affect the build time now that I've figured out how to workaround the hang that happens from the build system doing `cat /proc/registry/HKEY_LOCAL_MACHINE/Hardware/Description/System/CentralProcessor` on macOS when doing an incremental build. But we might as well start removing the subsystems (e.g., directories) not needed to build ksh as well as remove components of a subsystem like this PR does. Any objection?

siteshwar:
Sounds good to me.

PR #111, opened 2017-11-09
Remove code for libcs library

siteshwar:
ksh93 does not make use of this library. It should be removed from the sources.


krader1961:
LGTM. The Travis failures look like false positives and I can build with this on macOS.

siteshwar:
I have tried restarting the travis builds, but below test has consistently failed on suse. I need to investigate it further.

```
shtests: line 361: 18547: Memory fault(coredump)
test builtins(shcomp) failed at 2017-11-09+23:33:30 with exit code 267 [ 196 tests 267 errors ]
```

krader1961:
The reason this PR is causing Travis failures is because *src/cmd/coshell* requires the *libcs* code. And, AFAICT, that is only needed for some unit tests on some platforms. I attempted to delete *src/cmd/coshell* and found that `bin/package test` subsequently failed on some, but not all, platforms I tested.

Since the Solaris OS is now officially dead do we care if we break support for it in order to continue keeping `ksh` relevant? Obviously I would like to see the `ksh` command work on as many platforms as possible. The question is what do we do during a transition period.

siteshwar:
If the test cases pass on Linux, FreeBSD and macOS. Feel free to merge your changes.

krader1961:
Let's leave *lib/libcs* in place for the moment. Once we have unit tests running under Meson we can start removing the Nmake related code. That may make it possible to remove *cmd/coshell* which would allow us to remove *lib/libcs*. If not we can take a closer look at why this is needed to run the unit tests on Solaris. Even though Solaris it now officially dead that change in status was just a couple of months ago so I think we should try to support it for the immediate future.

jelmd:
Solaris is supported at least 'til 2032. Why do you not check, before making such unfounded guesses/statements?

PR #114, opened 2017-11-09
Remove code for libexpr library

siteshwar:
It is not a depndency for ksh93 and should be dropped from the sources


krader1961:
LGTM

PR #115, opened 2017-11-09
Avoid buffer overrun after duplicating fd in iousepipe()

siteshwar:
Duplicating a fd may return a descriptor that is greater than shp->gd->lim.open_max, this can cause a buffer overrun while assigning a value to shp->fdstatus[subpipe[2]]. This patch fixes it. Thanks to Paulo Andrade <pandrade@redhat.com> for the patch.
Resolves: rhbz#1506344

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -134,6 +134,8 @@ static bool iousepipe(Shell_t *shp)
 		return(true);
 	}
 	subpipe[2] = sh_fcntl(fd,F_dupfd_cloexec,10);
+	if(subpipe[2] >= shp->gd->lim.open_max)
+		sh_iovalidfd(shp,subpipe[2]);
 	shp->fdstatus[subpipe[2]] = shp->fdstatus[1];
 	while(close(fd)<0 && errno==EINTR)
 		errno = err;


src/cmd/ksh93/sh/xec.c
```diff
@@ -134,6 +134,8 @@ static bool iousepipe(Shell_t *shp)
 		return(true);
 	}
 	subpipe[2] = sh_fcntl(fd,F_dupfd_cloexec,10);
+	if(subpipe[2] >= shp->gd->lim.open_max)
+		sh_iovalidfd(shp,subpipe[2]);
```
krader1961: I don't understand this logic. Why not do it this way:

```
if (!sh_iovalidfd(shp, subpipe[2]) {
    sh_close(subpipe[1]);
    sh_close(subpipe[2]);
    return false;
}
```

The `sh_iovalidfd()` function is a nop and returns true if the fd is not -1 and less than `open_max`; otherwise it resizes the array. So there isn't any need to do that test before calling that function. Also, since it returns false if there is a problem we should check for that and bail out. Of course, I don't really know what should be done in the body of that `if` block since I haven't learned enough about the ksh internals. It may be a "can't happen" situation that calls for an `abort()` (or whatever is idiomatic for ksh in such situations).

krader1961:
Also, looking at the patches applied by the Solaris team I see they've fixed the same problem in other places as well. I'm going to make a concerted effort to merge those patches. I think that effort will fix this bug as well so let's hold off on merging this for a couple of days.

krader1961:
It turns out [patch 100-CR6964621](https://github.com/oracle/solaris-userland/blob/master/components/ksh93/patches/100-CR6964621.patch) addresses this particular issue as well as several others in the same vein. So I'm going to close this as redundant since I'm working on merging the changes by the Solaris team.

P.S., The Solaris patches don't do any better at handling `sh_iovalidfd()` returning false. This is a huge stinking source of bugs and unpredictable behavior. So my earlier comment about this change applies to the Solaris fix for the same problem. Which makes me a sad puppy since I've always thought of the Korn shell source code adhering to a higher standard of software engineering than bash or zsh.

PR #118, opened 2017-11-09
Fix exporting vars in subshells

### LINKED ISSUE #7: variable exported in function in a subshell is also visible in a different subshell
I've observed this issue in a Solaris 11 system with all the ksh93 versions(alpha, beta and the master versions) as well as in Ubuntu 14 ( ksh2012).

Here's a testcase which reproduces the issue.
```
# cat test1.ksh
# !/usr/bin/ksh

function proxy { 
         export MYVAR="bla" 
        child 
        unset MYVAR 
} 

function child { 
        echo "MYVAR=$MYVAR" >> /var/tmp/debug.log 
} 

function test { 
        $(child) 
        $(proxy) 
        $(child) 
} 

rm /var/tmp/debug.log 
test 
cat /var/tmp/debug.log 
# ./test1.ksh

MYVAR= 
MYVAR=bla 
MYVAR=bla <------------------------------ this should not happen 
# 
```

The following patch which removes an optimization fixes the issue. If there is any other patch, please let me know.

```
--- INIT.2012-08-01.old/src/cmd/ksh93/sh/subshell.c     2016-03-01 04:01:06.513890578 -0800
+++ INIT.2012-08-01/src/cmd/ksh93/shsubshell.c  2016-03-01 04:02:43.617872391 -0800
@@ -260,9 +260,6 @@
        shp = sp->shp;
        dp = shp->var_tree;
-       /\* don't bother to save if in newer scope */
-       if(sp->var!=shp->var_tree && sp->var!=shp->var_base && shp->last_root==shp->var_tree)
-               return(np);
      if((ap=nv_arrayptr(np)) && (mp=nv_opensub(np)))
      {
              shp->last_root = ap->table;
```

The issue seems to be happening because of the following optimization in sh_assignok function in sh/subshell.c.
```

 /\* don't bother to save if in newer scope */ 
 if(sp->var!=shp->var_tree && shp->last_root==shp->var_tree) 
        return(np); 
```

This optimization prevents saving the variables in situations where they don't need to be restored. If we remove the optimization, the issue will go away.

krader1961:
This change applies the fix from here:

https://github.com/oracle/solaris-userland/blob/master/components/ksh93/patches/250-22561374.patch

It also adds a regression test.

Fixes #7

### DIFF
--- a/src/cmd/ksh93/sh/subshell.c
+++ b/src/cmd/ksh93/sh/subshell.c
@@ -271,9 +271,7 @@ Namval_t *sh_assignok(register Namval_t *np,int add)
 		return(np);
 	shp = sp->shp;
 	dp = shp->var_tree;
-	/* don't bother to save if in newer scope */
-	if(sp->var!=shp->var_tree && sp->var!=shp->var_base && shp->last_root==shp->var_tree)
-		return(np);
+
 	if((ap=nv_arrayptr(np)) && (mp=nv_opensub(np)))
 	{
 		shp->last_root = ap->table;
--- a/src/cmd/ksh93/tests/subshell.sh
+++ b/src/cmd/ksh93/tests/subshell.sh
@@ -720,4 +720,34 @@ then	(( size = 117*1024 ))
 	fi
 fi
 
+# ========================================
+# Test that variables exported in subshells don't affect the outer shell.
+# Regression test for issue #7.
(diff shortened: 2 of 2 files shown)


krader1961:
Merged as c26719b9ac1d1021556a714ea03d1aa3f7194d9e.

PR #119, opened 2017-11-10
Remove feature tests and code for nl_types.h

siteshwar:
I have also disabled building catopen.c which was causing issues while compilation, as we can use standard catopen() function.


krader1961:
LGTM. Builds on macOS and FreeBSD11 and passes tests.

PR #120, opened 2017-11-10
Remove headers for bytesex.h and endian.h

siteshwar:


krader1961:
On macOS:

```
In file included from /Users/krader/projects/3rd-party/ast/src/lib/libsum/sumlib.c:167:
/Users/krader/projects/3rd-party/ast/src/lib/libsum/sum-sha2.c:67:10: fatal error: 'endian.h' file not found
#include <endian.h>
         ^
```

It looks like the portable way to get the symbols in that header is to `#include <sys/types.h>` since endian.h may be in different directories depending on the OS.

krader1961:
I confirmed that the following change lets it build on macOS, FreeBSD and Linux:

```
diff --git a/src/lib/libsum/sum-sha2.c b/src/lib/libsum/sum-sha2.c
index b11a657..8b79ccb 100644
--- a/src/lib/libsum/sum-sha2.c
+++ b/src/lib/libsum/sum-sha2.c
@@ -64,7 +64,7 @@
 #define __undef__USE_BSD
 #define __USE_BSD
 #endif
-#include <endian.h>
+#include <sys/types.h>
 #ifdef __undef__USE_BSD
 #undef __undef__USE_BSD
 #undef __USE_BSD
```

PR #123, opened 2017-11-13
Meson build configurations for libast

siteshwar:
Steps to build:

* In the root directory of sources execute :
```
mkdir build
cd build
meson ..
```

* To build the source:
```
ninja
```


bin/libast_prereq.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
ignatenkobrain: this should be bash

krader1961: > this should be bash

Agreed since the script uses features, notably `x=(a b c)` syntax, that is not supported by the real Bourne shell. In practice it probably doesn't matter since most systems make */bin/sh* an alias for */bin/bash* but better to be explicit. Also, there's a lot of weird whitespace in the script.

meson.build
```diff
@@ -0,0 +1,15 @@
+project('ksh93', 'c')
+run_command('bin/setup_cmds.sh')
+run_command('bin/libast_prereq.sh')
+
+add_global_arguments('-D_PACKAGE_ast', language: 'c')
```
ignatenkobrain: add_project_arguments

bin/libast_prereq.sh
```diff
@@ -0,0 +1,102 @@
...
+iffe_sh_tests=('preroot.sh'  'sig.sh')
+
+iffe_sh_tests_2=('param.sh' )
+
+iffe_sh_tests_3=('libpath.sh' )
+
+pushd $base_dir/src/lib/libast/features
+
+iffe -v -X ast -X std -c 'cc -D_BLD_DLL -D_BLD_ast    ' run param.sh
+mv $base_dir/src/lib/libast/features/FEATURE/param.sh $base_dir/src/lib/libast/features/FEATURE//param
```
krader1961: s,//,/,

krader1961: Also, pretty much everywhere else in this script you doublequote paths like these. It's unlikely `$base_dir` will contain problematic chars (e.g., a space) but better to be safe and quote the strings.

bin/libast_prereq.sh
jelmd: @krader1961: Not sure about your definition of "on most systems", but on Ubuntu machines /bin/sh defaults to **dash** since ~2008 (and probably on most other Debian based boxes since ~2011), on Solaris/SmartOS & friends to [amd64/]ksh93 since ~2010, on older Solaris versions (10-) to a minimal bourne shell! ...

So if non-POSIX features are used, the exact shell name in mind should be used, yes.

krader1961:
Yay! I also had to add a `#!/bin/bash` header to *bin/package*. Otherwise `meson ..` dies with

```
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/Users/krader/projects/3rd-party/ast/bin/package'
```

on macOS. On Ubuntu it fails with this:

```
../src/lib/libast/features/conftab.c:16:1: warning: missing braces around initializer [-Wmissing-braces]
 {
 ^
../src/lib/libast/features/conftab.c:16:1: note: (near initialization for ‘_ast_conf_prefix’)
../src/lib/libast/features/conftab.c:188:114: error: ‘_PC_MIN_HOLE_SIZE’ undeclared here (not in a function)
 { "MIN_HOLE_SIZE", { 0UL, 0 }, { 0UL, 0 }, CONF_LIMIT|CONF_MINMAX|CONF_UNDERSCORE, CONF_POSIX, 1, CONF_pathconf, _PC_MIN_HOLE_SIZE },
                                                                                                                  ^~~~~~~~~~~~~~~~~
../src/lib/libast/features/conftab.c:217:22: error: ‘OPEN_MAX’ undeclared here (not in a function)
 { "OPEN_MAX_CEIL", { OPEN_MAX, 0 }, { 0UL, 0 }, CONF_LIMIT|CONF_LIMIT_DEF, CONF_AST, 1, CONF_nop, -1 },
                      ^~~~~~~~
[3/463] Compiling c object 'src/lib/libast/ast@sha/aso_aso.c.o'
../src/lib/libast/aso/aso.c:44:19: warning: ‘lib’ defined but not used [-Wunused-const-variable=]
 static const char lib[] = "libast:aso";
                   ^~~
ninja: build stopped: subcommand failed.
```

krader1961:
On macOS after I fix the header in *bin/package* `meson ..` yields this weird output:

```
  File "/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/Users/krader/projects/3rd-party/ast/bin/package'
/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py,: cannot open `/usr/local/Cellar/python3/3.6.3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py,' (No such file or directory)
line:                                                                                                  cannot open `line' (No such file or directory)
1344,:                                                                                                 cannot open `1344,' (No such file or directory)
in:                                                                                                    cannot open `in' (No such file or directory)
_execute_child:                                                                                        cannot open `_execute_child' (No such file or directory)
```

Looks like there is some unexpected output (from running *bin/package*?) that Meson is trying to interpret.

bin/libast_prereq.sh
krader1961: Argh! Note that on FreeBSD 11 bash lives at */usr/local/bin/bash* and there is no symlink to it from */bin/bash*. We may need to do `#!/usr/bin/env bash` in these scripts. Or perhaps just mandate that either */bin/ksh* or */bin/bash* be valid even if that means the user has to manually create symlinks.

bin/libast_prereq.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
+
+# This script is used for feature detection and setting up other configurations that are required to build libast
+set -x
+set -e
+
+# This script is run from an unspecified directory so we have to determine directory paths
+# http://mesonbuild.com/Reference-manual.html#run_command
+script_path=`realpath $0`
```
jelmd: Hmmm, realpath is basically a Linux thing (and would be nice to have a similar ksh93 builtin).
So IMHO ``script_path=$( cd `dirname $0`; printf "$PWD" )`` or something similar would be better, and later `_xxx_dir=/${bla_dir%/*}` instead of ``_xxx_dir=`dirname bladir` ``  (efficiency).

krader1961: @jelmd is correct. The Fish shell project had to implement a fallback builtin for `realpath` because it isn't found on all the systems supported by the project. We should avoid using it if possible.

bin/libast_prereq.sh
dannyweldon: While, it's nice to use arrays, shouldn't this just be a plain POSIX script?

Eg.  Instead of:

    iffe_tests_2=( 'tv' 'tmx' )

Just use:

    iffe_tests_2="tv tmx"

Instead of pushd and popd, just use

    cd dir
    cd -

bin/libast_prereq.sh
```diff
@@ -0,0 +1,102 @@
...
+    
+    cp "$base_dir/src/lib/libast/features/FEATURE/$iffe_test" "$base_dir/src/lib/libast/features/$iffe_test.h"
+done
+
+$bin_dir/conf -v $base_dir/src/lib/libast/comp/conf.tab cc -D_BLD_DLL -D_BLD_ast
+
+# iffe behaves strangely when I try to pass it a string made with double quotes
+# I pass this function as an argument to iffe while processing .c files
+function cc_fun {
+    cc -D_BLD_DLL -D_BLD_ast -I. -I.. -Icomp -I../comp -Iinclude -I../include -Istd -I../std -I../features  "$@"
```
jelmd: Wondering, whether there is a way to say `gcc` instead of `cc` (or honor the CC env var) ...

bin/libast_prereq.sh
dannyweldon: > Hmmm, realpath is basically a Linux thing (and would be nice to have a similar ksh93 builtin).

There is already a realpath builtin in libcmd in the beta branch.

krader1961: > There is already a realpath builtin in libcmd in the beta branch.

Yes but that isn't available when running `meson`. As painful as it is we need to do one of two things: a) mandate that */bin/bash* or */bin/ksh* be available so we can use its features, or b) assume only Bourne shell capabilities are available. The latter excludes array vars and commands like `realpath`.

krader1961:
After changing the script headers to `#!/usr/bin/env bash` I was able to successfully run `meson ..`  on FreeBSD 11 and macOS. The `ninja` command to actually build things still has problems. For example, on macOS it looks like a feature test or dependency is missing:

```
  "_iconv", referenced from:
      __ast_iconv_open in comp_iconv.c.o
     (maybe you meant: __ast_iconv_open, __ast_iconv_close , __ast_iconv_list , __ast_iconv_move , __ast_iconv_write , __ast_iconv , __ast_iconv_name )
  "_iconv_close", referenced from:
      __ast_iconv_close in comp_iconv.c.o
     (maybe you meant: __ast_iconv_close)
  "_iconv_open", referenced from:
      __ast_iconv_open in comp_iconv.c.o
     (maybe you meant: __ast_iconv_open)
ninja: build stopped: subcommand failed.
```

But at least we're moving in the right direction. I would be happy to see this merged after fixing the fundamental problems (e.g., the dependency on `realpath` and use off array vars) even if it only supports building libast on a small subset of platforms we care about. We can then chip away at the issues that keep it from being used on other platforms.

bin/libast_prereq.sh
jelmd: Keep in mind, that on some systems /bin/ksh is or points to ksh88 (or even worse pdksh, mksh, ...)!

krader1961: > Keep in mind, that on some systems /bin/ksh is or points to ksh88 (or even worse pdksh, mksh, ...)!

True. But as long as we limit the features we use to the minimum required by POSIX 1003 that shouldn't be a problem. The question is whether we want to have these scripts work with pre-POSIX 1003 shells. I don't think that compatibility is worth it for the same reasons we're no longer going to support pre-ANSI C compilers. We should draw a line in the sand that a POSIX 1003 shell is the bare minimum required to build ksh. Which should not be a problem given that Meson requires a fairly modern Python implementation.

siteshwar: This script is just a temporary solution to run ```iffe``` feature tests for Meson builds. I think it would be fine to just depend on bash here. Ideally we should get rid of this script and move all the feature tests to Meson.

siteshwar:
@krader1961 Can you try the latest changes from this branch on FreeBSD ? I suspect that ```iffe``` behaves differently when it was run with BSD version of ```/bin/sh```. I have modified it to always run under ```bash```. On top of my changes from this branch I had to add this patch:

```
index 4a0c31d..7325e66 100644
--- a/src/lib/libast/meson.build
+++ b/src/lib/libast/meson.build
@@ -14,4 +14,4 @@ subdir('misc')
 subdir('tm')
 subdir('vmalloc')
 
-library('ast', files, include_directories : incdir)
+library('ast', files, include_directories : incdir, link_args: '-lm')
```

And I was able to reduce the errors to :
```
/root/upstream/ast/build/../src/lib/libast/path/pathkey.c:209: undefined reference to `environ'
src/lib/libast/ast@sha/misc_cmdarg.c.o: In function `cmdopen_20120411':
/root/upstream/ast/build/../src/lib/libast/misc/cmdarg.c:127: undefined reference to `environ'
src/lib/libast/ast@sha/misc_procopen.c.o: In function `procopen':
/root/upstream/ast/build/../src/lib/libast/misc/procopen.c:472: undefined reference to `environ'
/root/upstream/ast/build/../src/lib/libast/misc/procopen.c:473: undefined reference to `environ'
/root/upstream/ast/build/../src/lib/libast/misc/procopen.c:474: undefined reference to `environ'
src/lib/libast/ast@sha/misc_procopen.c.o:/root/upstream/ast/build/../src/lib/libast/misc/procopen.c:474: more undefined references to `environ' follow
cc: error: linker command failed with exit code 1 (use -v to see invocation)
```

krader1961:
I had to manually install an updated `meson` on Ubuntu 16.10 since the one in its repos is too old. After doing that Meson was able to build *libast.so*.

On FreeBSD 11 Meson fails to build *libast.so*. The link command does not include `-lm` so in addition to not finding symbol `environ` it fails to find all the math functions:

```
cc  -o src/lib/libast/libast.so [long list of AST object files] -Wl,--no-undefined -Wl,--as-needed -shared -fPIC -Wl,--start-group -Wl,-soname,libast.so -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/usr/home/krader/projects/3rd-party/ast/build/src/lib/libast
```

If I manually add the `-lm` to the final `cc` invocation the missing math function errors go away. I haven't figured out what the problem is with the `environ` references not being resolved. But the errors go away if I remove the `-Wl,--no-undefined` flag and I then get a *libast.so* file. We probably just need to tell Meson it should omit that flag.

On MacOS 10.13.1 it seems to build all the object files but fails while building the library:

```
  "_iconv", referenced from:
      __ast_iconv_open in comp_iconv.c.o
     (maybe you meant: __ast_iconv_open, __ast_iconv_close , __ast_iconv_list , __ast_iconv_move , __ast_iconv_write , __ast_iconv , __ast_iconv_name )
  "_iconv_close", referenced from:
      __ast_iconv_close in comp_iconv.c.o
     (maybe you meant: __ast_iconv_close)
  "_iconv_open", referenced from:
      __ast_iconv_open in comp_iconv.c.o
     (maybe you meant: __ast_iconv_open)
ninja: build stopped: subcommand failed.
```

The final command to build the library is missing `-liconv` (as well as `-lm`) so a feature test for that library needs to be added:

```
cc  -o src/lib/libast/libast.dylib [long list of AST object files] -shared -install_name /Users/krader/projects/3rd-party/ast/build/src/lib/libast/libast.dylib '-Wl,-rpath,$ORIGIN/'
```

bin/libast_prereq.sh
siteshwar: ```realpath``` and ```dirname``` are available on Linux and FreeBSD. I think on macOS you can get it from coreutils package in homebrew. As I have mentioned earlier, this script is just a workaround to run iffe tests, so we should not worry too much about compatibility.

krader1961: > As I have mentioned earlier, this script is just a workaround to run iffe tests, so we should not worry too much about compatibility.

Agreed. We just need to keep this sort of thing in mind as we evolve the build process.

krader1961:
I've got it working on FreeBSD by invoking it via `meson -D b_lundef=false ..` and with this change to *src/lib/libast/meson.build*:

```
@@ -14,4 +14,7 @@ subdir('misc')
 subdir('tm')
 subdir('vmalloc')

-library('ast', files, include_directories : incdir)
+cc = meson.get_compiler('c')
+m_dep = cc.find_library('m', required : false)
+
+library('ast', files, include_directories : incdir, dependencies : m_dep)
```

I expect to figure out how to add `-liconv` when necessary later today. If you don't beat me to it, @siteshwar, I'll go ahead and merge this with those changes. I'll also start a *CONTRIBUTING.md* document to capture what we've learned so far about using Meson to build libast.

siteshwar:
You can merge it.

krader1961:
Merged as 2285957e along with ece713b to make it work on macOS and FreeBSD. Next: add the necessary Meson config files to build ksh.

siteshwar:
I will check regarding configurations for ksh.

krader1961:
@siteshwar, I'm going to assume you didn't mean to remove the *bin/package* script and reinstate it so that Travis builds still work until we can switch completely to Meson.

siteshwar:
I removed it by mistake.

PR #124, opened 2017-11-17
Meson configurations for ksh93 and libcmd

siteshwar:
I am putting up this pull request for initial review.  This is currently work in progress and should not be merged until finished. I am able to get a working ksh binary on Linux with these changes.

EDIT: This pull request is ready to merge now.


bin/libast_prereq.sh
```diff
@@ -15,7 +15,7 @@ PATH=$bin_dir:$PATH
 # Legacy build system keeps files generated by iffe with different names under different directories
 # Each array listed below is processed separately and headers generated by iffe are kept in separate directories
 # in a way that is compatible with legacy build system
-iffe_tests=( 'api' 'aso' 'ccode' 'common' 'float' 'fs' 'iconv' 'lib' 'locale' 'mmap' 'ndbm' 'options' 'prog' 'sched' 'sfio' 'siglist' 'sizeof' 'spawn' 'standards' 'sys' 'sys_calls' 'time' 'tmlib' 'tty' 'tv' 'tvlib' 'vfork' 'vmalloc' 'wait')
+iffe_tests=( 'api' 'aso' 'ccode' 'common' 'float' 'fs' 'iconv' 'lib' 'syslocale' 'mmap' 'ndbm' 'options' 'prog' 'sched' 'sfio' 'siglist' 'sizeof' 'spawn' 'standards' 'sys' 'sys_calls' 'time' 'tmlib' 'tty' 'tv' 'tvlib' 'vfork' 'vmalloc' 'wait')
```
siteshwar: Meson does not honor order of include paths and this feature test was clashing with another test with same name under ksh93 directory. So I changed it's name.

src/cmd/ksh93/bltins/typeset.c
```diff
@@ -1102,7 +1102,6 @@ int	b_builtin(int argc,char *argv[],Shbltin_t *context)
 		sh_addlib(tdata.sh,library,arg,NiL);
 	}
 	else
-#endif /* SHOPT_DYNAMIC */
```
krader1961: It's going to take me a couple of days to review this PR and test it on various platforms. But I'm confused by this particular set of changes. Why would switching from Nmake to Meson necessitate changing which blocks of code are predicated on the build time `SHOPT_DYNAMIC` symbol?

Also, it looks like this flag defaults to one (under Nmake) and there is no feature detection that sets it to zero. So it is presumably expected to be set via an environment variable to zero if the person building ksh does not want to support dynamic loading of builtins. It's not obvious we should continue to support that as a build time option. It would probably be better to just automatically enable/disable the feature based on the capabilities of the platform. And today it's likely to always be enabled for any platform we're targeting.

siteshwar: By default ksh compiles with ```SHOPT_DYNAMIC``` option. I believe compilation was not tested when this flag is off so the ```ifdefs``` were misplaced.

krader1961: @siteshwar, Are you saying this step in switching to Meson does not set the `SHOPT_DYNAMIC` symbol? If so this portion of the change may be perfectly reasonable. But I can't help but wonder if we shouldn't instead just set the symbol to one to mimic what happens by default under Nmake (i.e., *bin/package build*). Then defer fixing the conditional compilation logic to a separate update. Alternatively we should create a file or Github issue where we track these types of deviations from the established defaults and just do whatever makes the transition to Meson the easiest in the short term.

siteshwar: Compiling with this option has a dependency on ```libdll```. So for now I kept it ```off```. I would prefer to investigate later if we can remove dependency over ```libdll``` (or choose to compile it).

src/cmd/ksh93/meson.build
```diff
@@ -0,0 +1,37 @@
...
+ksh93_c_args += [ '-DSHOPT_PFSH' ]
+ksh93_c_args += ['-DSHOPT_KIA' ]
+ksh93_c_args += [ '-DSHOPT_COMPLETE' ]
+ksh93_c_args += [ '-DSHOPT_OPTIMIZE' ]
+ksh93_c_args += [ '-DSHOPT_RAWONLY' ]
+ksh93_c_args += [ '-D_PACKAGE_ast' ]
+ksh93_c_args += [ '-DSHOPT_FIXEDARRAY' ]
+ksh93_c_args += [ '-DSHOPT_MULTIBYTE' ]
+ksh93_c_args += [ '-DKSHELL' ]
+#ksh93_c_args += [ '-DSHOPT_DYNAMIC' ]
```
siteshwar: I have kept ```SHOPT_DYNAMIC``` off here.

src/cmd/ksh93/bltins/typeset.c
siteshwar: and "by default" I meant the legacy build system compiles it.

krader1961: > Compiling with this option has a dependency on libdll. So for now I kept it off. 

Seems reasonable. It's better to do this transition to Meson incrementally rather than try to make everything work with the first commit. But a "TODO" comment to remind us to revisit this might be a good idea.

bin/ksh93_prereq.sh
```diff
@@ -0,0 +1,43 @@
...
+for iffe_test in ${iffe_tests[@]}; do
+    iffe -v -X ast -X std -c 'cc' run $iffe_test
+done
+
+for iffe_test in ${iffe_tests_2[@]}; do
+    iffe -v -X ast -X std -c 'cc' run $iffe_test
+    cp "$base_dir/src/cmd/ksh93/features/FEATURE/$iffe_test" "$base_dir/src/cmd/ksh93/features/$iffe_test.h"
+done
+
+iffe -v -c 'cc' run "$base_dir/src/cmd/ksh93/features/math.sh" "$base_dir/cmd/ksh93/data/math.tab"
```
siteshwar: This generates an empty table with Meson. @krader1961  Any ideas about it ?

siteshwar:
I tested these changes on FreeBSD and it works fine.

bin/ksh93_prereq.sh
krader1961: I'll take a look but I'm strongly inclined to simply omit this feature test and move the *math.c* module into the main ksh source tree. Any platform we care about today supports all the math functions this module requires. If this module fails to compile and produce a usable `ksh` binary then the system has serious problems vis-a-vis providing the features ksh requires. The only reason I did it this way in the Nmake build system was to minimize the footprint of the change while still shaving a whole minute off the build time.

krader1961:
On macOS it fails like this:

```
  "_asocasptr", referenced from:
      _sh_chktrap in sh_fault.c.o
      _job_post in sh_jobs.c.o
ninja: build stopped: subcommand failed.
```

I see similar failures on FreeBSD and Ubuntu. A quick glance at the related code (and iffe tests in *src/lib/libast/features/aso*) left me grinding my teeth since it seems to fall back to the platform providing the missing function which is absurd since the function in question is unique to the AST project. TBD is what is missing in the Meson config.

bin/ksh93_prereq.sh
krader1961: My opinion is that *src/cmd/ksh93/bltins/meson.build* should include `'bltins/math.c'` and the relevant `iffe` test in *bin/ksh93_prereq.sh* should be removed. Then `cp src/cmd/ksh93/data/math.c src/cmd/ksh93/bltins/`. As I said in my previous comment any platform ksh currently builds on provides all the necessary math functions so the *math.c* module should "just work" and thus there isn't any reason to perform the `iffe` dance for those functions. And if that isn't true then all the people who have argued with me that `ksh` should support math functions like `scalb()` even though ksh provides no fallback if the platform does not provide the function need to explain why we should support those platforms.

siteshwar:
On FreeBSD I used ```-D b_lundef=false``` option for Meson and I did not get any errors.

bin/ksh93_prereq.sh
siteshwar: I have manually copied it under ```src/cmd/ksh93/bltins/``` and added it to meson build. Also I left a todo note on the original file for removal.

src/cmd/ksh93/bltins/typeset.c
siteshwar: I have added TODO notes.

krader1961:
> On FreeBSD I used -D b_lundef=false option for Meson and I did not get any errors.

Me too. I thought this might have something to do with all the cleanup commits I made. So I applied this on top of your last commit (d11bdae( to reinstate *bin/package*. I'm still getting the failure about `_asocasptr` being undefined when doing the final link of the ksh binary (where the `b_lundef` flag doesn't matter). Also, I could have sworn this PR allowed me to build the ksh binary. But then this error started occurring. It's stemming from this warning:

```
../src/cmd/ksh93/sh/fault.c:489:13: warning: implicit declaration of function 'asocasptr' is invalid in C99 [-Wimplicit-function-declaration]
                                        while (asocasptr(&shp->siginfo[sig],ip,0)!=ip);
                                               ^
```

TBD is why the *aso.h* header isn't being included when that module is compiled on my systems.

krader1961:
It would be nice to understand why the *aso.h* header is getting implicitly included on your systems, @siteshwar, but not on mine. Regardless, this is a IWYU situation and is fixed by the following patch. Too, all of the `#include <...>` statements that are actually including project headers need to be changed to the double-quote form. I'm going to see if that breaks the Nmake build and if it fixes the implicit *aso.h* include in Meson.

```
diff --git a/src/cmd/ksh93/sh/fault.c b/src/cmd/ksh93/sh/fault.c
index 0a62d1d..c48fa21 100644
--- a/src/cmd/ksh93/sh/fault.c
+++ b/src/cmd/ksh93/sh/fault.c
@@ -27,6 +27,7 @@
  */

 #include       "defs.h"
+#include       "aso.h"
 #include       <fcin.h>
 #include       "io.h"
 #include       "history.h"
diff --git a/src/cmd/ksh93/sh/jobs.c b/src/cmd/ksh93/sh/jobs.c
index ecc59b9..e5ac3f4 100644
--- a/src/cmd/ksh93/sh/jobs.c
+++ b/src/cmd/ksh93/sh/jobs.c
@@ -30,6 +30,7 @@
  */

 #include       "defs.h"
+#include       "aso.h"
 #include       <wait.h>
 #include       "io.h"
 #include       "jobs.h"
```

siteshwar:
I started getting same errors after rebasing with your latest changes. It looks like some commit from your recent clean up is causing this behavior. There are strange dependencies across sources, so we need to be careful while removing any code. It can also cause test cases to fail (for e.g. I still have not merged  https://github.com/att/ast/pull/111).

krader1961:
Understood. That's why I didn't remove *cmd/std* or *cmd/coshell* as they seem to be needed by some unit tests on some platforms. And, of course, *cmd/cpp* and *lib/libpp* can't be removed until we no longer need to support `bin/packge`. Still, there is absolutely nothing in the code I dropped that should be required to build/test `ast-ksh` and isn't for the Nmake based build/test. Still, it's a best practice for every .h and .c module to explicitly include each header it needs rather than relying on transitive includes.

krader1961:
Given the recent discussion about symbols like `_asocasptr` not being defined I took a closer look at the code. It's a @#$(*&!@$# mess. Consider this block of code in *src/cmd/ksh93/include/jobs.h*:

```
#if     _hdr_aso
#   include     "aso.h"
#else
#   define ASO_LOCK     1
#   define ASO_UNLOCK   2
#   define asolock(a,b,c)
#endif /* _hdr_aso */
```

Note that *jobs.h* then goes on to use `asoincint()` and several other related functions that would only be available if the *aso.h* header was available. Clearly there is a lot of ancient baggage that is no longer relevant and which needs to be ruthlessly removed because the assumptions are no longer valid.

In the same vein I just noticed *src/cmd/kshlib/codex/* relies on *src/lib/libcodex* that was recently removed in commit 116ba04 by @siteshwar. I do not think we should reinstate that code. In fact, had @siteshwar not removed that library I would have done so. And I subsequently removed several other AST libraries that are "optional" to building and testing ksh but someone might expect to be present.

The only google reference I could find to the codex code was this [blog article](https://blog.fpmurphy.com/2010/08/ksh93-hash-builtins.html) from 2010. I don't see how it is possible for anyone to create dynamically loaded builtin commands given the state of the AST (i.e., ksh) project unless they have downloaded this project and built their plugin against that source.

siteshwar:
@krader1961 From my side this pull request is ready to merge. If you have fixed the issues around linker errors, feel free to merge it.

siteshwar:
Also, I would like to start running test cases with Meson builds asap. If some test cases break when we remove code, we should be able to catch it early.

krader1961:
Merged as commit c3de4d7 on top of my trivial change to add the necessary `#include "aso.h"` to to ksh93 modules. Mostly thanks to @siteshwar we can now build `ksh` using Meson. Hurrah!

And, yes, the obvious next step is to get the unit tests working under Meson. Then add Meson to the Travis config. I am looking forward to being able to start doing things like running the code through clang-format, oclint, cppcheck, and iwyu.

siteshwar:
I don't see the updated code in ```beta``` branch. Did you forget to push it ?

krader1961:
Sort of. I was waiting for Travis to confirm this didn't break anything in my fork and got side-tracked :smile: You should see it now in this repo.

PR #126, opened 2017-11-20
Meson configurations for libcoshell

siteshwar:
Added configurations for compiling libcoshell with meson and enabled SHOPT_COSHELL


krader1961:
LGTM. Builds on macOS, FreeBSD, Ubuntu and `ldd` (`otool -L` on macOS) shows that it is linked into the ksh binary.

PR #128, opened 2017-11-20
Enable running test cases for meson builds

siteshwar:
Execute following command from build directory :
```
meson test -t 6 -v
```
Default timeout in Meson is 30 seconds. Some of the tests take lot of time to execute, so set ```-t``` (timeout multiplier) to 6 (180 seconds) or more.

### DIFF
--- a/bin/run_tests.sh
+++ b/bin/run_tests.sh
@@ -1,0 +1,12 @@
+# This script is used to run tests in Meson
+script_path=`realpath "$0"`
+bin_dir=`dirname "$script_path"`
+base_dir=`dirname "$bin_dir"`
+
+export SHELL="$base_dir/build/src/cmd/ksh93/ksh"
+cd "$base_dir/src/cmd/ksh93/tests"
+
+# shtests executes tests in 3 different modes: posix, utf8 and shcomp
+# by default, all 3 types are on.
+# TODO: Skip shcomp tests for now and enable them later
+"$SHELL" ./shtests --posix --utf8
--- a/src/cmd/ksh93/tests/basic.sh
+++ b/src/cmd/ksh93/tests/basic.sh
@@ -183,20 +183,22 @@ x=$( (/bin/echo foo) ; (/bin/echo bar) )
 if	[[ $x != $'foo\nbar' ]]
 then	err_exit " ( (/bin/echo);(/bin/echo bar ) failed"
 fi
-cat > $tmp/script <<\!
-if	[[ -p /dev/fd/0 ]]
-then	builtin cat
-	cat - > /dev/null
-	[[ -p /dev/fd/0 ]] && print ok
-else	print no
-fi
-!
-chmod +x $tmp/script
-case $( (print) | $tmp/script;:) in
-ok)	;;
-no)	err_exit "[[ -p /dev/fd/0 ]] fails for standard input pipe" ;;
-*)	err_exit "builtin replaces standard input pipe" ;;
-esac
+
+echo 'TODO: Skipping test "builtin replaces standard input pipe"'
+#cat > $tmp/script <<\!
+#if	[[ -p /dev/fd/0 ]]
+#then	builtin cat
+#	cat - > /dev/null
+#	[[ -p /dev/fd/0 ]] && print ok
+#else	print no
+#fi
+#!
+#chmod +x $tmp/script
+#case $( (print) | $tmp/script;:) in
+#ok)	;;
+#no)	err_exit "[[ -p /dev/fd/0 ]] fails for standard input pipe" ;;
+#*)	err_exit "builtin replaces standard input pipe" ;;
+#esac
 print 'print $0' > $tmp/script
 print ". $tmp/script" > $tmp/scriptx
 chmod +x $tmp/scriptx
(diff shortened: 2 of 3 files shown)


siteshwar:
There are many test cases failing with the Meson build. We have to fix them one by one separately.

siteshwar:
There were merge conflicts in this branch after I merged https://github.com/att/ast/pull/126. I have fixed it now.

bin/run_tests.sh
```diff
@@ -0,0 +1,10 @@
+# This script is used to run tests in Meson
+script_path=`realpath "$0"`
+bin_dir=`dirname "$script_path"`
+base_dir=`dirname "$bin_dir"`
+
+export SHELL="$base_dir/build/src/cmd/ksh93/ksh"
+cd "$base_dir/src/cmd/ksh93/tests"
+
+# Skip shcomp tests for now
```
krader1961: It's not clear what this comment means. There is a *./src/cmd/ksh93/tests/shtests* script but no corresponding *shcomp* script. In any event it's generally better to write such coments as TODO's; e.g., `TODO: Add ./shcomp after we've implemented xyz.`. That way we can `grep TODO` to find things that need to be done.

bin/run_tests.sh
```diff
@@ -0,0 +1,10 @@
...
+"$base_dir/build/src/cmd/ksh93/ksh" ./shtests --posix --utf8
```
krader1961: I would use the `SHELL` var you exported above to help ensure the paths don't get out of sync.

src/cmd/ksh93/tests/basic.sh
```diff
@@ -536,7 +538,9 @@ float sec=SECONDS
 (( (SECONDS-sec) < .7 ))  && err_exit '. script does not restore output redirection with eval'
 
 file=$tmp/foobar
-builtin cat
+echo "TODO: Skipping call to builtin cat"
```
krader1961: The existing Nmake based build/test enables the builtin `mktemp` command which isn't present in the Meson built `ksh`. That leads to most of the tests failing with this error because they don't pass an explicit template and the BSD version of the command doesn't provide a default template:

```
mktemp: option requires an argument -- t
```

We could change all the test scripts by replacing `mktemp -dt` with `mktemp -dt tmp.XXXXXXXXXX` but it's probably better to figure out why we're not getting the builtin versions of these external commands aren't available.

bin/run_tests.sh
siteshwar: ```shcomp``` tests are executed from within ```shtests``` script. I will update the comment to make it more clear.

src/cmd/ksh93/tests/basic.sh
siteshwar: The builtins are present under ```src/lib/libcmd``` and I have not started compiling them yet. We should investigate which one of the builtins we should keep supporting.

krader1961: FYI, I edited all of the test scripts and made the aforementioned change to the `mktemp` invocations. That allows all of the tests to run. On second thought I think we should do that. We can always add explicit tests of the builtin `mktemp` if they don't already exist. I don't see any reason for the test harness to depend on its behavior since simply providing an explicit template works on BSD and Linux.

krader1961: > We should investigate which one of the builtins we should keep supporting.

There are people watching this project which feel quite strongly that all existing behavior _must be retained_. See, for example, the discussion about the obscure math functions that I will bet $100 have never been used in a ksh program. They will no doubt argue that all of the existing builtin versions of commands (that are externals in every other shell) must remain available as builtins. I'm ambivalent about it.

Consider the `realpath` command. The BSD and GNU versions are radically different except for their default behavior (i.e., when invoked with no flags). And some platforms supported by fish-shell don't have that command. So what the fish project decided to do was detect at run time whether a `grealpath` (in case the GNU version was installed under that name) exists and alias it to `realpath` if it does, if not does `realpath` exist, and if not then fallback to a fish builtin that takes no flags and implements only the default behavior of the external command (had it been found).

In general I think it is really dangerous to automatically expose a builtin with the same name as a widely available  external command such as `cat` or `mktemp`. But this is a discussion we should have in an issue rather than a PR.

krader1961:
BTW, One of the things we need to do is ensure that changes of this nature only modify the build/test directory tree. At the moment this change, and some related changes, modify the *src/cmd/ksh93/* tree.  I don't think those unwanted side-effects should keep us from merging this PR which is why I'm going to go ahead and merge it along with my change to deal with the BSD `mktemp -dt` issue. But we'll eventually want to find ways to not modify the non-build source tree when doing things like building and testing the code.

src/cmd/ksh93/tests/basic.sh
jelmd: Please read the `builtin` documentation - at least what `builtin --man` shows. So binding a builtin command to SH_CMDLIB_DIR does not cause any collisions or unexpected behavior. If the user has $SH_CMDLIB_DIR (which is per default `/opt/ast/bin`) at the beginning of its PATH, he is asking exactly for this builtin, and exactly this enhanced/platform compatible/... behavior is expected. If not, it gets not used at all. So there is no need to kill builtins like cat or realpath.

siteshwar:
@krader1961 Can you push your changes ? I want to start cleaning up the tests.

krader1961:
Merged as 024aa567 along with f53308f9 to convert the `mktemp -dt` commands.

PR #130, opened 2017-11-21
Enable running tests in parallel with Meson

siteshwar:
To run all the tests:

```
meson test -t 6
```

To run a specific test case, execute with it's name:

```
meson test alias
```


src/cmd/ksh93/tests/path.sh
```diff
@@ -265,11 +265,11 @@ status=$($SHELL -c $'trap \'print $?\' ERR;/dev/null 2> /dev/null')
 
 # universe via PATH
 
-builtin getconf
-getconf UNIVERSE - att # override sticky default 'UNIVERSE = foo'
+#getconf
+#getconf UNIVERSE - att # override sticky default 'UNIVERSE = foo'
```
krader1961: This should probably just be removed as I don't believe there is any dual universe OS still in widespread use we care about.

src/cmd/ksh93/meson.build
```diff
@@ -42,7 +42,34 @@ m_dep = cc.find_library('m', required : false)
...
+#test('ksh93 tests', ksh93_exe, args: ['../bin/run_tests.sh'])
+
+test_dir = '../src/cmd/ksh93/tests/'
+
+# This variable is used by some tests while executing subtests
+shell_var = 'SHELL=' + ksh93_exe.full_path()
+
+# These are the default locales used by legacy test script
+# '' is POSIX locale
+locales = [ '', 'C.UTF-8' ]
```
krader1961: I'd prefer to be explicit and use `'POSIX'` or `'C'` but I don't know what the AST locale code does with those values (yet another reason to use the libc implementation for these types of things).

src/cmd/ksh93/meson.build
```diff
@@ -42,7 +42,34 @@ m_dep = cc.find_library('m', required : false)
...
+
+# Skipping builtin_poll.sh
+all_tests = [ 'alias.sh', 'append.sh', 'arith.sh', 'arrays.sh', 'arrays2.sh', 'attributes.sh', 'basic.sh', 'bracket.sh', 'builtins.sh', 'case.sh', 'comvar.sh', 'comvario.sh', 'coprocess.sh', 'cubetype.sh', 'directoryfd.sh', 'enum.sh', 'exit.sh', 'expand.sh', 'functions.sh', 'glob.sh', 'grep.sh', 'heredoc.sh', 'io.sh', 'leaks.sh', 'locale.sh', 'math.sh', 'nameref.sh', 'namespace.sh', 'options.sh', 'path.sh', 'pointtype.sh', 'pty.sh', 'quoting.sh', 'quoting2.sh', 'readcsv.sh', 'recttype.sh', 'restricted.sh', 'return.sh', 'select.sh', 'sh_match.sh', 'sigchld.sh', 'signal.sh', 'statics.sh', 'subshell.sh', 'substring.sh', 'tilde.sh', 'timetype.sh', 'treemove.sh', 'types.sh', 'variables.sh', 'vartree1.sh', 'vartree2.sh', 'wchar.sh' ]
+
+# The test cases are executed in parallel by default
+foreach testname: all_tests
+    test_path = join_paths(test_dir, testname)
+    # Strip out '.sh' from filename
+    # This makes it easy to execute tests by name. For e.g. meson test alias
+    # TODO: Consider tests with '.' in filename ?
```
krader1961: It doesn't look like Meson supports `rsplit()` which would be the natural way to handle this in Python. If we really want to support test file names of the form `x.y.z.sh` then this should do the trick: `testname = '.'.join(testname.split('.')[:-1])`. But I'm inclined to just disallow names of that form and require hyphens or underscores.

krader1961:
LGTM! This is just all kinds of awesome. I verified that on macOS (by running `htop`) that `meson test` was actually running the tests in parallel. Ship it.

src/cmd/ksh93/meson.build
ignatenkobrain: agree, C.UTF-8 is not in glibc upstream even.. Although Debian, openSUSE, Fedora do it downstream

siteshwar: For now I will keep it as it is and leave a todo comment there.

src/cmd/ksh93/meson.build
```diff
@@ -42,7 +42,34 @@ m_dep = cc.find_library('m', required : false)
...
 ksh93_exe = executable('ksh', ksh93_files, c_args: ksh93_c_args , include_directories: ksh93_incdir, link_with: [ libast, libcmd, libcoshell], dependencies : [ m_dep ] )
 
-# TODO: Below command run all the tests sequentially
-# We should split out the tests and check if can run them in parallel
-# This command is run from build directory
-test('ksh93 tests', ksh93_exe, args: ['../bin/run_tests.sh'])
+# TODO: Remove bin/run_tests.sh script
+#test('ksh93 tests', ksh93_exe, args: ['../bin/run_tests.sh'])
+
+test_dir = '../src/cmd/ksh93/tests/'
```
ignatenkobrain: it is better to avoid relative paths..

src/cmd/ksh93/meson.build
ignatenkobrain: sound like an RFE for meson, mind opening one?

siteshwar: I will mention in a comment that tests should not have ```.``` in filename.

src/cmd/ksh93/meson.build
siteshwar: Fixed.

siteshwar:
I believe there should not be any issues to run the tests in parallel. If we start seeing any strange test failures then we should check if it is caused by running tests in parallel.

PR #131, opened 2017-11-22
Disable shcomp tests for functions

siteshwar:


krader1961:
Merged. But note that, unrelated to this specific PR, I had to revert d542c6ebf because it was causing the heredoc tests to fail. Looks like we can't defer getting the magic builtins working under a Meson build.

PR #132, opened 2017-11-22
Cleanup builtin tests

siteshwar:
Disable tests for comparing presence of builtins for standard commands
Remove a test for universes


krader1961:
merged

Issue #125, opened 2017-11-20
Make the ksh source code conform to a consistent style


krader1961:
The style of the ksh source code is extremely inconsistent. That makes it harder for humans to correctly interpret the code. There is a reason that companies like Google have [code style guides](https://google.github.io/styleguide/cppguide.html) that every change must adhere to. Now that we can compile ksh with the Meson build tool it is time to think about reformatting the code to a consistent style. I did this in the [fish-shell project](https://github.com/fish-shell/fish-shell/issues/571) and it made a dramatic improvement to the readability of the code.

I don't care what style standard is used for the AST ksh code. I don't care if the opening brace appears on a new line by itself. I don't care (much) if tabs or spaces are used (and whether indentation is 2, 4 or 8 spaces). I do care, quite strongly, that the code is consistent in its style. An inconsistent code style increases the likelihood that people will misunderstand what the code means and make changes that have bugs. As I write this the code style is extremely inconsistent.

There is one aspect of the current code style, not explicitly addressed by style guides like [Google's](https://google.github.io/styleguide/cppguide.html), that I do feel extremely strongly about. The block that follows a conditional must be enclosed in braces unless it appears on the same line. I have found two bugs in OSS code due to this pattern:

```
if (some_condition)
    statement_A;
    statement_B;
```

That pattern is very common in the AST ksh code.

Having said that I do prefer the [Google style](https://google.github.io/styleguide/cppguide.html); albeit with four, rather than two, space indents. However, the `clang-format` command supports a wide variety of code styles so whatever the community maintaining `ksh` decides is preferable can be accommodated.

kdudka:
This would be a show stopper for downstream consumers, especially for enterprise level distros as I already commented on https://github.com/att/ast/issues/82#issuecomment-345632314.  Please do not change the code beyond what is really necessary to fix existing bugs.

melbit-dannyw:
Plus, it feels like sacrilege.  :smile:  I find the code a thing of beauty, such as things like [this](https://github.com/att/ast/blob/3f54fd611f536639ec30dd53c48e5ec1897cc7d9/src/cmd/ksh93/bltins/cflow.c#L52-L54).

siteshwar:
I would like to investigate if code can be made more consistent in a way that is easier to backport. Otherwise given the current state of ksh93, the number of patches that are backported to distros are rather less. It would be good to have an opinion from package maintainers from other distros/operating systems about it.

qbarnes:
[...]

PR #135, opened 2017-11-23
Implement config for clang-format and editors

krader1961:
This creates a config file for the clang-format tool. It also establishes
a `.editorconfig` file to help establish the proper settings for editing
code in this project.

Finally, it includes one module that was reformatted with clang-format
followed by some manual cleanups to illustrate how this affects the
code. The latter being necessary because clang-format won't do things like
add missing braces. I verified the resulting object file is unchanged
using the `objdump -d` technique.

First step to fixing #125

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 100
+
+[{Makefile,*.in}]
+indent_style = tab
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,22 @@
+# Use the Google style with these modifications and a couple of others:
+#
+# 1) Lines can be up to 100 chars long rather than 80.
+# 2) Use a four space indent rather than two spaces.
+#
+Language: Cpp
+BasedOnStyle: Google
+ColumnLimit: 100
+IndentWidth: 4
+
+# The following directives override defaults established above.
+
+# We don't want OCLint pragmas to be reformatted.
+CommentPragmas: '^!OCLINT'
+
+# Allow slightly overlong lines.
+PenaltyExcessCharacter: 10
+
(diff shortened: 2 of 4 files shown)


krader1961:
In case you haven't noticed the unit tests are extremely flakey. The Fedora test on Travis failed thusly:

```
test signal begins at 2017-11-23+03:33:37
test signal passed at 2017-11-23+03:35:03 [ 49 tests 0 errors ]
test signal(C.UTF-8) begins at 2017-11-23+03:35:03
shtests[333]: eval: line 1: 27600: Memory fault(coredump)
test signal(C.UTF-8) failed at 2017-11-23+03:36:25 with exit code 267 [ 49 tests 267 errors ]
test signal(shcomp) begins at 2017-11-23+03:36:25
test signal(shcomp) passed at 2017-11-23+03:37:49 [ 49 tests 0 errors ]
```

Yet it passes just fine on my macOS and Ubuntu systems. I restarted the failing Travis test and it subsequently passed.

Once we're using Meson to manage the build/test process and we've reformatted the code to a consistent style I'd like to see us focus on reducing the variability of the unit tests. We really need to greatly reduce the false positive test failures being seeing on Travis.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+indent_size = 4
+indent_style = space
```
siteshwar: Can we keep tabs for indentation ?

krader1961: > Can we keep tabs for indentation ?

In case anyone missed it see my [comment](https://github.com/att/ast/issues/125#issuecomment-346701165) in the issue that prompted this PR.

krader1961:
I merged this without the change to *src/cmd/ksh93/bltins/cflow.c*. I intend to start reformatting the code ASAP so that we can also start addressing all the compiler and lint tool warnings. I will also be introducing scripts in the *bin* directory to make it easy to run the source being (or was) modified through these tools. So that you can just type something like `bin/style` to run your change (whether committed or not) through `clang-format`.

PR #136, opened 2017-11-23
Remove use of register keyword

siteshwar:
Remove use of register keyword from ksh93, libcmd and libast sources

### DIFF
--- a/src/cmd/ksh93/bltins/alarm.c
+++ b/src/cmd/ksh93/bltins/alarm.c
@@ -58,7 +58,7 @@ static void	trap_timeout(void*);
  */
 static void *time_add(struct tevent *item, void *list)
 {
-	register struct tevent *tp = (struct tevent*)list;
+	struct tevent *tp = (struct tevent*)list;
 	if(!tp || item->milli < tp->milli)
 	{
 		item->next = tp;
@@ -79,9 +79,9 @@ static void *time_add(struct tevent *item, void *list)
 /*
  * delete timeout item from current given list, delete timer
  */
-static 	void *time_delete(register struct tevent *item, void *list)
+static 	void *time_delete(struct tevent *item, void *list)
 {
-	register struct tevent *tp = (struct tevent*)list;
+	struct tevent *tp = (struct tevent*)list;
 	if(item==tp)
 		list = (void*)tp->next;
 	else
@@ -111,12 +111,12 @@ static Time_t getnow(void)
 
 static void	print_alarms(void *list)
 {
-	register struct tevent *tp = (struct tevent*)list;
+	struct tevent *tp = (struct tevent*)list;
 	while(tp)
 	{
 		if(tp->timeout)
 		{
-			register char *name = nv_name(tp->node);
+			char *name = nv_name(tp->node);
 			if(tp->flags&R_FLAG)
 			{
 				double d = tp->milli;
@@ -134,7 +134,7 @@ static void	print_alarms(void *list)
 
 static void	trap_timeout(void* handle)
 {
-	register struct tevent *tp = (struct tevent*)handle;
+	struct tevent *tp = (struct tevent*)handle;
 	tp->sh->trapnote |= SH_SIGALRM;
 	if(!(tp->flags&R_FLAG))
 		tp->timeout = 0;
@@ -146,8 +146,8 @@ static void	trap_timeout(void* handle)
 
 void	sh_timetraps(Shell_t *shp)
 {
-	register struct tevent *tp, *tpnext;
-	register struct tevent *tptop;
+	struct tevent *tp, *tpnext;
+	struct tevent *tptop;
 	char	ifstable[256];
 	Fcin_t  savefc;
 	shp->trapnote &= ~SH_SIGALRM;
@@ -189,7 +189,7 @@ void	sh_timetraps(Shell_t *shp)
 static char *setdisc(Namval_t *np, const char *event, Namval_t* action, Namfun_t
  *fp)
 {
-        register struct tevent *tp = (struct tevent*)fp;
+        struct tevent *tp = (struct tevent*)fp;
 	if(!event)
 		return(action?"":(char*)ALARM);
 	if(strcmp(event,ALARM)!=0)
@@ -266,10 +266,10 @@ static const Namdisc_t alarmdisc =
 
 int	b_alarm(int argc,char *argv[],Shbltin_t *context)
 {
-	register int n,rflag=0;
-	register Namval_t *np;
-	register struct tevent *tp;
-	register Shell_t *shp = context->shp;
+	int n,rflag=0;
+	Namval_t *np;
+	struct tevent *tp;
+	Shell_t *shp = context->shp;
 	while (n = optget(argv, sh_optalarm)) switch (n)
 	{
 	    case 'r':
--- a/src/cmd/ksh93/bltins/cd_pwd.c
+++ b/src/cmd/ksh93/bltins/cd_pwd.c
@@ -41,7 +41,7 @@
 /*
  * Invalidate path name bindings to relative paths
  */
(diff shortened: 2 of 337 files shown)


siteshwar:
I believe it sometimes makes harder to check for stacktraces for crash, so it's better to do it earlier.

siteshwar:
It's causing below tests to fail on Travis :
```
test exit begins at 2017-11-23+16:11:56
	exit.sh[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
test exit failed at 2017-11-23+16:11:56 with exit code 1 [ 8 tests 1 error ]
test exit(C.UTF-8) begins at 2017-11-23+16:11:56
	exit.sh[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
test exit(C.UTF-8) failed at 2017-11-23+16:11:57 with exit code 1 [ 8 tests 1 error ]
test exit(shcomp) begins at 2017-11-23+16:11:57
	shcomp-exit.ksh[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
test exit(shcomp) failed at 2017-11-23+16:11:57 with exit code 1 [ 8 tests 1 error ]
```

siteshwar:
However it passes when I run the same test through Meson:
```
> meson test exit
ninja: Entering directory `/home/situ/upstream/ast/build'
ninja: no work to do.
1/1 exit                                    OK       0.10 s

OK:         1
FAIL:       0
SKIP:       0
TIMEOUT:    0
```

krader1961:
I'm not seeing `test exit` failures on macOS, FreeBSD, or Ubuntu with this PR applied. I haven't reviewed it yet but assuming it does nothing more than remove the `register` keyword any failures imply a serious bug with the compiler tool chain. Because it shouldn't matter if a variable is forced to remain in a register via the keyword or the compiler decided that doing so would produce faster code.

siteshwar:
I believe the test is written incorrectly, it should be :

```
$SHELL -c 'builtin -f cmd getconf; getconf --"?-version"; exit 0' >/dev/null 2>&1 || err_exit 'ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?' 

```

```
$SHELL -c 'builtin -f cmd getconf && getconf --"?-version" && exit 0' >/dev/null 2>&1 || err_exit 'ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?' 
```

So it should fail with Meson too as we don't support dynamically loading builtin libraries yet.

krader1961:
I removed the redirection from that statement and it's writing this before failing:

```
exit.sh: line 78: 13015: Memory fault(coredump)
	exit.sh[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
```

And that was with and without the change you proposed, @siteshwar. Obviously the Travis test environment is different from the environments on my servers. TBD is what that difference is. I reviewed this change and didn't see any obvious mistakes.

Removing the third of the commits, the "register keyword from ksh93", eliminates the segfault. But with the proposed rewrite of the test (`;` => `&&`) it fails this way:

```
  version         getconf (AT&T Research) 2013-12-01
	exit.sh[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
```
 I reinstated the semicolons and tested whether each of the first two commands is exiting with a non-zero status. The `builtin -f cmd getconf ` is exiting with a zero status. The `getconf --"?-version"` is exiting with a non-zero status.

I'm still digging into exactly what part of the change is causing the failure by slowly bisecting the problematic commit.

krader1961:
I've pinpointed the part of the change that results in the segfault. It's this hunk for *src/cmd/ksh93/sh/main.c*:

```
@@ -120,10 +120,10 @@ bool sh_source(Shell_t *shp, Sfio_t *iop, const char *file)

 int sh_main(int ac, char *av[], Shinit_f userinit)
 {
-       register char   *name;
-       register int    fdin;
-       register Sfio_t  *iop;
-       register Shell_t *shp;
+       char    *name;
+       int     fdin;
+       Sfio_t  *iop;
+       Shell_t *shp;
        struct stat     statb;
        int             i;      /* set for restricted shell */
        bool            rshflag;        /* set for restricted shell */
```

Reinstating only those `register` keywords "fixes" the Travis failure of the "exit" test. Pretty clearly this must be a case of using a var before it is initialized. However, it's not obvious if it is one of these vars or another var whose placement on the stack is affected by the presence or absence of the `register` keyword.

krader1961:
The mystery deepens. I wrote some code to dump a primitive backtrace. I did that since there doesn't seem to be any way to get a core dump in a Docker container on Travis. What it shows is this when the "exit" test causes a SIGSEGV:

```
0   dump_backtrace + 87
1   handle_sigsegv + 26
2   /lib64/libc.so.6(+0x37720) [0x7f3e10663720]
3   b_builtin + 1857
4   sh_exec + 7263
5   sh_exec + 18690
6   /source/arch/linux.i386-64/bin/ksh() [0x4165bb]
7   sh_main + 3349
8   main + 37
9   __libc_start_main + 234
10  _start + 42
```

Which implies the `b_builtin()` function is using the value of a stack local var before it has been initialized. Gee, why am I not surprised.

krader1961:
Sure enough, just initializing every stack local pointer var in `b_builtin()` fixes the failure:

```
diff --git a/src/cmd/ksh93/bltins/typeset.c b/src/cmd/ksh93/bltins/typeset.c
index 68f1a74..78d5fb5 100644
--- a/src/cmd/ksh93/bltins/typeset.c
+++ b/src/cmd/ksh93/bltins/typeset.c
@@ -1003,15 +1003,15 @@ Shbltin_f sh_getlib(Shell_t* shp, char* name, Pathcomp_t* pp)
  */
 int    b_builtin(int argc,char *argv[],Shbltin_t *context)
 {
-       register char *arg=0, *name;
+       register char *arg=0, *name = NULL;
        register int n, r=0, flag=0;
-       register Namval_t *np;
+       register Namval_t *np = NULL;
        long dlete=0;
        struct tdata tdata;
        Shbltin_f addr;
-       Stk_t   *stkp;
+       Stk_t   *stkp = NULL;
        void *library=0;
-       char *errmsg;
+       char *errmsg = NULL;
 #ifdef SH_PLUGIN_VERSION
        unsigned long ver;
        int list = 0;
```

I don't yet know which of those missing initializations is the culprit. I would be surprised if tools like `oclint` (or just the compiler warnings everyone has been ignoring) do not pinpoint the problem. For now I'm just going to merge that change so that we can move ahead with merging this PR.

For @jelmd and everyone else telling me not to modernize the code I would ask that you put in as much effort as I did to get to the root cause of this problem. I'm not willing to invest this much effort if I have to continually deal with inconsistently formatted, hard to read, code that takes an order of magnitude longer to build and test than it should. Debugging this issue took me three days. Not because I am unskilled at debugging these types of problems. When I left IBM I was being paid more than $100K USD a year to solve these types of problems. The problem is the current code base and build tools are awful. Which is absurd for a tool as important as `ksh`. Note that I've omitted the expletives (swear words). So if you sense that I am pissed off you probably don't understand just how pissed off I am given the pushback by some people to modernize the source code and build/test tool chain which would have made solving this problem much easier.

krader1961:
FWIW, looking at the code I think it's the uninitialized `register Namval_t *np;` symbol that is the problem. The logic in the code looks like there may be ways it is used without being initialized. Sure enough if I just remove the initialization of that var the Travis test throws a SIGSEGV.

krader1961:
I've pushed this to the `beta` branch on top of my fix (commit 96b316ab6) for the `b_builtin()` bug.

PR #137, opened 2017-11-24
Fix a crash in print builtin

siteshwar:
There are some tests crashing due to use of uninitialized value in print builtin while parsing arguments. This should fix it.

### DIFF
--- a/src/cmd/ksh93/bltins/print.c
+++ b/src/cmd/ksh93/bltins/print.c
@@ -174,6 +174,7 @@ int    b_print(int argc, char *argv[], Shbltin_t *context)
 	int sflag = 0, nflag=0, rflag=0, vflag=0;
 	Namval_t *vname=0;
 	Optdisc_t disc;
+	memset(&disc, 0, sizeof(disc));
 	disc.version = OPT_VERSION;
 	disc.infof = infof;
 	opt_info.disc = &disc;


siteshwar:
```subshell``` test was always crashing on my system. There are many such cases of use of unitialized values that we will have to fix.

**Comment by krader1961** [comment]
Which is why I want to get the project into a state where we can apply tools like `oclint` and `cppcheck` because in my experience they are very good at identifying such problems.

File: src/cmd/ksh93/bltins/print.c [hunk 1]
```diff
@@ -174,6 +174,7 @@ int    b_print(int argc, char *argv[], Shbltin_t *context)
 	int sflag = 0, nflag=0, rflag=0, vflag=0;
 	Namval_t *vname=0;
 	Optdisc_t disc;
+	memset(&disc, 0, sizeof(disc));
```
krader1961: Would brace initialization be better by being more explicit? I'm guessing the problem is that the `catalog` member isn't initialized. Since the code is explicitly setting two of the four structure members in the following statements it would probably be better to do

```
Optdisc_t disc = { OPT_VERSION, 0, NULL, infof };
```

Even better would be to use tagged initialization is we want to be so bold as to use that feature. Otherwise I would replace the `memset()` with `disc.catalog = NULL`.

krader1961: And `disc.flags = 0;`.

siteshwar: I think it's safer to use ```memset()``` just in case ```Optdisc_t``` structure gets modified.

krader1961: Okay, it's six or one-half dozen either way. Even if just to ensure each member is set to zero I prefer `Optdisc_t disc = {0};` but `memset()` is fine too. Ship it.

FWIW, While I may not live long enough to see it I'd much prefer to see constructor functions implemented and used for these types of structures. While I really dislike C++ the concept of constructors is one feature I do like.

PR #138, opened 2017-11-24
Fix tests for builtins

siteshwar:
* Fix test case for deleting a builtin
* Fix a crash while parsing arguments in source builtin
* Rename 'time' feature test under libast to 'systime'


krader1961:
Just a FYI, @siteshwar, that I'm not going to look at any of your PRs after PR #136 until we have resolved why that one is causing a SIGSEGV on Travis but not our local systems.

siteshwar:
All tests for builtins are fixed now.

src/cmd/ksh93/bltins/misc.c
```diff
@@ -218,6 +218,11 @@ int    b_dot_cmd(int n,char *argv[],Shbltin_t *context)
 	struct checkpt buff;
 	Sfio_t *iop=0;
 	short level;
+	Optdisc_t disc;
```
krader1961: It looks like the `optinit()` macro is supposed to be used to do this initialization. But this is another example where we should replace the AST code with standard functions. In this case `getopt_long()`. So I'm going to let it stand.

krader1961:
Merged.

PR #139, opened 2017-11-24
Workaround feature test for using sockets for pipes

siteshwar:
While running feature tests under meson, libast.so is not available. It should be sufficient to check if feature test for using sockets instead of real pipes only compiles.
This fixes io tests.

### DIFF
--- a/src/cmd/ksh93/features/poll
+++ b/src/cmd/ksh93/features/poll
@@ -4,7 +4,7 @@ lib	select,poll,socket
 lib	htons,htonl sys/types.h sys/socket.h netinet/in.h
 lib	getaddrinfo sys/types.h sys/socket.h netdb.h
 typ	fd_set sys/socket.h sys/select.h
-tst	pipe_socketpair note{ use socketpair() for peekable pipe() }end execute{
+tst	pipe_socketpair note{ use socketpair() for peekable pipe() }end compile{
 	#include <ast.h>
 	#include <signal.h>
 	#include <sys/types.h>
@@ -105,7 +105,6 @@ tst	socketpair_devfd -D_AST_INTERCEPT=0 note{ /proc/.../fd/N or /dev/fd/N handle
 	}
 }end
 tst	socketpair_shutdown_mode note{ fchmod() after socketpair() shutdown() }end execute{
-	#include <ast.h>
 	#include <sys/types.h>
 	#include <sys/stat.h>
 	#include <sys/socket.h>
--- a/bin/ksh93_prereq.sh
+++ b/bin/ksh93_prereq.sh
@@ -18,10 +18,16 @@ iffe_tests_2=( nvapi shellapi )
 
 iffe_tests_3=( nfsd acct execargs pstat )
 
+function cc_fun {
+    cc -D_BLD_ast -I../../../lib/libast/include/ -I../../../lib/libast/features/  "$@"
+}
+
+export -f cc_fun
+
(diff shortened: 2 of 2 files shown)


src/cmd/ksh93/features/poll
```diff
@@ -4,7 +4,7 @@ lib	select,poll,socket
 lib	htons,htonl sys/types.h sys/socket.h netinet/in.h
 lib	getaddrinfo sys/types.h sys/socket.h netdb.h
 typ	fd_set sys/socket.h sys/select.h
-tst	pipe_socketpair note{ use socketpair() for peekable pipe() }end execute{
+tst	pipe_socketpair note{ use socketpair() for peekable pipe() }end compile{
```
siteshwar: Can we assume pipes are peekable on all modern OS's ?

krader1961: What does "peekable" mean? This appears to be mostly relevant to the *src/lib/libast/sfio/sfpkrd.c* module. Specifically, the `sfpkrd()` function. Looking at the code it is clear it involves a feature only available on SysVR4 based systems (like Solaris and DYNIX/ptx) that implement the STREAMS subsystem.

I have no idea if this change is safe let alone correct. This is another example where the AST code is an impenetrable mess of support for platforms we probably no longer care about. Or at least, in the case of Solaris, might be willing to support as second class distros with possibly suboptimal behavior.

siteshwar: It is enabled on Linux so definitely it's not specific to Solaris.

krader1961: Yes, but note that it is predicated on symbol `_stream_peek`. See *src/cmd/ksh93/sh/io.c* where you'll find lines like

```
if _pipe_socketpair && !_stream_peek
```

Pipes are definitely not "peekable" on any system other than one derived from SysVR4 with support for the STREAMS subsystem. Being "peekable" means it supports `ioctl(I_PEEK)`. I can't easily tell what this test is actually trying to determine.

Note that documents like https://linux.die.net/man/3/ioctl leave me very confused as I do not believe that Linux supports STREAMS devices. At least not without 3rd-party extensions. See https://en.wikipedia.org/wiki/STREAMS.

krader1961: Given the ambiguity of the code it might be correct that `socketpair()` based pipes are "peekable" in the sense meant by ksh. But it looks to me like the ksh concept of "peekable" definitely is not true for pipes on any system other than one derived from SysVR4 which implements STREAM pipes.

krader1961: This may be a situation where we should just make the change and deal with any problems down the road. Your call, @siteshwar, whether the risk is worth it.

siteshwar: Let's just ship it.

krader1961: Fine by me. I'll merge this as soon as I verify it doesn't break building on my local systems.

krader1961: Note that my recent change to emit a backtrace upon receipt of SIGSEGV broken the *bin/package* based build on FreeBSD.  So I can't verify this change on that platform without using `meson`. I'm reluctant to figure out how to add the relevant test for `-lexecinfo` to the Nmake based build since it seems like we are really close to being able to use Meson for builds locally and on Travis.

krader1961:
Merged as commit 71ccfdaa.

PR #140, opened 2017-11-24
Fix tests for locales

siteshwar:


src/cmd/ksh93/tests/locale.sh
```diff
@@ -342,8 +342,8 @@ then	LC_ALL=en_US.UTF-8
 	x=$(export LC_ALL=C.UTF-8; printf "hello\u[20ac]\xee world")
 	LC_ALL=C.UTF-8 eval $'[[ $(print -r -- "$x") == $\'hello\\u[20ac]\\xee world\' ]]' || err_exit '%q with unicode and non-unicode not working'
 	if	[[ $(whence od) ]]
-	then	got='68 65 6c 6c 6f e2 82 ac ee 20 77 6f 72 6c 64 0a'
-		[[ $(print -r -- "$x" | od -An -tx1) == "$got" ]] || err_exit "incorrect string from printf %q"
+	then	got=' 68 65 6c 6c 6f e2 82 ac ee 20 77 6f 72 6c 64 0a' # od command prepends a space in output
```
krader1961: That's only true for GNU `od`. The BSD `od` inserts multiple leading spaces. The right thing to do is strip leading and trailing whitespace. I'll make that change before merging.

krader1961:
Merged with my addition of `sed` to eliminate surrounding whitespace so the vagaries of GNU vs BSD `od` don't matter.

jelmd:
Thought you wanna make it more efficient/faster instead of introducing more overhead, slowness. Using sed just to remove leading whitespaces is really a very poor choice. ksh's parameter expansion should be used instead.

PR #141, opened 2017-11-24
Set LD_LIBRARY_PATH variable before invoking tests

siteshwar:
Set LD_LIBRARY_PATH to load libast, libcmd and libcoshell libraries from build root.
This fixes test for options


krader1961:
Merged as commit 89e5bad7.

PR #143, opened 2017-11-25
Remove use of register keyword from libcoshell

siteshwar:


siteshwar:
I missed it while doing #136

krader1961:
Merged as commit e497a3d6f1251f0f87f723a642f3d38de411b582.

PR #146, opened 2017-11-25
Enable backtrace on receipt of SIGSEGV

### LINKED ISSUE #144: Enable backtraces on receipt of SIGSEGV
We've recently identified numerous instances of using variables before they are initialized. This often leads to a SIGSEGV being sent to the shell. Most users don't have core dumps enabled and it's not easy to arrange for core dumps in Travis CI. So write a basic stack backtrace when SIGSEGV is received. This is what allowed me to find the root cause of issue #142. And as I type this the Travis OpenSUSE environment is failing thusly:

```
test builtins(C.UTF-8) begins at 2017-11-25+17:49:43
TODO: Enable these tests when the custom builtins of external commands are available.
TODO: Skipping test - 'cd with no arguments fails if HOME is unset'. It should be fixed later.
shtests[333]: eval: line 1: 11399: Memory fault(coredump)
test builtins(C.UTF-8) failed at 2017-11-25+17:49:56 with exit code 267 [ 195 tests 267 errors ]
```

krader1961:
Resolve #144

### DIFF
--- a/src/cmd/ksh93/sh/fault.c
+++ b/src/cmd/ksh93/sh/fault.c
@@ -26,6 +26,11 @@
  *
  */
 
+#define _GNU_SOURCE 1
+#include <dlfcn.h>
+#include <execinfo.h>
+#include <stdio.h>
+
 #include	"defs.h"
 #include	"aso.h"
 #include	"fcin.h"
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/fault.c
```diff
@@ -50,6 +55,38 @@ struct Siginfo
...
+      void *callstack[128];
+      const int n_max_frames = sizeof(callstack) / sizeof(callstack[0]);
+      int n_frames = backtrace(callstack, n_max_frames);
+      char **symbols = backtrace_symbols(callstack, n_frames);
+
+      int i;
+      for (i = skip_levels; i < n_frames; i++) {
+	 Dl_info info;
+	 if (dladdr(callstack[i], &info) && info.dli_sname) {
+	       fprintf(stderr, "%-3d %s + %td\n", i - skip_levels,
```
siteshwar: Can we use ```write()``` here ?  ```printf()``` is not signal safe.

src/cmd/ksh93/sh/fault.c
```diff
@@ -119,6 +156,10 @@ void	sh_fault(int sig)
 	char		*trap;
 	struct checkpt	*pp = (struct checkpt*)shp->jmplist;
 	int	action=0;
+
+	if(sig==SIGSEGV) {
+	    dump_backtrace(100, 0);
```
siteshwar: I see that you have installed this function as a signal handler. Why we need to call it again here ?

src/cmd/ksh93/sh/fault.c
krader1961: The problem is how to get formatted output. Even `snprintf()` isn't async safe. However, it appears that in practice `snprintf()` is safe if you don't use `%f` (which is known to call `malloc()` in some implementations) or other advanced formatting options. Also, the other functions being used, e.g., `dladdr()`, aren't async safe. 

We could forego the formatting and just write the `info.dli_sname` string. But I'm hoping to find a reasonably portable way to determine line numbers from the byte offset currently displayed.

The benefits of displaying this info far outweigh the risk that any of these calls will interact badly with the context at the time of the fault. But I do agree we should eliminate the `fprintf()` since performing stdio is likely to be the function most likely to interact badly with the interrupted context.

src/cmd/ksh93/sh/fault.c
krader1961: This ensures we get a backtrace if the user has done `trap 'do_something' SEGV`. Personally I think ksh should disallow trapping a handful of signals including SIGSEGV. There is absolutely nothing ksh script can meaningfully do on receipt of that signal. SIGSEGV is fundamentally different from signals like SIGTERM or SIGUSR1.

krader1961:
I've merged this since it is going to be crucial to dealing with Travis CI failures. There is a lot of room for improvement. I'd like to find a way to translate the function byte offset into a line number. Obviously that's possible since programs like `gdb` and `lldb` do so. Whether it is practical for ksh is an open question. But at least now we'll have some idea about the function call sequence that resulted in the SIGSEGV fault.

PR #148, opened 2017-11-26
Remove feature test for options from libast

siteshwar:
and enable another test with same name in ksh93


krader1961:
LGTM and merged as commit 560de81. But it's not clear to me what was being tested.

Issue #129, opened 2017-11-21
Should ksh provide builtins for external commands like `cat` or `mkdir`?


siteshwar:
ksh93 provides number of shell builtins for standard commands under [this directory](
https://github.com/att/ast/tree/master/src/lib/libcmd). Only some of them are compiled in legacy build system by default and listed under special path ```/opt/ast/bin```:

```
/opt/ast/bin/basename
/opt/ast/bin/cat
/opt/ast/bin/chmod
/opt/ast/bin/cmp
/opt/ast/bin/cut
/opt/ast/bin/dirname
/opt/ast/bin/getconf
/opt/ast/bin/head
/opt/ast/bin/logname
/opt/ast/bin/sync
/opt/ast/bin/uname
/opt/ast/bin/wc
```

To enable these builtins you have to add ```/opt/ast/bin``` to ```$PATH```. We should investigate if we should continue to provide shell builtins for these standard commands.

krader1961:
Some food for thought....

When the [fish-shell](https://github.com/fish-shell/fish-shell) project considers adding a new builtin they carefully consider whether the builtin name might conflict with a widely used external command.

Fish also realized that not every platform provides a `realpath` command. But rather than make it a builtin that always overrides an external `realpath` command they used the function autoload mechanism to dynamically check for a) a `grealpath` command (in case someone had installed the GNU version under that name), b) a `realpath` command, and only if those aren't found  c) fallback to the builtin. This minimizes the chance of surprising the user by ensuring they get the full featured implementation provided by their OS if one is available. If not fallback to the builtin so that fish scripts can assume they can at least do `realpath /a/symlinked/path`.

Most of those commands aren't too hard to justify as being optionally available as a builtin in terms of portability and/or performance. But `sync` is a really odd command to have as a builtin. When was the last time anyone used that command? Regardless it can't be justified as a builtin on the grounds or portability or performance. So why is it included?

dannyweldon:
> Most of those commands aren't too hard to justify as being optionally
> available as a builtin in terms of portability and/or performance. But sync
> is a really odd command to have as a builtin. When was the last time anyone
> used that command? Regardless it can't be justified as a builtin on the
> grounds or portability or performance. So why is it included?

One possibility is if ksh is being used as a root shell in single user
mode, having sync built in could be useful if no other commands are
available, to allow root to get the filesystem saved for consistency
before powering off the machine.

krader1961:
[...]

PR #149, opened 2017-11-26
Disable directory fd test for mkdir/rmdir/cat builtins

siteshwar:


dannyweldon:
Please hold off on this as I have not had time to be able to respond again to #129

siteshwar:
@dannyweldon There is a todo comment in the code and I am not ruling out the possibility to reenable this test. I am just disabling it for now.

krader1961:
@dannyweldon, Don't panic. I, for one, have not reached a decision on whether all of those commands should be standard builtins. I think there are good arguments for and against. What @siteshwar and I are trying to do with these changes is just get the project into a state where we can switch to Meson for local and Travis builds. If and when it is decided that things like `cat` should continue to be standard builtins we'll ensure these tests are re-enabled and work.

dannyweldon:
Thanks.  Okay, that's fine.

krader1961:
Merged as commit d606a1314.

PR #151, opened 2017-11-26
Add dependency check for libdl

siteshwar:


src/cmd/ksh93/meson.build
```diff
@@ -41,14 +41,15 @@ cc = meson.get_compiler('c')
 # On some platforms the math functions (e.g., `sin()`) that we need are found in
 # libc, on others they are found in libm.
 m_dep = cc.find_library('m', required : false)
-# On some platforms the `backtrace()` and related functions are found in libc,
+# On some platforms the `backtrace()` and related functions are found in libdl,
```
krader1961: This is sort of true. On Linux `backtrace()` is in libc but `dladdr()` is in libdl. :smile:

I'll merge this in a couple of minutes after I verify your LD_LIBRARY_PATH change works on Travis along with a *meson.build* cleanup I layered on top of it.

krader1961:
Merged as commit 91879e1a with a tweak to the comment in the *meson.build* file.

PR #154, opened 2017-11-27
Fix a crash in command builtin

siteshwar:


krader1961:
Merged as e5ed812fb8.

PR #155, opened 2017-11-27
Set LD_LIBRARY_PATH while invoking shcomp tests

siteshwar:
This fixes shcomp test case for options


krader1961:
Merged as 9dc3073c8d.

PR #156, opened 2017-11-27
Disable shcomp tests for io, namespace and treemove

siteshwar:
I have marked them to be fixed later

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -143,7 +143,11 @@ foreach testname: all_tests
 
     # shcomp tests
     lang_var = 'LANG='
-    test(testname + '(shcomp)', ksh93_exe,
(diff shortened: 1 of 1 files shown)


siteshwar:
These tests are [already disabled](https://github.com/att/ast/blob/1c16b34680b5723c5c8dd88eca5787c0231f133f/src/cmd/ksh93/tests/shtests#L355
) for the legacy build system.


Rest of the test cases are now passing with Meson.

krader1961:
Merged as 26ab2f4 though I changed it from a series of "testname != literal" to "not tests_to_skip.contains(testname)".

> Rest of the test cases are now passing with Meson.

I'm still seeing lots of failures on macOS, FreeBSD and Ubuntu. However, I'm seeing failures with the Nmake build system as well on those systems. TBD is whether the Meson failures are the same as the Nmake failures.

PR #157, opened 2017-11-27
Enable meson build on travis

siteshwar:


siteshwar:
Travis seems to buffer output to stdout. While running tests under Meson, output to stderr appears before output to stdout. I have not found a way to workaround it yet.

krader1961:
LGTM. Merged as 0593baf7. Awesome work, @siteshwar. It looks like this reduces the build/test time on Travis by approximately 50%. It's still slower than it should be given what is being tested but we're moving in the right direction. The tests are still flakey. The OpenSUSE environment had five test failures while the Fedora environment had one the first time I ran this and zero the second time.

I'm going to create a PR for you to review in a few minutes that eliminates the need for the `-t 6` and displays the full Meson test log.

The Travis log is huge due to all the compiler warnings. I had to download the raw log to see the results. So I'm going to go to work on running the code through clang-format and fixing the things that are causing those warnings.

siteshwar:
I had made some minor change in this pull request with 9dc1691a710893c180fd31872725f8e2bb8045a3 (Added ```CFLAGS``` variable). I will push this change directly to ```beta``` now.

siteshwar:
Regarding the travis build failures on opensuse, I believe all of them are caused by memory corruption. I am still investigating it.

siteshwar:
I made the travis build less verbose with https://github.com/att/ast/commit/a7a0ae3ad30dcbe9d28e39fd8f0cd275b64c7a4d

Issue #147, opened 2017-11-26
Travis test failure suggests a double-free or uninitialized var bug


krader1961:
With the fix for #146  a segfault being seen on Travis OpenSUSE shows this stack:

```
test builtins(C.UTF-8) begins at 2017-11-25+22:01:50
TODO: Enable these tests when the custom builtins of external commands are available.
TODO: Skipping test - 'cd with no arguments fails if HOME is unset'. It should be fixed later.
0   dump_backtrace + 50
1   handle_sigsegv + 26
2   /lib64/libc.so.6(+0x34950) [0x7f87d6216950]
3   vmregion + 169
4   _ast_free + 198
5   /source/arch/linux.i386-64/bin/ksh() [0x57f06b]
6   sfsetbuf + 4302
7   /source/arch/linux.i386-64/bin/ksh() [0x55f996]
8   /lib64/libc.so.6(+0x37139) [0x7f87d6219139]
9   /lib64/libc.so.6(+0x37185) [0x7f87d6219185]
10  /source/arch/linux.i386-64/bin/ksh() [0x4202d7]
11  /source/arch/linux.i386-64/bin/ksh() [0x419499]
12  sh_main + 3222
13  main + 37
14  __libc_start_main + 245
15  _start + 41
shtests[333]: eval: line 1: 11383: Memory fault(coredump)
test builtins(C.UTF-8) failed at 2017-11-25+22:02:04 with exit code 267 [ 195 tests 267 errors ]
```

Just a guess but this looks like a double-free bug. Or perhaps freeing a block that wasn't malloc'd --- possibly due to an uninitialized pointer var containing a random value. It would be nice if we didn't have to fix bugs in the SFIO code but I don't think we'll be able to replace it with libc stdio in the near future.

siteshwar:
Here is a more readable stacktrace from valgrind :
```
==3403== Invalid free() / delete / delete[] / realloc()
==3403==    at 0x4C2A35C: free (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==3403==    by 0x4E82B6A: sfsetbuf (sfsetbuf.c:521)
==3403==    by 0x4E769E7: _sfcleanup (sfmode.c:104)
==3403==    by 0x5ACC138: __run_exit_handlers (in /lib64/libc-2.22.so)
==3403==    by 0x5ACC184: exit (in /lib64/libc-2.22.so)
==3403==    by 0x444EA9: sh_done (fault.c:757)
[...]

PR #160, opened 2017-11-28
Fix an invalid memory free()

siteshwar:
Resolves: #147

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -311,13 +311,11 @@ retry:
 		return(0);
 	}
 
-	memset(hp, 0, histmask*sizeof(off_t));
 	shgd->hist_ptr = hist_ptr = hp;
 	hp->histshell = (void*)shp;
 	hp->histsize = maxlines;
 	hp->histmask = histmask;
 	hp->histfp= sfnew(NIL(Sfio_t*),hp->histbuff,HIST_BSIZE,fd,SF_READ|SF_WRITE|SF_APPENDWR|SF_SHARE);
-	memset((char*)hp->histcmds,0,sizeof(off_t)*(hp->histmask+1));
 	hp->histind = 1;
 	hp->histcmds[1] = 2;
 	hp->histcnt = 2;
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -310,6 +310,8 @@ retry:
 		sh_close(fd);
 		return(0);
 	}
+
+	memset(hp, 0, histmask*sizeof(off_t));
 	shgd->hist_ptr = hist_ptr = hp;
 	hp->histshell = (void*)shp;
 	hp->histsize = maxlines;
--- a/src/lib/libast/sfio/sfsetbuf.c
+++ b/src/lib/libast/sfio/sfsetbuf.c
@@ -517,7 +517,7 @@ setbuf:
 
 	f->flags = (f->flags & ~SF_MALLOC)|sf_malloc;
 
-	if(obuf && obuf != f->data && osize > 0 && (oflags&SF_MALLOC))
+	if(obuf && (obuf != f->data) && (osize > 0) && (oflags&SF_MALLOC))
 	{	free((Void_t*)obuf);
 		obuf = NIL(uchar*);
 	}
--- a/src/cmd/ksh93/include/defs.h
+++ b/src/cmd/ksh93/include/defs.h
@@ -298,7 +298,7 @@ struct shared
 /* error exits from various parts of shell */
 #define	NIL(type)	((type)0)
 
-#define new_of(type,x)	((type*)malloc((unsigned)sizeof(type)+(x)))
+#define new_of(type,x)	((type*)calloc(sizeof(type)+(x), 1U))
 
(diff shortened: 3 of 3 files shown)


src/cmd/ksh93/edit/history.c
```diff
@@ -310,6 +310,8 @@ int  sh_histinit(void *sh_context)
 		sh_close(fd);
 		return(0);
 	}
+
+	memset(hp, 0, histmask*sizeof(off_t));
```
krader1961: Please explain to me what the preceding allocation is doing and why this is correct. The *src/cmd/ksh93/include/defs.h* module says

```
#define new_of(type,x)  ((type*)malloc((unsigned)sizeof(type)+(x)))
```

Note that this module has the same definition guarded by `#if !KSHELL`. No idea what that is all about. But in any case why is it subtracting from the size of the `History_t` struct:

```
if(!(hp=new_of(History_t,(--histmask)*sizeof(off_t))))
```

`histmask` is just a power of two so it shouldn't be negative which means that this use should not be negated AFAICT. There isn't a single other use of `new_of()` that passes a negative value for the second parameter.
So this memset looks correct but the preceding `new_of()` is either wrong or needs a comment explaining what is going on.

krader1961: Also, there aren't many uses of `new_of` and it just obfuscates the code. Seems to me the right thing to do is replace its uses in this function with `malloc()` and just do the addition explicitly rather than hiding it in that `new_of()` macro.

siteshwar: Yeah, the code is strange. The for loop tries to get a number that is greater than number of history items to allocate and is a multiple of 2. Not sure why 1 is being subtracted later. We can probably refactor it.

kdudka: The last member of ```History_t``` is ```off_t   histcmds[2]```, which is nullified separately by:
```
memset((char*)hp->histcmds,0,sizeof(off_t)*(hp->histmask+1));
```

If you want to nullify the metadata, it should be written like this:
```
memset(hp, 0, sizeof(History_t));
```
... or just:
```
memset(hp, 0, sizeof *hp);
```
... which is less prone to programming mistakes.

siteshwar: ```memset(hp, 0, sizeof(History_t));```
```memset(hp, 0, sizeof *hp);```

I don't think we can use ```sizeof``` operator here because the amount of memory allocated to ```hp``` is not ```sizeof(History_t)```. It is calculated separately in the code above in a ```for``` loop.

**Comment by krader1961** [comment]
> Not sure why 1 is being subtracted later.

Doh! Because of the odd anti-pattern and weird spacing I mistook the decrement for negation. There is way too many places in the code like this where it tries to do too much in a single statement.

The dynamically sized portion of the struct probably does not need to be zeroed but I think we should be pessimistic and just zero the entire region like you've done.

File: src/lib/libast/sfio/sfsetbuf.c [hunk 2]
```diff
@@ -518,7 +519,12 @@ size_t	size;	/* buffer size, -1 for default size */
 	f->flags = (f->flags & ~SF_MALLOC)|sf_malloc;
 
 	if(obuf && obuf != f->data && osize > 0 && (oflags&SF_MALLOC))
-	{	free((Void_t*)obuf);
+	{	
+		/* 
+		Invalid memory free:  https://github.com/att/ast/issues/147
+		Removing it should not cause any memory leaks
```
krader1961: It is far from clear this is true. The conditional above suggests this should only be done when in fact the buffer needs to be freed because we've replaced one buffer with another. The problem may be that `SF_MALLOC` may be set in `oflags` when in fact the buffer was not malloced. I think there is a deeper bug here than just a `free()` that isn't needed. Isn't the problem simply that this line

```
hp->histfp= sfnew(NIL(Sfio_t*),hp->histbuff,HIST_BSIZE,fd,SF_READ|SF_WRITE|SF_APPENDWR|SF_SHARE);
```

was assuming `hp->histbuff` was NULL when in fact it could contain a random value because the allocated memory wasn't zeroed? Doesn't the problem go away with the other half of this change?

krader1961:
Oops. Accidentally pressed the "close and comment" button.

siteshwar:
@krader1961 You are correct. It seems my testing was not right and just initializing the [history variable](https://github.com/siteshwar/ast/commit/c39264286acdd291cbad8c38f3f6a5dd6703f5b3) fixes the crash. In any case I have [fixed the ```if``` condition](https://github.com/att/ast/pull/160/commits/5c75820e68cdbf92d2278c6cffebc3660ff7ec57) too

siteshwar:
I still see this log in valgrind https://github.com/att/ast/issues/147#issuecomment-347521333

krader1961:
Merged with one change: you missed the other definition of `new_of()` in *src/cmd/ksh93/edit/history.c*.

siteshwar:
The other macro is used only when ```KSHELL``` is not defined, so it's never used. But it's fine if you fixed that too.

krader1961:
FWIW, I fixed it because I didn't want to go down the rat hole of figuring out why that symbol exists. For the moment I just assumed that building shcomp or something else compiled that module without that symbol. I have a long, and growing longer, list of such issues. :smile: The sooner we   prune such dead wood the better.

PR #162, opened 2017-11-29
Run more tests concurrently on Travis

siteshwar:
By default, meson runs as many tests concurrently as number of cores present on the system. We should be able to run more tests concurrently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
     - docker pull ${OS_TYPE}
 
 script:
+    # TODO: Check how to set MESON_TESTTHREADS dynamically
     - echo > build.sh "set -e;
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ script:
         ninja;
         echo ==== Running unit tests;
         ulimit -n 1024;
+        export MESON_TESTTHREADS=16;
```
krader1961: Ugh! I dislike seeing hardcoded values like this but I agree that Meson is overly pessimistic given that the unit are not even close to being CPU bound. I can't see anything in the Meson documentation that would allow us to scale this based on the actual number of cores. So go ahead and merge this with the addition of a `TODO` comment about how this should be scaled. It would probably be better is this was moved out of this specific script to a global scope but that isn't essential.

PR #163, opened 2017-11-29
Remove dependencies not required to build on Travis

siteshwar:
These were required by legacy build system when we were doing full build of ast repository. They can be safely removed now.


krader1961:
LGTM. I was going to investigate this very issue but you beat me to it. So go ahead and merge it.

Issue #158, opened 2017-11-28
Test "src/cmd/ksh93/tests/path.sh" has "not found" command errors


krader1961:
While working on the config of the Meson based Travis tests I saw the following in the test log. The test passed but there is obviously something wrong given the "not found" errors.

```
86/156 path                                    OK       3.15 s
--- command ---
SHELL='/source/build/src/cmd/ksh93/ksh' LANG='' LD_LIBRARY_PATH='/source/build/src/lib/libast:/source/build/src/lib/libcmd:/source/build/src/lib/libcoshell' /source/build/src/cmd/ksh93/ksh /source/src/cmd/ksh93/tests/path.sh
--- stderr ---
builtin: chmod: not found
/source/src/cmd/ksh93/tests/path.sh[160]: chmod: not found [No such file or directory]
builtin: /bin/getconf: not found
-------
```

krader1961:
The PR doesn't fix the problems on my macOS server. The problem is this line in *src/cmd/ksh93/tests/path.sh*:

```
for PATH in $path :$path $path: .:$path $path: $path:. $PWD::$path $PWD:.:$path $path:$PWD $path:. :$PWD
```

Notice the lack of double-quotes. If the original `PATH` or `PWD` contains spaces chaos ensues. This happens to be true on my macOS server because `PATH` contains `/Applications/VMware Fusion.app/Contents/Public`.

So when I hear people say, without irony, that ksh93 is meant to be a general purpose programming language that competes with Python and the like I shake my head in disbelief. The POSIX rules for parsing statements are a disaster.

krader1961:
Another test bogosity. The *path.sh* script does `PATH=$path` just before exiting. Why? It's completely pointless.

jelmd:
LOL - if an admin/developer here gets hit by a "whitespace in path" he gets some cleaning rags and some days to cleanup the labs - should be enough time to think over, what went wrong ... ;-)

jelmd:
Why `PATH=$path`? Because the one, who wrote it, is an experienced programmer and knows, that it is important to restore state of the target under test, so that new tests can be append w/o any trouble?

PR #168, opened 2017-11-30
Fix path to getconf in path.sh

siteshwar:


src/cmd/ksh93/tests/path.sh
```diff
@@ -247,7 +247,7 @@ y=$(whence rm)
 whence getconf > /dev/null  &&  err_exit 'getconf should not be found'
 #builtin /bin/getconf
 PATH=/bin
-PATH=$(getconf PATH)
+PATH=$(/usr/bin/getconf PATH)
```
krader1961: I've got a slightly different fix for this as part of my change for issue #158 that doesn't assume `getconf` is not in a particular directory and adds a TODO to remind us to fix this if and when getconf becomes a builtin again.

PR #171, opened 2017-11-30
Fix a bug with overlapping buffers

siteshwar:


krader1961:
Merged. I confirmed it fixes the final bug that was keeping the `path` test from passing on macOS.

PR #173, opened 2017-12-01
Drop some standard string functions from libast

siteshwar:
These functions should be provided by standard libraries


siteshwar:
There are other similar functions present under 'src/lib/libast/comp'. I will drop those functions too.

krader1961:
LGTM. I assume from your comment that you'll also be removing things like `strcasecmp()` which are certain to be available on every platform we care about.

PR #174, opened 2017-12-01
Drop some posix compatibility functions

siteshwar:


siteshwar:
I am leaving out some posix replacement functions as libast has support for ```C_EU.UTF-8``` locale and removing these functions causes related tests to fail.

krader1961:
What the hell is a `C_EU.UTF-8` locale? There is no such animal AFAIK in any distro I've ever used or sanctioned by any standard. In any case this looks reasonable to me so LGTM.

krader1961:
Sigh. I didn't actually test this change on macOS. I just did so and it fails:

```
[442/543] Linking target src/lib/libast/libast.dylib.
FAILED: src/lib/libast/libast.dylib
cc  -o src/lib/libast/libast.dylib 'src/lib/libast/ast@sha/features_conftab.c.o' 'src/lib/
  "_mkfifoat", referenced from:
      _ast_mkfifoat in port_intercept.c.o
     (maybe you meant: _ast_mkfifoat)
  "_mknodat", referenced from:
      _ast_mknodat in port_intercept.c.o
     (maybe you meant: _ast_mknodat)
```

I don't want to revert it so I'll see if there is a simple fix.

siteshwar:
It looks like these functions are not used at all, so you can just remove references to them from sources.

PR #176, opened 2017-12-01
Use valid file descriptor in a subshell

siteshwar:
When calling a function inside of command substitution and the function both opens and closes the file descriptor "3", the output generated by the function is not returned as a result of the command substitution.
    
Resolves: rhbz#1116072


src/cmd/ksh93/sh/io.c
```diff
@@ -1881,6 +1881,25 @@ void sh_vexsave(Shell_t *shp,int fn,int fd,Spawnvex_f vexfun, void *arg)
 	}
 }
 
+int safefdnumber(Shell_t* shp, int sfd)
+{
+	int fd;
+
+	while(1)
```
krader1961: This looks like a fix from the Fedora patches you linked to earlier. But it is questionable on several grounds. Such as the assertion only  fd 3 is affected. Also, this `while(1)` block looks to be bogus since ksh is not multithreaded. So the enclosed `for()` loop should be invariant. And even if it was multithreaded this mechanism should be protected by a lock rather than trying to execute the enclosed code more than once. Finally, notice the `break;` at the bottom of this `while() { }` loop which means it is never run more than once.

I also don't understand the innermost `continue;` statement. You could remove it and the code would behave the same.

I don't doubt the the idea embodied in this patch is correct but the fix is highly suspect.

src/cmd/ksh93/tests/functions.sh
```diff
@@ -1230,4 +1230,26 @@ foo()
...
+# both opens and closes the file descriptor "3", the output generated by 
+# the function is not returned as a result of the command substitution.
+function run_in_subshell
+{
+    # Create a temporary file
+    tmpfile=$(mktemp)
+    ls > $tmpfile
+
+    # Bug is not exhibited if choosing a higher-numbered file
+    # descriptor (e.g. 4).
```
krader1961: I'll bet $20 this statement is wrong. It's only true if there are any unused (i.e., closed) file descriptors between 2 and the "higher-numbered file descriptor". That's because all UNIX implementations, going back to the very first, guarantee that doing something that returns a file descriptor (e,g., calling `open()`) returns the lowest numbered unused file descriptor. So you're less likely to notice this problem if the function uses a fd other than 3.

krader1961: I'm all for incorporating fixes from other distros (e.g., Fedora) but I would like to avoid including bogus ideas like this.

src/cmd/ksh93/sh/io.c
siteshwar: You are right. I have made some fixes based on your comments.

krader1961:
This still looks wrong to me. The idea is to find the lowest number unused fd greater than or equal to some minimum (e.g., 3). But the logic of the function doesn't do that AFAICT. It looks like it works only by accident. I'm going to take a closer look at how the `filemap` array is managed.

krader1961:
As I suspected the fix is incorrect. All you have to do is add additional redirections to the test script found in https://bugzilla.redhat.com/show_bug.cgi?id=1116072. For example,

```
exec 4< $tmpfile
exec 6< $tmpfile
```

Along with the corresponding 

```
exec 4<&-
exec 6<&-
```

This illustrates why test cases are so important.

krader1961:
Closing since I've opened PR #199 with a better fix and test case.

PR #181, opened 2017-12-03
Modify regression test suite

siteshwar:
Taken from http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20100826-fixregr.patch

### DIFF
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -303,9 +303,9 @@ then	err_exit "printf '%..*s' not working"
 fi
 [[ $(printf '%q\n') == '' ]] || err_exit 'printf "%q" with missing arguments'
 # we won't get hit by the one second boundary twice, right?
-[[ $(printf '%T\n' now) == "$(date)" ]] ||
-[[ $(printf '%T\n' now) == "$(date)" ]] ||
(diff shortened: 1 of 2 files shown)


siteshwar:
I could not find reference to any bug for it.

krader1961:
Merged.

kdudka:
Any chance to make the commit messages actually useful?

The fact that you have modified regression test suite is pretty obvious from the diffstat.  It would help me (as well as anybody else investigating the change few years from now) if the commit message actually explained **why** the change was made.

siteshwar:
@kdudka I usually leave references to related bugs in the commit messages. I tried to search for reference for this change, but could not find any. That's why I left this comment https://github.com/att/ast/pull/181#issuecomment-348731341.

kdudka:
I agree that it is good to reference bug reports when available.  But even if you have a bug number to refer to, I do not think it should be used as substitution for a proper commit message.  The corresponding bug tracker might not be reachable when you need it later on.

Even for Github issues and PRs, there is no guarantee that people will be able to view them few years from now.  I am hereby asking you to make the commit messages useful on their own wherever possible.  Thanks in advance!

siteshwar:
@kdudka Ok, I will try to add more description in the commit messages in future. Thanks!

PR #182, opened 2017-12-03
Fix memory leak in associative array

siteshwar:
Fix memory leak while unsetting element of an associative array
Resolves: rhbz#1189294

### DIFF
--- a/src/cmd/ksh93/tests/leaks.sh
+++ b/src/cmd/ksh93/tests/leaks.sh
@@ -67,7 +67,7 @@ curr_vsz=$(ps -o vsz $$ | tail -n 1 | tr -d '\n')
 
 if [[ $init_vsz -lt $curr_vsz ]];
 then
-    err_exit "Memory leak on associate arrary"
+    err_exit "Memory leak on associative array"
 fi
 
 #TODO: Enable these tests when vmstate is a builtin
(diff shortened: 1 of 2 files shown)


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1189294

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-assoc-unset-leak.patch

src/cmd/ksh93/tests/leaks.sh
```diff
@@ -41,52 +40,85 @@ function test_reset
...
+#TODO: Enable these tests when vmstate is a builtin
 
-read -C stat <<< "$data"
-a=0$(vmstate --format='+%(size)u')
-for ((i=0; i < 500; i++))
-do      read -C stat <<< "$data"
-done
-b=0$(vmstate --format='+%(size)u')
-(( b > a )) && err_exit 'memory leak with read -C when using <<<'
+#builtin vmstate 2>/dev/null || exit 0
```
krader1961: Why does this test depend on the `vmstate` builtin if it is testing for memory leaks unrelated to that command? It should be possible to test for associative array leaks without resorting to unpredictable commands like `vmstate`. Including the preceding, not commented out, `ps -o vsz $$ | ...` pipeline. Regardless why not just remove this commented out code?

siteshwar: Did you mean to remove ```+#builtin vmstate 2>/dev/null || exit 0``` ? I am fine with removing it.

These are legacy tests and I am not sure why they have choosen to use ```vmstate``` builtin. Using ```ps``` command instead of ```vmstate``` sounds good to me.

krader1961:
merged

PR #183, opened 2017-12-03
Fix couple of warnings generated by covscan

siteshwar:


siteshwar:
Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-covsfix.patch

krader1961:
FYI, I'm setting up a new macOS server as my primary workstation. I decided to not do a simple restore from my backups in order to eliminate a lot of cruft that has accumulated over the years. But that has meant that a lot of things I take for granted need extra work to get setup and working. So I'm going to be slow to respond to these issues for a few days.

krader1961:
Yay! My new Mac Pro server/workstation is mostly working. By which I mean that critical services (i.e., daemons such as Apache HTTPD and MySQL) are once again working. So I should be able to start reviewing these PRs tomorrow. FWIW, a lot of the pain was because I have things like PostFix setup to relay email from local services to Gmail. Not to mention running things typical users don't use like Perceptive Automation's "Indigo" software for home automation. Getting those mechanisms working from scratch was more painful than I anticipated.

src/cmd/ksh93/sh/init.c
```diff
@@ -1382,9 +1382,11 @@ static void put_mode(Namval_t* np, const char* val, int flag, Namfun_t* nfp)
 				mode = *(double*)val;
 		}
 		else
+		{
```
krader1961: This is a good example for why I feel quite strongly that parens should never be omitted unless the single statement block is placed on the same line as the controlling condition.

krader1961:
Merged.

PR #184, opened 2017-12-03
Add test case for wrong exit status in pipe

siteshwar:
Related: rhbz#1117316


siteshwar:
We can hold this pull request until we have found fix for the problem.

PR #185, opened 2017-12-03
Fix reading file through command substituion

siteshwar:
* Fix for reading a file through command substituion when any of stdin, stdout or stderr is closed.
* Add test case for reading file in command substitution when any of stdin, stdout or stderr is closed.
 
Resolves: rhbz#1066589

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -1319,6 +1319,7 @@ int	sh_redirect(Shell_t *shp,struct ionod *iop, int flag)
 				if(flag==SH_SHOWME)
 					goto traceit;
 				fd=sh_chkopen(fname);
+				fd=sh_iomovefd(shp, fd);
 			}
 			else if(sh_isoption(shp,SH_RESTRICTED))
 				errormsg(SH_DICT,ERROR_exit(1),e_restricted,fname);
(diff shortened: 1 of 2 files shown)


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1066589

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-filecomsubst.patch

siteshwar:
The bug is private but I have provided a test case for it.

src/cmd/ksh93/sh/io.c
```diff
@@ -1319,6 +1319,7 @@ int	sh_redirect(Shell_t *shp,struct ionod *iop, int flag)
 				if(flag==SH_SHOWME)
 					goto traceit;
 				fd=sh_chkopen(fname);
+				fd=sh_iomovefd(shp, fd);
```
krader1961: This breaks the `directoryfd` test. Without this change the added unit tests still succeed. Presumably this issue has already been fixed in some other fashion. We could merge just the added unit tests. Want to take a closer look, @siteshwar?

siteshwar:
Related test fail on Linux without this patch. I will take a look.

siteshwar:
So this patch is causing regression and they were not caught earlier because the test for directoryfd were added later (with this commit c8623daccb6fd5f68a63febf357ff9e965fc14be). It looks like the file descriptor is getting closed on exec due to https://github.com/att/ast/blob/master/src/cmd/ksh93/sh/io.c#L611 so it's not available in the subshell that is required by the test.

I am going to close this pull request and will open a separate issue to track it.

PR #186, opened 2017-12-03
Fix a crash after unsetting a function

siteshwar:
Resolves: rhbz#1105138

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -3456,7 +3456,7 @@ static void sh_funct(Shell_t *shp,Namval_t *np,int argn, char *argv[],struct arg
 #endif
 	nv_putval(SH_PATHNAMENOD,shp->st.filename,NV_NOFREE);
 	shp->pipepid = pipepid;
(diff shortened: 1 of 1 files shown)


siteshwar:
This is the reproducer for the crash :
```
function ftest {
    ftest2
}

function ftest2 {
    unset -f ftest
}

ftest
```

however I could not reproduce the crash with latest master. Still opening this pull request for review, if I missed anything.

siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1105138

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-fununset.patch

siteshwar:
The bug is private but I have provided testcase for it in https://github.com/att/ast/pull/186#issuecomment-348797139 from bugzilla.

krader1961:
My reading of the code suggests that this problem can no longer happen. So while the extra condition guarding the decrement of `np->nvalue.rp->running` can't cause any problems it also should never keep the decrement from being done. So I'm going to close this as unneeded.

PR #187, opened 2017-12-03
Clarify handling of INT and QUIT signals

siteshwar:
The INT and QUIT signals for an invoked command are ignored if the
command is followed by & and the monitor option is active.

Resolves: rhbz#1019334


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1019334

krader1961:
Merged.

PR #188, opened 2017-12-04
Fix crash on login when disk is full

siteshwar:
Resolves: rhbz#1212992


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1212992

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20140801-diskfull.patch

krader1961:
merged

PR #189, opened 2017-12-04
Fix a crash when ksh is used as login shell

siteshwar:
Resolves: rhbz#1200534

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -1202,7 +1202,7 @@ tryagain:
 			if((io||argn))
 			{
 				Shbltin_t *bp=0;
-				static char *argv[1];
+				static char *argv[2];
 				int tflags = 1;
 				if(np &&  nv_isattr(np,BLT_DCL))
 					tflags |= 2;


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1200534 (private)

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20140801-arraylen.patch

siteshwar:
It is a memory corruption. I could not reproduce it but the bug says it crashes with below lines when ksh is used as login shell:

```
function dlog
{
  fc -ln -0
}
trap dlog DEBUG
>/tmp/blah
```

krader1961:
Thank you for including a link to the Redhat bugzilla issue but those are not visible to anyone without the requisite permission such as myself. Ideally the Redhat issues would be mirrored here or made publicly visible but that is hard to justify. Since I merged several of the Solaris OS patches without requiring more detail regarding the reason for the change I am in no position to ask that these (the Redhat Fedora patches) be held to a higher standard. At this point there are so many problems with the ksh code that I am willing to see any bug fixes merged if they do not have any obvious problems.

siteshwar:
I did not realize that there were other private bugzillas than the one I mentioned in this pull request. I will try to make them public or put a short description of the discussion in related pull requests.

krader1961:
Merged but with a tweak to explicitly initialize the static array. Technically it isn't necessary since static vars are supposed to be backed by zero filled pages of memory but I dislike relying on that subtlety since it is a bad habit that can cause people to omit the initialization for stack local vars where it is needed.

PR #190, opened 2017-12-04
Fix -u parameter for typeset

siteshwar:
* Correctly handle -u parameter for typeset when it is used with -x
* Add test cases for -u parameter for typeset

Resolves: rhbz#1188377


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1188377 (private)

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-xufix.patch

I am providing test case for the bug.

krader1961:
LGTM. Merged.

PR #191, opened 2017-12-04
Fix expanding multibyte characters after $1-$9

siteshwar:
Resolves: rhbz#1256495


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1256495

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-mb-after-argvar.patch

krader1961:
LGTM. Merged.

PR #193, opened 2017-12-05
Fix a double fork when ksh is used as login shell

siteshwar:
Resolves: rhbz#1217236


siteshwar:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1217236

Patch: http://pkgs.fedoraproject.org/cgit/rpms/ksh.git/tree/ksh-20120801-nohupfork.patch

krader1961:
Merged as commit 63a23132 with a fix to the function call to include the missing `shp` argument.

PR #195, opened 2017-12-06
Add install argument for all targets

siteshwar:
Enable installation for ksh, shcomp, libast, libcmd and libcoshell.


ignatenkobrain:
not sure if I like indentation, but adding install kwds is good ;)

siteshwar:
Fixed the indentation.

PR #199, opened 2017-12-08
Pick an unused fd to capture subshell output

### LINKED ISSUE #198: Closing file descriptors can affect capturing subcommand output
This was reported as [rhbz#1116072](https://bugzilla.redhat.com/show_bug.cgi?id=1116072). I'm copying it here for clarity because the original bug report (especially the test case) is wrong about several points. Also, the Redhat fix has multiple problems and only fixes the problem accidentally and even then only for the specific test case in rhbz#1116072 not other variations of that test case.

krader1961:
A subshell that manipulates file descriptors, such as redirecting fd 3
then closing it, interferes with capturing its output. Ensure we use a
file descriptor that is not in use.

Fix #198

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -1881,6 +1881,24 @@ void sh_vexsave(Shell_t *shp,int fn,int fd,Spawnvex_f vexfun, void *arg)
 	}
 }
 
+// Return the lowest numbered fd that is equal to or greater than the requested
+// `min_fd` and which is not currently in use.
+int sh_get_unused_fd(Shell_t* shp, int min_fd) {
+	int fd;
+
+	while (true) {
+	    if (fcntl(min_fd, F_GETFD) == -1) {
+		for(fd = 0; fd < shp->topfd; fd++) {
+		    if (filemap[fd].save_fd == min_fd || filemap[fd].orig_fd == min_fd) break;
+		}
+		if (fd == shp->topfd) break;
+	    }
+	    min_fd++;
+	}
+
+	return min_fd;
+}
+
 /*
  *  close all saved file descriptors
  */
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -1881,6 +1881,24 @@ void sh_vexsave(Shell_t *shp,int fn,int fd,Spawnvex_f vexfun, void *arg)
 	}
 }
 
+// Return the lowest numbered fd that is equal to or greater than the requested
```
siteshwar: Rest of the ksh code uses ```/*...*/``` type of comments.

src/cmd/ksh93/sh/io.c
```diff
@@ -1881,6 +1881,24 @@ void sh_vexsave(Shell_t *shp,int fn,int fd,Spawnvex_f vexfun, void *arg)
...
+// `min_fd` and which is not currently in use.
+int sh_get_unused_fd(Shell_t* shp, int min_fd) {
+	int fd;
+
+	while (true) {
+	    if (fcntl(min_fd, F_GETFD) == -1) {
+		for(fd = 0; fd < shp->topfd; fd++) {
```
siteshwar: Fix indentation.

siteshwar:
LGTM.

src/cmd/ksh93/sh/io.c
krader1961: The weird indentation is because of the inconsistent indentation of the code which confuses the Vim plugin I use that attempts to detect and set the indentation behavior. Once I've reformatted the code to have a consistent format this will no longer be an issue.

src/cmd/ksh93/sh/io.c
krader1961: Yep, but not for much longer :smile:

krader1961:
merged

PR #202, opened 2017-12-08
Remove iffe feature test for external commands

siteshwar:
These external commands are now detected by meson


krader1961:
Fantastic. LGTM. Merged.

PR #204, opened 2017-12-08
Remove files related to math feature tests

siteshwar:
We are not running feature tests for math related functions in the new build
system and required functions are directly compiled from system libraries.


krader1961:
LGTM. Merged.

PR #206, opened 2017-12-08
Remove feature test for some standard functions

siteshwar:
Remove feature test for some standard functions like nice, fork, fchdir etc.


krader1961:
LGTM. Merged. I had to make one tweak: remove the remove everything after the definition of the `strval()` funtion in *src/cmd/ksh93/sh/streval.c* so it would build on FreeBSD (which doesn't define a `struct exception` type). Which is fine because the `matherr()` function isn't actually used and has been deprecated for many years.

PR #208, opened 2017-12-09
Remove feature test for resource limits

siteshwar:


src/cmd/ksh93/features/rlimits
```diff
@@ -1,27 +0,0 @@
...
-lib	getrlimit,getrlimit64,ulimit,vlimit
-typ	rlim_t,rlim64_t sys/types.h sys/time.h sys/resource.h
-
-cat{
-
-#if _sys_resource
-#	include <sys/resource.h>
-#	if _lib_getrlimit64
-#		undef	getrlimit
-#		define	getrlimit	getrlimit64
```
siteshwar: I think it is safe to assume system libraries will automatically use the right function.

krader1961: Yes, that should be a safe assumption. The only time this might be needed is when cross compiling and I strongly suspect it's not needed even then. This looks like a workaround from decades ago when 64-bit architectures were bleeding edge.

krader1961:
merged

PR #211, opened 2017-12-09
Remove feature test for signals

siteshwar:


krader1961:
This change is going to require more work since functions like `sigqueue()` are not universally available even on modern UNIX OS; c.f., https://www.gnu.org/software/gnulib/manual/html_node/sigqueue.html

This fails to build on macOS but does build on FreeBSD 11, Ubuntu and OpenSolaris. Since the code already handles whether or not `sigqueue()` is available (via the `_lib_sigqueue` symbol) it should be sufficient to add a `has_function()` or `has_header_symbol ()` test to the *meson.build* file.

siteshwar:
I have added fd58daf3af536b7ff58ef9d730fe8a25eda64d70 to this pull request.

krader1961:
Fantastic. LGTM. Merged.

PR #212, opened 2017-12-09
Remove feature test for socket related functions

siteshwar:
From the iffe feature test I am keeping macros for following
features :

- Use socketpair() for peekable pipe()
    
   I think it can be safely assumed to be true for POSIX systems.
    
- fchmod() after socketpair() shutdown()
    
   I have added Meson feature test for it


jelmd:
You think or you know? Why don't you look it up?

krader1961:
It's a shame we can't block certain user IDs from commenting.

krader1961:
This causes a lot of tests to fail on macOS and FreeBSD with this type of error:

```
/Users/krader/projects/3rd-party/ast/src/cmd/ksh93/tests/alias.sh: line 83: cannot create pipe [Invalid argument]
```

I'll take a closer look to figure out why since in principal this change should be fine.

siteshwar:
It is because my assumption was wrong and ```_socketpair_shutdown_mode``` macro should be false on FreeBSD.

siteshwar:
@krader1961 It's fixed now.

src/cmd/ksh93/meson.build
```diff
@@ -106,6 +106,45 @@ if cc.has_function('sigqueue', prefix: '#include <signal.h>')
...
+	#include <sys/socket.h>
+	int main()
+	{
+		int		sfd[2];
+		struct stat	st0;
+		struct stat	st1;
+		if (socketpair(AF_UNIX, SOCK_STREAM, 0, sfd) < 0 ||
+		    shutdown(sfd[0], 1) < 0 ||
+		    shutdown(sfd[1], 0) < 0)
+			return(1);
```
krader1961: I know you just copy/pasted this code but it would be nice if some of the bogosities like the extraneous parens around the `return` expression was removed. :smile:

krader1961:
LGTM

src/cmd/ksh93/meson.build
siteshwar: :)

PR #213, opened 2017-12-10
Remove feature test for system supported options

siteshwar:
- Always enable SHELLMAGIC macro, but probably this can be removed as it
  is always true for POSIX systems.

- Added meson test for SHOPT_DEVFD.

- Explicitly disable SHOPT_PFSH as it seems it is only supported on
  Solaris.


src/cmd/ksh93/include/defs.h
```diff
@@ -573,4 +572,10 @@ static struct lconv *lp;
 
 #define _pipe_socketpair    1   /* use socketpair() for peekable pipe() */
 
+/*
+ * SHELLMAGIC is enabled if operating system supports shebang
+ * TODO: Shall we remove this macro from sources ?
```
krader1961: Yes, this should be removed. This hasn't been an issue for nearly four decades. See https://en.wikipedia.org/wiki/Shebang_(Unix). Even when I got my first hands on experience with UNIX around 1985 kernel shebang support was widespread.

src/cmd/ksh93/meson.build
```diff
@@ -22,7 +27,6 @@ ksh93_c_args = ['-DERROR_CATALOG=0',
                 '-DSHOPT_MULTIBYTE',
                 '-DSHOPT_NAMESPACE',
                 '-DSHOPT_OPTIMIZE',
-                '-DSHOPT_PFSH',
```
krader1961: I agree with removing this. Enabling features for a specific OS (as opposed to working around bugs or limitations of a specific OS) does not belong in the generic project.

krader1961:
LGTM. Do you want to remove the code the depends on `SHOPT_PFSH` or shall I?

siteshwar:
Looks like this got closed by mistake. I will have to open another one as I did a force push.

krader1961:
Yes, I did not mean to close this PR. Sorry about that.

PR #214, opened 2017-12-10
Remove feature test for some headers

siteshwar:
It either checks for standard headers or headers for systems that we do not want to support.

### DIFF
--- a/bin/ksh93_prereq.sh
+++ b/bin/ksh93_prereq.sh
@@ -16,8 +16,6 @@ iffe_tests=( options )
 
 iffe_tests_2=( nvapi shellapi )
 
-iffe_tests_3=( nfsd acct execargs pstat )
-
 function cc_fun {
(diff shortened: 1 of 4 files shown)


krader1961:
LGTM although we should just remove the `SHOPT_ACCT` symbol and the code predicated on it. Same for `SHOPT_PFSH`. Maybe open an issue to remind us to do that?

siteshwar:
I did a pull request to remove ```SHOPT_PFSH``` (#216). Will do the same for ```SHOPT_ACCT```.

siteshwar:
On a second thought, why do you want to remove support for ```SHOPT_ACCT``` ? We can probably keep it and decide about it later.

krader1961:
I suggested removing the code depending on `SHOPT_ACCT` because it should never be true after this change is merged. Unless, of course, someone forces the symbol to be true. Also, there is no legitimate reason for a shell like ksh to explicitly manipulate the "accounting" info of any given OS.

PR #215, opened 2017-12-10
Remove feature test for system supported options

siteshwar:
- SHELLMAGIC macro was enabled if operating system supports shebang.
  I have removed it from the sources.

- Added meson test for SHOPT_DEVFD.

- Explicitly disable SHOPT_PFSH as it seems it is only supported on
  Solaris.


siteshwar:
Merging as it has been already reviewed in #213

PR #216, opened 2017-12-10
Remove SHOPT_PFSH and related code from sources

siteshwar:
SHOPT_PFSH makes use of execution profiles database (exec_attr) which
is only supported on Solaris. So it is decided to remove it from the
sources.

See github pull request #213

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -39,17 +39,6 @@
     #include    <dlldefs.h>
 #endif
 
-#if SHOPT_PFSH 
-#   ifdef _hdr_exec_attr
-#	include	<exec_attr.h>
-#   endif
-#   if     _lib_vfork
-#	include     <ast_vfork.h>
-#   else
-#	define vfork()      fork()
-#   endif
-#endif
-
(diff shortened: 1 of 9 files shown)


src/cmd/ksh93/sh/path.c
```diff
@@ -243,7 +179,7 @@ static pid_t path_xargs(Shell_t *shp,const char *path, char *argv[],char *const
 				saveargs = 0;
 			}
 		}
-		else if(spawn && !sh_isoption(shp,SH_PFSH))
+		else if(spawn /*&& !sh_isoption(shp,SH_PFSH)*/)
```
krader1961: Why comment out this condition rather than just remove it? Other than this (and the same issue below) this LGTM.

krader1961:
LGTM although I think you should just remove the two remaining `SHOPT_PFSH` references that you commented out in *src/cmd/ksh93/sh/path.c*.

src/cmd/ksh93/sh/path.c
siteshwar: I left it there just to make it easy to debug on Solaris (if we ever have to). This code is related to forking, so if there are issues around it we can at least take a hint from the commented code.

PR #217, opened 2017-12-10
Save iffe generated api headers for ksh93

siteshwar:
These api headers are generated with each build by iffe. Macros
in the generated header are based on api version. We should save
the generated headers and find a better scheme later for using
different versions of same function.

### DIFF
--- a/src/cmd/ksh93/features/nvapi.h
+++ b/src/cmd/ksh93/features/nvapi.h
@@ -1,0 +1,32 @@
+/* : : generated from nvapi by iffe version 2013-11-14 : : */
+#ifndef _NV_API_H
+#define _NV_API_H	1
+#define _sys_types	1	/* #include <sys/types.h> ok */
+#ifndef _API_nv
+#define _API_nv	_API_shell
+#endif
+#define NVAPI(rel)	( _BLD_nv || !_API_nv || _API_nv >= rel )
+
+#ifndef _NV_API_IMPLEMENT
+#define _NV_API_IMPLEMENT	1
+
+
+#if !defined(_API_nv) && defined(_API_DEFAULT)
+#define _API_nv	_API_DEFAULT
+#endif
+
+#if NVAPI(20120720)
+#undef	nv_lastdict
+#define nv_lastdict	nv_lastdict_20120720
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/features/nvapi.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef _API_nv
+#define _API_nv	_API_shell
+#endif
+#define NVAPI(rel)	( _BLD_nv || !_API_nv || _API_nv >= rel )
+
+#ifndef _NV_API_IMPLEMENT
+#define _NV_API_IMPLEMENT	1
+
+
+#if !defined(_API_nv) && defined(_API_DEFAULT)
```
krader1961: This line is a good example why we need to rewrite this from scratch at some point. Notice that this can never be true because six lines earlier `_API_nv` is defined. I'm not asking that this be fixed now as it's better if this remains the literal output of iffe in this first step towards a simpler, lighter weight, mechanism.

src/cmd/ksh93/features/shellapi.h
```diff
@@ -0,0 +1,109 @@
...
+#undef	sh_strnum
+#define sh_strnum	sh_strnum_20120720
+#endif
+
+#if SHELLAPI(20120720)
+#undef	sh_trap
+#define sh_trap	sh_trap_20120720
+#endif
+
+#define _API_shell_MAP	"sh_addbuiltin_20120720 sh_bltin_tree_20120720 sh_eval_20120720 sh_exit_20120720 sh_fd2sfio_20120720 sh_fun_20120720 sh_funscope_20120720 sh_getscope_20120720 sh_isoption_20120720 sh_menu_20120720 sh_offoption_20120720 sh_onoption_20120720 sh_pathopen_20120720 sh_reinit_20120720 sh_run_20120720 sh_setscope_20120720 sh_strnum_20120720 sh_trap_20120720"
```
krader1961: Another bogosity: This symbol isn't used and as structured isn't useful. So much cruft, too few free hours to apply a broom :smile:

krader1961:
LGTM

PR #219, opened 2017-12-11
Remove intercepts for standard functions

siteshwar:
- Remove intercepts for standard functions

  These are interecepts for standard POSIX functions. We should directly
   use the POSIX functions.

- Do not pass O_INTERCEPT flag to POSIX functions

   It is an extension of ast and is not supported by POSIX functions.


siteshwar:
It is causing some tests to fail on FreeBSD. Apparently because how the intercept functions handled paths under ```/dev/fd```.

Simplest reproducer that works with Linux but fails under FreeBSD :

```
tmp=/tmp
print hello > $tmp/foo
redirect {fd}< $tmp
echo $(< ~{fd}/foo)
```

siteshwar:
Should be fixed through ce69c59

src/lib/libast/include/ast_intercept.h
```diff
@@ -1,247 +0,0 @@
...
-#define creat		ast_creat
-
-#undef	dup
-#define dup		ast_dup
-
-#undef	dup2
-#define dup2		ast_dup2
-
-#undef	eaccess
-#define eaccess		ast_eaccess
```
krader1961: Annoyingly macOS doesn't provide `eaccess()`. So while I agree that this intercept mechanism should be removed we do need a Meson feature test for `eaccess()` and a fallback implementation if it isn't available. Even better would be to eliminate the use of this function since it can result in security bugs. It's always better to just open the file and deal with that returning EACCESS.

siteshwar: @krader1961 There are still some tests failing on FreeBSD. Mostly because of how ```/dev/fd``` behaves differently on it. I tried mounting ```fdescfs```, but still tests fail. I believe ce69c59 should not be required. Can you take a look at it ?

krader1961: > Can you take a look at it ?

Yes but it might take a couple of days since a) I'm switching to a new phone and b) the keyboard on my primary workstation is broken.

krader1961:
FYI, I'm taking a look at this. After resolving conflicts with recent commits this appears to have failure modes similar to those without this change. Which means a lot of test failures but fewer than before. It still fails to build on macOS due to the lack of `eaccess()`. I haven't determined whether `sh_isdevfd_prefix()` is needed but do agree it should not be necessary.

krader1961:
The failure on BSD is because `ast_openat()` (which is what actually gets run when `openat()` is called prior to this change) is really just a wrapper around the libast `pathopen()` function which has a lot of logic for dealing with opening directories and other special files (e.g., sockets).

This change would be perfectly safe if the code used clear abstractions instead of things like this intercept mechanism which obscures what is actually happening. The simplest fix for now is to replace everywhere in the ksh93 code that calls `openat()` with `ast_openat()` to make it clear which function is being invoked. Long term we'll need to decide on a suitable convention for such wrappers that doesn't hide what is going on.

krader1961:
Also, note that `pathopen()` calls `ast_openat()`. The purpose of the `O_INTERCEPT` flag is to break the loop that would otherwise ensue by allowing `ast_openat()` to differentiate the `openat()` call from the `pathopen()` call. Definitely an anti-pattern.

krader1961:
So replacing the two `openat()` calls in *src/cmd/ksh93/bltins/cd_pwd.c* with equivalent calls to `pathopen()` fixes some of the test failures. However, this test still fails:

```
if      cd /dev/fd/$fd/..
then    [[ $(pwd -P) == '/' ]] || err_exit 'physical directory should be /'
else    err_exit  "cd to /dev/fd/$fd/.. failed"
fi
```

That's because the `cd` builtin depends on the `ast_chdir()` intercept that has been removed. I'm pretty unhappy about all the spooky action at a distance code we're touching.

I also noticed that the code we're removing wraps most (all?) of the syscalls in the `RESTART()` macro to handle syscalls returning -1/EINTR. I have to assume that removing that logic could result in observable differences in behavior. It also has magic such as the `LOCAL()` macro which the comment says has something to do with threads.

Sorry, @siteshwar, but we can't merge this at this time. I think we're going to have to make this change in the context of a much larger effort to refactor the code.

PR #226, opened 2017-12-12
Remove 'sched' feature test from libast

siteshwar:


siteshwar:
@krader1961 I have tested the existing pull requests related to removing feature tests on Linux and FreeBSD, so I will start merging them. If you face any issues on macOS, feel free to open an issue.

PR #228, opened 2017-12-12
Add meson feature test for checking posix_spawn()

siteshwar:
This feature test was removed with 'spawn' iffe tests, but it caused
some test cases to fail on openSUSE, so I am restoring it back as
Meson feature test


krader1961:
LGTM

PR #229, opened 2017-12-13
Remove some feature tests from libast

siteshwar:
-   Remove 'map.c' feature test from libast

    It was used to map standard POSIX functions to it's AST counterparts.

-   Remove 'syslocale' feature from libast
    
    Instead use Meson for the test

-   Remove 'tmlib' feature test from libast
    
    Keeping definition of tmlocaltime() for compatibility

-   Remove 'tvlib' feature test from libast

-   Remove 'sys_calls' feature test from libast
    
    Also remove alternate implementation for getcwd() function

-   Remove 'ndbm' feature test from libast


src/lib/libast/features/syslocale.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
krader1961: This fails on FreeBSD 11 because bash is installed as */usr/local/bin/bash*. It doesn't look like this depends on bash behavior so you could change this to `#!/bin/sh`. Alternatively, `#!/usr/bin/env bash`.

krader1961:
Other than the bash path problem with the *src/lib/libast/features/syslocale.sh* script this LGTM.

siteshwar:
Fixed and merged.

PR #230, opened 2017-12-13
Remove more libast feature tests

siteshwar:
- Remove 'sig.sh' feature test from libast
    
  Removed unnecessary definitions from the autogenerated file and saved it
  under libast include directory.

- Remove 'preroot.sh' feature test from libast
    
  Also, remove source code related to preroot support


krader1961:
LGTM

PR #231, opened 2017-12-14
Remove dependency over some iffe tests

siteshwar:


bin/libast_prereq.sh
```diff
@@ -29,7 +29,7 @@ iffe_sh_tests=()
 
 iffe_sh_tests_2=('param.sh' )
 
-iffe_sh_tests_3=('libpath.sh' )
+iffe_sh_tests_3=('')
```
krader1961: Shouldn't this just be an empty array initializer rather than containing an empty string?

src/lib/libast/meson.build
```diff
@@ -7,12 +7,20 @@ cc = meson.get_compiler('c')
...
-syslocale_output = syslocale_result.stdout()
 
 if syslocale_result.returncode() == 0
+    syslocale_output = syslocale_result.stdout()
     libast_c_args += [ '-D_locale_utf8_str=' + syslocale_output ]
 endif
 
+libpath_result = run_command('features/libpath.sh')
+
+
```
krader1961: Unneeded blank line.

bin/libast_prereq.sh
siteshwar: I fixed it in a later commit.

src/lib/libast/meson.build
```diff
@@ -21,6 +21,145 @@ if libpath_result.returncode() == 0
     libast_c_args += [ '-DCONF_LIBPATH=' + libpath_output ]
 endif
 
+
+align_bound_code = '''
```
krader1961: This is okay for now but I would be happier if big blobs of C code wasn't embedded in *meson.build* files. We should establish a pattern where these blobs are in freestanding `*.c` files that are referenced by build rules.

siteshwar: These tests are large because they are directly taken from ```iffe```. I believe they can be trimmed down in size. We can have a look at it later. I will leave a TODO comment.

PR #232, opened 2017-12-16
Remove some libast feature tests

siteshwar:


krader1961:
This fails on OpenSUSE 42.3 64-bit (the copy/paste is truncating the lines):

```
In file included from ../src/lib/libast/tm/tvtouch.c:42:0:
../src/lib/libast/tm/tvtouch.c: In function ‘tvtouch’:
../src/lib/libast/features/tv.h:22:36: error: ‘struct stat64’ has no member named ‘st_atim
 #define ST_ATIME_NSEC_GET(st) ((st)->st_atimespec.tv_nsec)
                                    ^
../src/lib/libast/tm/tvtouch.c:192:19: note: in expansion of macro ‘ST_ATIME_NSEC_GET’
   am[0].tv_usec = ST_ATIME_NSEC_GET(&st) / 1000;
                   ^
../src/lib/libast/features/tv.h:24:36: error: ‘struct stat64’ has no member named ‘st_mtim
 #define ST_MTIME_NSEC_GET(st) ((st)->st_mtimespec.tv_nsec)
                                    ^
../src/lib/libast/tm/tvtouch.c:202:19: note: in expansion of macro ‘ST_MTIME_NSEC_GET’
   am[1].tv_usec = ST_MTIME_NSEC_GET(&st) / 1000;
                   ^
```

I see the same failure on Ubuntu 16.10 64-bit.

siteshwar:
I compiled it under openSUSE Leap 42.3 on 64-bit virtual machine and it works fine.

krader1961:
The problem is I still had *src/lib/libast/features/tv.h* from previous builds and that was found in the include search path before *src/lib/libast/include/tv.h*. Removing the obsolete header fixes the build on OpenSUSE and Ubuntu. But breaks the build on macOS and FreeBSD because they name those entries differently. For example, `st_atimespec` not `st_atim`:

```
cc  -Isrc/lib/libast/ast@sha -Isrc/lib/libast -I../src/lib/libast -Isrc/lib/libast/tm -I..
../src/lib/libast/tm/tvtouch.c:192:19: error: no member named 'st_atim' in 'struct stat'
                am[0].tv_usec = ST_ATIME_NSEC_GET(&st) / 1000;
                                ^~~~~~~~~~~~~~~~~~~~~~
```

So we still need a feature test to determine which struct stat member names to use.

krader1961:
Urgh! FreeBSD 11 also uses `st_atim` rather than `st_atimespec`. Which leaves just macOS as the odd duck among the distros we've been testing on.

siteshwar:
@krader1961 I have added a commit that should fix compilation on macOS

krader1961:
LGTM. Defining and checking two `#define` symbols seems like overkill since we only need one but I don't feel strongly enough about it to ask that you change it :smile:

siteshwar:
Yeah, let's see if we can get rid of it altogether in future.

PR #233, opened 2017-12-17
Remove 'lib' feature test from libast

siteshwar:


krader1961:
There might be unusual platforms we want to support which this change might break building on (e.g., Haiku). But I'd rather remove all that code now and deal with making it build on those platforms in the future when someone asks that their favorite oddball platform be supported.

However, as usual macOS is slightly behind the times since it does not provide `pipe2()`. Also, there is no `sys/filio.h` header on Ubuntu 16.10 or OpenSUSE LEAP.

siteshwar:
Clean build works fine for me on openSUSE.

krader1961:
> Clean build works fine for me on openSUSE.

I have a script that I use to test these changes which forces a clean build. However, it does not directly modify the content of directories like the */features/* directories. I manually removed the *build* directory then did `meson; ninja` on OpenSUSE 42.3 and it still fails:

```
In file included from ../src/cmd/ksh93/include/edit.h:39:0,
                 from ../src/cmd/ksh93/sh/defs.c:27:
../src/cmd/ksh93/include/terminal.h:69:24: fatal error: sys/filio.h: No such file or directory
 # include <sys/filio.h>
                        ^
```

siteshwar:
I have added a commit to fix compilation on macOS.

PR #235, opened 2017-12-17
Remove some libast feature tests

siteshwar:


src/lib/libast/path/pathprog.c
```diff
@@ -35,8 +35,7 @@
 #include <ctype.h>
 #endif
 
-#include "FEATURE/prog"
-
+/* TODO: Is it needed ? */
```
krader1961: If this isn't needed then neither is the /proc/self stuff above.  So you'll want to add

```
if compiler.has_header('mach-o/dyld.h')
    libast_c_args += [ '-D_hdr_mach_o_dyld']
endif
```

It should be possible to omit the check for the `_NSGetExecutablePath` function since if the header exists we should be able to assume we're building on macOS and the function exists.

krader1961:
LGTM

Issue #236, opened 2017-12-17
Compilation failure: undefined symbols or functions in MacOS 10.12


catull:
Cannot build the distribution without the patches under MacOS 10.12.

The first issue is in src/cmd/ksh93/sh/io.c.
Until yesterday, accept4() was defined as accept() with 3 arguments, and the fourth one was ignored.
This builds on MacOS 10.12.

The second issue is the lack of UTIME_NOW and UTIME_OMIT.
Not sure the patch works, it merely jump over statements that use the two symbols.

krader1961:
I already fixed the `accept4()` issue with commit 25f9cf1a. The problem with the UTIME symbols isn't present on macOS 10.13. Can you double-check that on macOS 10.12 those symbols aren't defined as I will be very surprised if they aren't.

krader1961:
By the way, @catull, this project needs more people like you contributing improvements. So regardless of the outcome of this specific issue I hope you will continue to comments or changes to the code.

catull:
@krader1961 
First off, I have not modified any header files in /usr/include.
The OS is macOS 10.12.6 with the latest security patches installed just the other day.

```shell
fgrep -r UTIME_ /usr/include
```

displays ....

```log
/usr/include/php/main/php_config.h:#define HAVE_UTIME_H 1
/usr/include/php/main/php_config.h:#define HAVE_UTIME_NULL 1
/usr/include/php/Zend/zend_virtual_cwd.h:#ifdef HAVE_UTIME_H
/usr/include/tidy/platform.h:  closed first, then set UTIME_NEEDS_CLOSED_FILE to 1.
/usr/include/tidy/platform.h:#ifndef UTIME_NEEDS_CLOSED_FILE
/usr/include/tidy/platform.h:#define UTIME_NEEDS_CLOSED_FILE 1
/usr/include/tidy/platform.h:#define UTIME_NEEDS_CLOSED_FILE 0
/usr/include/time.h:_CLOCK_PROCESS_CPUTIME_ID __CLOCK_AVAILABILITY = 12,
/usr/include/time.h:#define CLOCK_PROCESS_CPUTIME_ID _CLOCK_PROCESS_CPUTIME_ID
/usr/include/time.h:_CLOCK_THREAD_CPUTIME_ID __CLOCK_AVAILABILITY = 16
/usr/include/time.h:#define CLOCK_THREAD_CPUTIME_ID _CLOCK_THREAD_CPUTIME_ID
[...]

PR #237, opened 2017-12-17
This is a partial fix for issue #236

catull:
It fixes compilation errors in
  src/cmd/ksh93/sh/io.c
and
  src/lib/libast/tm/tvtouch.c.


krader1961:
Thanks for contributing. I already fixed the `accept4()` issue and don't see the other problem your PR addresses. On macOS 10.13 those symbols are in *sys/stat.h*.

Issue #210, opened 2017-12-09
Should we remove a bunch of compile time `-D` flags?


krader1961:
I've started working on restyling the ksh source. It's amazing how many `#if SOMETHING` pragmas there are and how much harder it makes it to read the code. Since many of them are always defined (see `ksh93_c_args` in *src/cmd/ksh93/meson.build*) I propose we simply drop those symbols and removing the `#if` pragmas that reference them. It's possible a couple of them should remain so people could, theoretically, build a ksh without the feature. But it looks to me like most of them should always be enabled.

siteshwar:
Yep, I noticed it too and most of these ```#if```'s should be removed.

krader1961:
To clarify my opening comment.... I don't see anyone ever compiling it without `-DSHOPT_MULTIBYTE` for example. Or `-DSHOPT_BASH` and `-DSHOPT_KIA`. The only one that vaguely looks like it might make sense to remain optional is `-DSHOPT_COSHELL`.

I'll send an email to the ast-developers mailing list as well but if anyone lurking can explain why it should be possible to build ksh without these features that would be helpful. Alternatively, if you strongly believe these conditional features should always be included say so.

DavidMorano:
> To clarify my opening comment.... I don't see anyone ever compiling it without -DSHOPT_MULTIBYTE for
>  example. Or -DSHOPT_BASH and -DSHOPT_KIA. The only one that vaguely looks like it might make 
> sense to remain optional is -DSHOPT_COSHEL

I do not know about all of these flags in  detail (off-hand), but I have had to turn OFF more than
one of these in order to get the bugs out within the optional code (on some platforms).
Among the very first of these options often needing to be turned OFF is -DSHOPT_MULTIBYTE !
Just my experience.  In fact, I wish that *more* of the whole of KSH was more optioned so that
more of the bugs can be optioned out.  For example, the "sophisticated" command line completion
and command line multi-line display are very buggy (causing core dumps) on some platforms.
I wish they could be optioned out!

In a few cases, I added compile-time options of my own in order to get rid of some code that was
causing bad bugs.  Just to be clear: "bad" bugs are often core dumps!

siteshwar:
It would be best to have such bugs reported and fixed. Working around bugs by trimming down features is not the option.

krader1961:
[...]

PR #241, opened 2017-12-19
Enable some features from the legacy build system

krader1961:
While working on issue #210 I noticed there were some features enabled
by the legacy Nmake based build system that were not enabled in the new
Meson based build system. Since that appears to be an oversight enable
those features so that we have parity with the legacy ksh93u+ build.
These symbols include SHOPT_POLL, SHOPT_FILESCAN, and SHOPT_EDPREDICT.

Whether or not the features controlled by those symbols should be in the
next stable release is TBD.

This also refactors the Meson build config to make it more obvious which
features are shared between the ksh and shcomp build targets.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -13,31 +13,39 @@ subdir('data')
 subdir('edit')
 subdir('sh')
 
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #242, opened 2017-12-19
Remove sfio, sys and fs feature tests from libast

siteshwar:


src/lib/libast/sfio/meson.build
```diff
@@ -22,3 +22,20 @@ libast_files += [
     'sfio/sfvprintf.c', 'sfio/sfvscanf.c', 'sfio/sfwalk.c', 'sfio/sfwr.c',
     'sfio/sfwrite.c'
 ]
+
+# number of bits in pointer
+charptr_size = cc.sizeof('char*')
```
krader1961: I'd prefer this be a void ptr but that's a nit and given the verbiage in the rest of the AST code it should probably remain framed as a "char ptr" for now. Which makes me a sad puppy :frown:

src/lib/libast/include/ast_fs.h
```diff
@@ -0,0 +1,10 @@
+/* : : generated from fs by iffe version 2013-11-14 : : */
+#pragma prototyped
+#ifndef _def_fs_features
+#define _def_fs_features	1
+
+#include <sys/stat.h>
+#include <sys/statvfs.h>
+#define FS_default	"ufs"
```
krader1961: Now is a good time to change this to something like "unknown". Even better, just omit this file and change the one place in *src/lib/libast/string/fmtfs.c* it is used to just hardcode the string "unknown". It should be a can't happen situation so it doesn't really matter what the string is but pretending it's an actual file system type such as "ufs" is not good.

krader1961:
Other than my comment about the `FS_default` symbol and associated include file LGTM.

PR #243, opened 2017-12-20
Add checks for utimensat() and utimets()

### LINKED ISSUE #236: Compilation failure: undefined symbols or functions in MacOS 10.12

krader1961:
Commit a2bc8be5 removed the feature test for those functions under the
not unreasonable assumption that modern platforms provide `utimensat()`.
Unfortunately even some relatively recent platforms, such as macOS 10.12
(Sierra) which was the latest version just a few months ago, don't have
`utimensat()` or `utimets()`. So reinstate feature tests for those
functions.

Fixes #236

### DIFF
--- a/src/lib/libast/meson.build
+++ b/src/lib/libast/meson.build
@@ -164,8 +164,14 @@ if align_bound_result.returncode() == 0
     endforeach
 endif
 
-if not cc.has_function('isnanl', prefix : '#include<math.h>')
+if not cc.has_function('isnanl', prefix: '#include <math.h>')
     libast_c_args += [ '-Disnanl=isnan' ]
+endif
+
+if cc.has_function('utimensat', prefix: '#include <sys/time.h>')
+    libast_c_args += [ '-D_lib_utimensat' ]
+elif cc.has_function('utimets', prefix: '#include <sys/time.h>')
+    libast_c_args += [ '-D_lib_utime' ]
 endif
 
 if cc.has_function('sysinfo', prefix : '#include <sys/sysinfo.h>')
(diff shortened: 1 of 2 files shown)


krader1961:
@catull, Can you test this change? It will be a few days before I can setup a macOS 10.12 VM so I can't verify this fixes the build on that release. But this should fix the build for you.

src/lib/libast/meson.build
```diff
@@ -164,10 +164,16 @@ if align_bound_result.returncode() == 0
...
 
-if not cc.has_function('isnanl', prefix : '#include<math.h>')
+if not cc.has_function('isnanl', prefix: '#include <math.h>')
     libast_c_args += [ '-Disnanl=isnan' ]
 endif
 
+if cc.has_function('utimensat', prefix: '#include <sys/time.h>')
+    libast_c_args += [ '-D_lib_utimensat' ]
+elif cc.has_function('utimets', prefix: '#include <sys/time.h>')
+    libast_c_args += [ '-D_lib_utime' ]
```
catull: @krader1961  Should be
```meson
libast_c_args += [ '-D_lib_utimets' ]
```

catull:
It builds on macOS 10.12.6, with and without the patch above.
Thanks.

PR #251, opened 2017-12-22
Remove 'limits.c' feature test from libast

siteshwar:
Also, pickup value of 'OPEN_MAX' dynamically by calling astconf()
function


src/cmd/ksh93/bltins/poll.c
```diff
@@ -360,7 +360,7 @@ bool get_compound_revents(Shell_t *shp, const char *parrayname, struct pollstat
 	}
 	fd = (int)nv_getnum(np);
 	nv_close(np);
-	if ((fd < -1) || (fd > OPEN_MAX))
+	if ((fd < -1) || (fd > strtol(astconf("OPEN_MAX", NiL, NiL), NiL, 0)))
```
krader1961: Ugh! FWIW, the code as it was before your change is very questionable. If nothing else it suggests a very bad API that needs more comments. Having said that your change is okay in so far as it is equivalent and doesn't make things worse.

krader1961:
LGTM

PR #254, opened 2017-12-22
There can only be one error.h

### LINKED ISSUE #253: Remove src/lib/libast/astsa/error.h, it is a trimmed version of src/lib/libast/include/error.h
The first header file defines a portion of the constants and structs that are already defined in the second header file.

The second one is included everywhere else.

You can see it with the following experiment:

- If the first header file is removed, everything still builds, i.e. compiles and links successfully.
Even after a `ninja -C build -t clean; ninja -C build`.

- The tests produce the same failures as before, on macOS 10.12.6.

Let's see what the result is on the supported platforms.

catull:
This PR removes src/lib/libast/astsa/error.h because it adds no benefit.
Without this header file, everything still builds and produces the same diagnostics as when the file is used in the build process.

Fixes #253.


catull:
Hmm, I test fails, coprocess in one build, not in both.

krader1961:
Merged. Thx.

PR #259, opened 2017-12-23
Align CFLAGS with definition in .travis.yml

### LINKED ISSUE #258: Inconsistent CFLAGS in meson.build vs. .travis.yml
`.travis.yml` defines
```sh
export CFLAGS='-fno-strict-aliasing -Wno-unknown-pragmas -Wno-missing-braces
-Wno-unused-result -Wno-return-type -Wno-int-to-pointer-cast -Wno-parentheses
-Wno-unused -Wno-unused-but-set-variable -Wno-cpp -Wno-char-subscripts';
```

`meson.build` in the root directory defines
```meson
add_project_arguments('-D_PACKAGE_ast', language: 'c')
add_project_arguments('-DMESON_BUILD=1', language: 'c')
add_project_arguments('-Wno-unknown-pragmas', language: 'c')
add_project_arguments('-g', language: 'c')
add_project_arguments('-DHOSTTYPE="' + hosttype + '"', language: 'c')
```
The middle one should be
```meson
...
add_project_arguments('-fno-strict-aliasing', language: 'c')
add_project_arguments('-Wno-unknown-pragmas', language: 'c')
add_project_arguments('-Wno-missing-braces', language: 'c')
add_project_arguments('-Wno-unused-result', language: 'c')
add_project_arguments('-Wno-return-type', language: 'c')
add_project_arguments('-Wno-int-to-pointer-cast', language: 'c')
add_project_arguments('-Wno-parentheses', language: 'c')
add_project_arguments('-Wno-unused', language: 'c')
add_project_arguments('-Wno-unused-but-set-variable', language: 'c')
add_project_arguments('-Wno-cpp', language: 'c')
add_project_arguments('-Wno-char-subscripts', language: 'c')
...
```

What about the the -D key/value pairs that are missing in `.travis.yml`; any thoughts ?

catull:
In development we need to have the same compiler flags defined as in Continous Integration.
This PR fixes #258.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,7 @@ before_install:
 script:
     # TODO: Check how to set MESON_TESTTHREADS dynamically
     - echo > build.sh "set -e;
-        hostType=`bin/package`;
-        export CFLAGS="-D_PACKAGE_ast -DMESON_BUILD=1 -DHOSTTYPE=$hostType -fno-strict-aliasing -Wno-unknown-pragmas -Wno-missing-braces -Wno-unused-result -Wno-return-type -Wno-int-to-pointer-cast -Wno-parentheses -Wno-unused -Wno-unused-but-set-variable -Wno-cpp -Wno-char-subscripts";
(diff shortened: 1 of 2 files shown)


siteshwar:
I am not sure if we should suppress these warnings by default. I did it in the travis build to make it less verbose. Users can set same ```CFLAGS``` in on local builds to avoid these warnings.

krader1961:
Thanks for the effort but we're not going to merge this for two reasons. First, there is no need to add the `-D` flags from *meson.build* to the Travis `CFLAGS` definition. The latter augments the flags in *meson.build*. Second, we do not want to suppress warnings outside of the Travis build. Eventually, when building produces far fewer warnings, we'll want to remove those CFLAGS from the Travis config.

PR #260, opened 2017-12-23
Musl support

### LINKED ISSUE #257: ksh on Linux/musl
I'd like to run ksh with the new musl libc.

So far I found three issues:
- `sh/fault.c` uses non-standard `<execinfo.h>` unconditionally, this should be checked by iffe.
  (Just don't install any SIGSEGV handler if this is not found IMO.)
- `libast/vmalloc/malloc.c` uses non-standard mallinfo, which is not provided (not sure why the iffe test for mallinfo succeeds, does it not use `$CC`?), anyway, this function is not used, can just be `ifdef`'d out.
- `comp/conf.tab` unconditionally uses lots of macros which where in POSIX 2004 but not in more recent versions, such as REGEX_VERSION; these should be guarded by `ifdef`.

With this, ksh compiles and runs the test suite pretty well, there are some things related to collation and numeric formatting that musl doesn't implement.

I can provide a patch if you want, but it's probably easier to do for someone who is familiar with the code base.

chneukirchen:
This partially fixes #257.

An issue that remains is that the `nxt` of `src/lib/libast/features/iconv`
expands into an invalid path.


bin/libast_prereq.sh
```diff
@@ -12,6 +12,8 @@ base_dir=`dirname "$bin_dir"`
 
 PATH=$bin_dir:$PATH
 
+: ${CC:=cc}
```
krader1961: This should be unnecessary but good catch that this script didn't honor the `CC` env var.

src/cmd/ksh93/meson.build
```diff
@@ -99,6 +99,12 @@ if cc.has_function('sigqueue', prefix: '#include <signal.h>')
     shcomp_c_args += [ '-D_lib_sigqueue=1' ]
 endif
 
+if cc.has_function('backtrace', prefix: '#include <execinfo.h>',
```
krader1961: I'm going to reject this part of the PR for two reasons. First, it should be sufficient and is more efficient to just check if that header is available. Second, all but two of the `#ifdef _lib_backtrace` directives can be eliminated. If the interface isn't available it's simpler and safer to just define a dummy `dump_backtrace()` function that does nothing. Doing it that way means that if anyone ever adds an explicit call to that function (as opposed to the SIGSEGV code path) the code will still compile and run. But thanks for pointing out that we shouldn't just assume this feature is available.

bin/libast_prereq.sh
siteshwar: We are in the process of deprecating ```iffe``` tests that would require this fix, but thanks for pointing it out.

chneukirchen:
Updated according to your review, thank you.

krader1961:
Thanks for contributing, @chneukirchen, but I'm not going to merge this. The first part, which changes the *bin/libast_prereq.sh* script, doesn't work because there is no `CC` env var exported by Meson. So it fails in several ways including not generating the `lctab.c` module. Rather than explore how to get Meson to make its choice of compiler visible to that script our time is better spent removing the need for that script.

Your second part is close to what I'd like to see but not quite there and I already wrote an equivalent change. I hope this experience hasn't discouraged you from contributing more changes.

chneukirchen:
According to https://github.com/mesonbuild/meson/issues/1752 setting CC is the canonical way to change the C compiler, as it has been for plain Makefiles and autoconf etc. The first part worked well in my testing, lctab.c is generated, but it may make sense to use the default cc for this file indeed, as it's just some file reformatting.

krader1961:
The issue you linked to only shows that if you export `CC` Meson will honor it (something that isn't explicitly documented). The problem is that we don't want to require the user export that var before running `meson`. And if you don't export that var then your change breaks *bin/libast_prereq.sh*. That could be made to work by replacing the `: ${CC:=cc}` line with an explicit test and export. I'd be willing to merge that as a stop gap until we remove that script.

chneukirchen:
Sorry, I missed that bug because the iffe creates files outside build, so one actually needs `git clean` to get into a clean slate.

`$CC` is interpolated everywhere, except for the exported(!) shell function `cc_fun`, ugh. I guess we'll all be happy when this script is gone. :) I'm not sure why cc_fun is needed tho, it works here if I just inline that call as in the iffe_sh_tests below.

PR #261, opened 2017-12-24
Remove 'ccode' feature test from libast

siteshwar:
Instead use macros for character code tests


krader1961:
LGTM.

P.S., While I've worked on IBM mainframes (including doing Linux support on them) I didn't realize there are multiple EBCDIC variants. Ewww!

Issue #262, opened 2017-12-24
Test on builtins fail under macOS - investigation


catull:
# Problem

The test `builtins` fails, because the shared library libast.dylib cannot be found.

The test displays these lines, when run with `meson test -v builtins`:
```log
...
dyld: Library not loaded: @rpath/libast.dylib
  Referenced from: ast.git/build/src/cmd/ksh93/ksh
  Reason: image not found
...
```

No other test fails this way.
Which means none of the other tests require to dynamically access types/functions defined in any of the 3 shared libraries.

`otool -L ksh` reports:

```log
ksh:
	@rpath/libast.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libcmd.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libcoshell.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
	/usr/local/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
```

# Two solutions
## Solution 1 (terrible)

One way that the test succeed is ... when I copy the 3 .shared library files to /usr/local/lib.

The test succeeds whether the variables {DY}LD_LIBRARY_PATH are set or not.

## Solution 2 (much better)

The trick is to assist the running shell and the OS to find the 3 shared libraries in development.
The magic environment variable is DYLD_FALLBACK_LIBRARY_PATH, currently I am in the build folder of the source tree.

```shell
export DYLD_FALLBACK_LIBRARY_PATH=src/lib
```

# Consideration

Setting DYLD_LIBRARY_PATH should not be used outside of development.
I never set it, never have had the need to.

krader1961:
Note that we are explicitly not supporting  dynamically loaded libraries during this transition period to a new build system:

```
src/cmd/ksh93/meson.build:# TODO: Add '-DSHOPT_DYNAMIC' when we have figured out how to get Meson to
```

So while mechanisms like exporting a `DYLD_FALLBACK_LIBRARY_PATH` var might help reduce failures on macOS it's not clear we want to do so long term. Unless, of course, implementing `-DSHOPT_DYNAMIC` is not sufficient to fix this issue.

catull:
I am afraid I cannot follow.

Are you suggesting the fix above is in or out ?

The special env variable is only needed in development at the moment.
[...]

PR #263, opened 2017-12-24
All 3 tests named `builtins*` crash

catull:
The problem is that the tests dynamically need to load libast.dylib.
The provided environment variable LD_LIBRARY_PATH does not help.

Instead, use DYLD_FALLBACK_LIBRARY_PATH=src/lib and the test not only does not crash anymore, but actually succeeds.

This PR fixes issue #262.


krader1961:
Closing per the discussion in #262.

Issue #264, opened 2017-12-24
Cannot compile w/ clang on macOS 10.12.6


catull:
When using clang (/usr/bin/clang) as compiler, not a single file compiles, due to:

```
ninja: Entering directory `build'
[1/536] Compiling C object 'src/lib/libcoshell/coshell@sha/costash.c.o'.
FAILED: src/lib/libcoshell/coshell@sha/costash.c.o
clang  -Isrc/lib/libcoshell/coshell@sha -Isrc/lib/libcoshell -I../src/lib/libcoshell -I../src/lib/libcoshell/../../cmd/ksh93/features/ -I../src/lib/libcoshell/../libast/features/ -I../src/lib/libcoshell/../libast/include -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -O0 -g -D_PACKAGE_ast -DMESON_BUILD=1 -Wno-unknown-pragmas -g '-DHOSTTYPE="darwin.i386-64"' -D_BLD_coshell -MMD -MQ 'src/lib/libcoshell/coshell@sha/costash.c.o' -MF 'src/lib/libcoshell/coshell@sha/costash.c.o.d' -o 'src/lib/libcoshell/coshell@sha/costash.c.o' -c ../src/lib/libcoshell/costash.c
In file included from ../src/lib/libcoshell/costash.c:26:
In file included from ../src/lib/libcoshell/colib.h:31:
In file included from ../src/lib/libcoshell/../libast/include/ast.h:34:
In file included from ../src/lib/libcoshell/../libast/include/ast_std.h:48:
../src/lib/libcoshell/../libast/include/ast_sys.h:125:15: error: expected parameter declarator
extern size_t   strlcat(char*, const char*, size_t);
                ^
/usr/include/secure/_string.h:111:44: note: expanded from macro 'strlcat'
  __builtin___strlcat_chk (dest, src, len, __darwin_obsz (dest))
                                           ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2
                               ^
In file included from ../src/lib/libcoshell/costash.c:26:
In file included from ../src/lib/libcoshell/colib.h:31:
[...]

PR #265, opened 2017-12-24
Fix for compiling with clang on macOS 10.12.6

catull:
This PR fixes issue #264.
Now both clang and gcc build the sources.

### DIFF
--- a/src/lib/libast/include/ast_sys.h
+++ b/src/lib/libast/include/ast_sys.h
@@ -45,7 +45,6 @@
 # include <sys/types.h>
 # include <stdint.h>
 # include <inttypes.h>
-# include <string.h>
 # include <unistd.h>
 # include <limits.h>
 # include <fcntl.h>
@@ -122,13 +121,10 @@ extern int	spawnvex_add(Spawnvex_t*, intmax_t, intmax_t, Spawnvex_f, void*);
 extern int	spawnvex_apply(Spawnvex_t*, int, int);
 extern intmax_t	spawnvex_get(Spawnvex_t*, int, int);
 extern int	spawnvex_close(Spawnvex_t*);
-# if defined(__DARWIN_C_LEVEL) && defined(__clang__)
-#  undef strlcat
-#  undef strlcpy
-# endif
 extern size_t	strlcat(char*, const char*, size_t);
 extern size_t	strlcpy(char*, const char*, size_t);
 extern void	swab(const void*, void*, ssize_t);
 # undef extern
 # include <stdarg.h>
+# include <string.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/lib/libast/include/ast_sys.h
```diff
@@ -122,8 +122,10 @@ extern int	spawnvex_add(Spawnvex_t*, intmax_t, intmax_t, Spawnvex_f, void*);
 extern int	spawnvex_apply(Spawnvex_t*, int, int);
 extern intmax_t	spawnvex_get(Spawnvex_t*, int, int);
 extern int	spawnvex_close(Spawnvex_t*);
+#if !defined(__DARWIN_C_LEVEL) || !defined(__clang__)
```
krader1961: @catull, Can you explain this fix? It doesn't make sense to me. First, macOS 10.12 provides these two functions and defines `size_t`. And the declarations inside this conditional block are essentially the same as those provided by the system headers. The compiler diagnostic implies that `size_t` is either not defined or incorrectly defined.

catull: In short, because the functions are macro-expanded to
```c
extern size_t __builtin___strlcat_chk (char*, const char*, size_t, __builtin_object_size (char*, 2 > 1 ? 1 : 0));
extern size_t __builtin___strlcpy_chk (char*, const char*, size_t, __builtin_object_size (char*, 2 > 1 ? 1 : 0));
```

IMHO `size_t` is defined and is not the culprit here.
I have a few questions to you.

1. Can you compile the sources on macOS 10.13.x out of master ?
2. Why are those functions redefined if they "are essentially the same as those provided by the system headers" ?  What is the motivation ?

I have prepared a new patch, which also indents the preprocessor directives, because it is hard to follow the nesting of `#ifdef` / `#else` / `#endif`.
In the past, you could not have white-spaces before a `#` and after it, either.
Modern C compiler have supported it for a while now.
Not suggesting this is a style we should adopt, just saying the code is easier to follow.

krader1961:
> Can you compile the sources on macOS 10.13.x out of master ?

Yes on my main server. But surprisingly, since it is a nearly identical config, building fails on my Macbook Pro. The only significant difference is the macOS patch level.

> What is the motivation ?

Hard to say with certainty but it appears these functions do not exist on Linux or AIX (and probably some other distros). The code doesn't use `strlcat()` AFAICT. Long term I'd like to see `strlcpy()` replaced with `strncpy()`. I'll double-check it in a few minutes but I think your more recent commit is probably the best compromise for dealing with the fact macOS and BSD distros provide these functions. However, it's a fragile fix because if any module includes the system *string.h* header before the AST header it will still break. But for now it's probably good enough.

krader1961:
I've merged this. Thanks. But in the future please do not mix changes to the style of the code with changes to fix a specific issue.

Issue #267, opened 2017-12-26
Tests types, types(shcomp) and types(C.UTF-8) on macOS - investigation


catull:
The three tests succeed, but do not actually run the whole test script.
What worse, they produce some SIGSEV.

In fact, if the lines [137-139 of src/cmd/ksh93/tests/types.sh](https://github.com/att/ast/blob/master/src/cmd/ksh93/tests/types.sh#L137-L139) are commented out, the tests run to the end without errors on macOS 10.12.6, all three of them.

This needs more elaborate investigation.

So far, both with clang (system), clang 5.0.1 and gcc 7.1 the same stack trace is produced:
```log
0   dump_backtrace + 69
1   handle_sigsegv + 25
2   _sigtramp + 26
3   3   ???                                 0x0000001600000000 0x0 + 94489280512
4   nv_putv + 562
5   put_chtype + 82
6   nv_putv + 520
7   nv_putval + 484
8   nv_open + 3936
9   sh_setlist + 7404
10  sh_exec + 4884
11  b_dot_cmd + 1721
12  sh_funct + 665
13  sh_fun_20120720 + 1077
14  type_init + 183
15  nv_settype + 1211
16  setall + 3755
17  b_typeset + 3659
18  sh_exec + 9057
19  exfile + 3343
20  sh_main + 4125
21  main + 38
22  start + 1
1/1 types                                   OK       0.13 s

OK:         1
FAIL:       0
SKIP:       0
TIMEOUT:    0
```

The test run types alone has actually one more line, squeezed in just before the "1/1 types  OK"
```log
22  start + 1
23  23  ???                                 0x0000000000000002 0x0 + 2
1/1 types                                   OK       0.13 s
```
but that is immaterial to this investigation.

catull:
The minimal set of shell script to reproduce this behaviour is

```ksh
function err_exit {
  print -u2 -r $1: "${@:2}"
}
alias err_exit='err_exit $LINENO'
typeset -T X_t=(
  typeset x=foo y=bar
  typeset s=${_.x}
  create() {
    _.y=bam
  }
)
X_t x
[[ ${x.x} == foo ]] || err_exit 'x.x should be foo'
[[ ${x.y} == bam ]] || err_exit 'x.y should be bam'
[[ ${x.s} == ${x.x} ]] || err_exit 'x.s should be x.x'
typeset -T Y_t=( X_t r )
Y_t z  ## <------ the SIGSEV happens when this statement is executed
[...]

PR #269, opened 2017-12-26
Make types and friends fail on SIGABRT

catull:
This PR prevents one test to fail as soon as it SIGSEVs.
It addresses a concern in issue #267.

Apologies for the extra lines that have their final white-space removed at the end.

### DIFF
--- a/src/cmd/ksh93/tests/types.sh
+++ b/src/cmd/ksh93/tests/types.sh
@@ -29,7 +29,8 @@ Command=${0##*/}
 integer Errors=0
 
 tmp=$(mktemp -dt tmp.XXXXXXXXXX) || { err_exit mktemp -dt failed; exit 1; }
-trap "cd /; rm -rf $tmp" EXIT
+trap 'rm -fr $tmp' EXIT
+trap 'rm -fr $tmp; exit 1' 0 1 2 3 5 7 8 9 10 15
 
 integer n=2
 


catull:
The line
```ksh
trap 'something' EXIT
```
make every script succeed!
It is a bad pattern.

catull:
It traps on macOS 10.12.6 and succeeds on my Fedora 27 box.

src/cmd/ksh93/tests/types.sh
```diff
@@ -29,7 +29,8 @@ Command=${0##*/}
 integer Errors=0
 
 tmp=$(mktemp -dt tmp.XXXXXXXXXX) || { err_exit mktemp -dt failed; exit 1; }
-trap "cd /; rm -rf $tmp" EXIT
+trap 'rm -fr $tmp' EXIT
+trap 'rm -fr $tmp; exit 1' 0 1 2 3 5 7 8 9 10 15
```
krader1961: This is completely unnecessary. If you do some experiments you should see that the parent shell is told the child shell died due to a signal. Most shells convert that into a status of 128 plus the signal number. We don't need to remove the temp file in this situation since it should never happen. And if it does happen we don't care that the temp file didn't get removed. This just adds unneeded and unwanted complexity.

Also, your commit comment implies this traps SIGABRT but on every system I've ever used that is signal 6 which is not in the list you're trapping.

P.S., You can't trap signal 9 (SIGKILL) and there is no signal zero. Ksh treats `0` as a synonym for `EXIT`. So this simply makes the preceding trap statement irrelevant.

krader1961:
It is interesting that this change causes the `types` script to fail with a SIGSEGV. But even if it didn't do that we don't need or want this specific change. What we do want is to modify the script which runs unit tests like this one to treat any non-zero exit status as a failure.

PR #270, opened 2017-12-27
Remove 'iconv' feature test from libast

siteshwar:
Remove 'std/iconv.h' header and replace all occurrences of <iconv.h>
with <ast_iconv.h>


catull:
Found a better way; watch for the PR coming your way.

src/lib/libast/include/ast_iconv.h
```diff
@@ -0,0 +1,85 @@
...
+#define _sys_types	1	/* #include <sys/types.h> ok */
+#define _hdr_iconv	1	/* #include <iconv.h> ok */
+#define _lib_iconv_open	1	/* iconv_open() in default lib(s) */
+#define _lib_iconv_close	1	/* iconv_close() in default lib(s) */
+#define _lib_iconv	1	/* iconv() in default lib(s) */
+#define _nxt_iconv "../include/iconv.h"	/* include path for the native <iconv.h> */
+#define _nxt_iconv_str "../include/iconv.h"	/* include string for the native <iconv.h> */
+#include <ast_common.h>
+#include <ccode.h>
+#include <../include/iconv.h>	/* the native iconv.h */
```
krader1961: Now that this header has a distinct name that doesn't conflict with the system header this should just be `#include <iconv.h>` and the `_nxt_iconv` stuff above removed.

src/lib/libast/meson.build
```diff
@@ -195,6 +195,10 @@ if cc.has_header('mach-o/dyld.h')
     libast_c_args += [ '-D_hdr_mach_o_dyld']
 endif
 
+if not cc.has_function('iconv_open', prefix : '#include <iconv.h>')
```
krader1961: Move the `cc.find_library()` checks below ahead of this code and add `dependencies: iconv_dep` this test.

krader1961:
You also need to replace the `#include <iconv.h>` with `#include "ast_iconv.h"` at least in the *src/lib* code and possibly in the *src/cmd* code.

krader1961:
And remove *src/lib/libast/std/iconv.h*.

P.S., Everyplace the code has a header that shadows a system header needs to be changed. Obviously not as part of this particular PR. I'm just pointing out how confusing and problematic the current pattern is.

krader1961:
With the changes I recommended this builds and the test results are unchanged on macOS 10.13 and 10.12 as well as FreeBSD 11 and Ubuntu 16.10 and OpenSUSE 42.3.

PR #275, opened 2017-12-28
The logic is the string must match the regexp "cpu[ 0-9]" w/o quotes

catull:
This PR corrects the logic.

### DIFF
--- a/src/lib/libast/aso/asoactivecpu.c
+++ b/src/lib/libast/aso/asoactivecpu.c
@@ -50,7 +50,7 @@ asoactivecpu(void)
 			ssize_t	n;
 			int	fd;
 			char	buf[8*1024];
-		
+
 			if ((fd = open("/proc/stat", O_INTERCEPT|O_RDONLY)) >= 0)
 			{
 				if ((n = read(fd, buf, sizeof(buf) - 1)) > 0)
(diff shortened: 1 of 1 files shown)


src/lib/libast/aso/asoactivecpu.c
```diff
@@ -59,9 +59,9 @@ asoactivecpu(void)
 					buf[n] = 0;
 					do
 					{
-						if (s[0] == 'c' && s[1] == 'p' && s[2] == 'u' && (s[3] == ' ' || s[3] >= '0' && s[3] <= '9'))
+						if (s[0] == 'c' && s[1] == 'p' && s[2] == 'u' && (s[3] == ' ' || (s[3] >= '0' && s[3] <= '9')))
```
krader1961: The parens aren't actually needed other than to clarify the code and quiet a compiler warning. The `&&` has higher precedence than `||` so the expression was already grouped this way. Nonetheless, do feel free to add the parens anywhere you see the compiler warning about this construct.

krader1961:
Merged. Thx.

src/lib/libast/aso/asoactivecpu.c
catull: clang does complain; perhaps gcc does not, cannot remember after so many iterations.

krader1961: > clang does complain; perhaps gcc does not, cannot remember after so many iterations.

Yes, but it is doing so not because the logic is necessarily wrong but to ensure you really did mean for the `a && b` portion to be evaluated as if it was surrounded by parens. This is another case where you need to be careful and consider whether the parens should be around the `x || y` subexpression.

catull: I am aware of C precedence rules and mathematical precedence.
In this case, I had the choice of keeping a warning in or out.
Two extra parentheses.
In future the compiler may yell "unnecessary parentheses", then one of use will remove them.
We win now, we will win then.
Fewer noisy lines to spend time on ...

krader1961: @catull, Take a deep breath and reread what I wrote. I was pointing out that your commit comment "This PR corrects the logic." is wrong. The logic was correct without your change. Having said that I agree with making this change (and similar changes in other modules) precisely to remove any ambiguity and eliminate compiler warnings.

PR #277, opened 2017-12-28
Correct $INSTALLROOT when runing bin/package clean

dannyweldon:
There was a bug that caused running "bin/package clean" twice to
remove the entire repo.

### DIFF
--- a/bin/package
+++ b/bin/package
@@ -4996,6 +4996,11 @@ admin)	while	test ! -f $admin_db
 	;;
 
 clean|clobber)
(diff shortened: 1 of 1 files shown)


krader1961:
Thx. Merged into the `2016-01-10-beta` branch (not master since we're no longer using that build tool in that branch).

dannyweldon:
Thank you.  Yes, I am just trying to tie off a few loose ends with 2016-01-10-beta.  There may be a few more coming and some that you might want to merge into master.

PR #278, opened 2017-12-28
Fix some logical typos in src/lib/libast/comp/iconv.c

catull:
Some typos involving assigment in if clause that evaluates to being non-zero used for boolean value.

Line 1170 actually may have had a hidden bug.
It is slightly different from line 1168.
```c
if (!t || !*t || *t == '-' && !*(t + 1) || !strcasecmp(t, name_local) || !strcasecmp(t, name_native))
if (!f || !*f || *f == '-' && !*(f + 1) || !strcasecmp(t, name_local) || !strcasecmp(f, name_native))
//                                                     ^
```
I believe the character pointed to should actually be an 'f', not a 't'.

### DIFF
--- a/src/lib/libast/comp/iconv.c
+++ b/src/lib/libast/comp/iconv.c
@@ -523,20 +523,21 @@ if (error_info.trace < DEBUG_TRACE) sfprintf(sfstderr, "%s: debug%d: AHA#%d _ast
 		else if (!(cp = (const _ast_iconv_list_t*)ccmaplist((_ast_iconv_list_t*)cp)))
 			cp = codes;
 	}
-	if (bp != 0)
+	cp = bp;
+	if (cp != 0)
 	{
-		cc = bp->ccode;
-		if (bp->canon)
+		cc = cp->ccode;
+		if (cp->canon)
 		{
-			if (bp->index)
+			if (cp->index)
 			{
 				for (m += sub[1]; *m && !isalnum(*m); m++);
 				if (!isdigit(*m))
-					m = bp->index;
+					m = cp->index;
 			}
 			else
 				m = "1";
-			b += sfsprintf(b, e - b, bp->canon, m);
+			b += sfsprintf(b, e - b, cp->canon, m);
 			if (cc == CC_UTF && *m != '8')
 				cc = CC_ICONV;
 		}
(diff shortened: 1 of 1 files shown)


src/lib/libast/comp/iconv.c
```diff
@@ -523,7 +523,7 @@ if (error_info.trace < DEBUG_TRACE) sfprintf(sfstderr, "%s: debug%d: AHA#%d _ast
 		else if (!(cp = (const _ast_iconv_list_t*)ccmaplist((_ast_iconv_list_t*)cp)))
 			cp = codes;
 	}
-	if (cp = bp)
+	if (cp == bp)
```
siteshwar: This is not a typo. It's a pattern used in ksh code for assignment.

catull: Pattern is one thing.
I need more than that.

Both `cp` and `bp` get assigned up to that line (line 526).

`bp` is `0` at the beginning, line 489.
`cp` starts out as a list, so it points to something, line 491.

`bp` may be assigned to the current value of `cp` in lines 509 and 514; it may still be `0`.
`cp` may be assigned a new value in line 524.

So, if `bp` remained `0` - never assigned anything, it would overwrite `cp` in line 526.
`bp` is not used past line 526.

In fact, this is equivalent code
```c
if (bp != 0) {
 // use bp instead of cp
 //....
}
```
Why assign it to `cp` if you can go ahead with `bp`.

krader1961: @siteshwar is correct although I think of it as an anti-pattern. As I restyle each module If add additional parens to silence the compiler warning and make it clear that it should be an assignment rather than equality test. Or, in a case like this, I'll generally hoist the assignment out of the `if`; e.g.,:

```
cp = bp;
if (cp) {
```

src/lib/libast/comp/iconv.c
```diff
@@ -523,20 +523,20 @@ if (error_info.trace < DEBUG_TRACE) sfprintf(sfstderr, "%s: debug%d: AHA#%d _ast
 		else if (!(cp = (const _ast_iconv_list_t*)ccmaplist((_ast_iconv_list_t*)cp)))
 			cp = codes;
 	}
-	if (cp == bp)
+	if (bp != 0)
```
krader1961: No. Do not make this transformation in a change whose sole purpose is eliminating a compiler warning about doing an assignment in a `if()` statement. Just hoist the assignment:

```
cp = bp;
if (cp)
```

Leave everything else alone. I don't care if you are confident this transformation is correct. The code is such a mess, and the unit tests so inadequate, that your elimination of the assignment to `cp` is too risky.

catull: Okay.

krader1961:
Merged with a couple of tweaks. You never need to write `if (x != 0)`; idiomatic C is just `if (x)`.

PR #280, opened 2017-12-28
Add support for additionally installed GNU libiconv.

catull:
It defines an extra compiler constant, LIBICONV_PLUG, in order to use the GNU libiconv function iconv_open and iconv_close.
As far as I can tell, there is no impact for Linux, or for people who only have libiconv which comes with the system.

### DIFF
--- a/src/lib/libast/include/ast_iconv.h
+++ b/src/lib/libast/include/ast_iconv.h
@@ -9,6 +9,7 @@
 #define _lib_iconv	1	/* iconv() in default lib(s) */
 #include <ast_common.h>
 #include <ccode.h>
+#define LIBICONV_PLUG 1 /* prefer more recent GNU libiconv */
 #include <iconv.h>	/* the native iconv.h */
 
 #define ICONV_VERSION		20121001L


catull:
It be great if this PR were addressed more timely.
@krader1961 @siteshwar 

OpenSUSE build failed once, on `signal`.
Fedora build was okay.

krader1961:
> It be great if this PR were addressed more timely.

@catull, Everyone working on this is doing so in their free time. None of us are getting paid for this work. It might take a few days before a PR is reviewed and rejected or merged. While we appreciate your enthusiasm and assistance in cleaning up the code the tone of your comment is not appreciated. It's been three fucking hours since you created this PR. You should wait at least a couple of days before making a polite "ping?" comment to remind someone to take a look at it.

krader1961:
Merged.

krader1961:
Sorry about that. Addressing IWYU warnings caused the system *iconv.h* header to be included before the *ast_iconv.h* header. This is a good example why doing things like putting directives like `#define LIBICONV_PLUG 1` in a header that then includes a different header is fragile. The code is littered with such land mines. Blindly defining it via the `CFLAGS` env var has it's own problems as doing so breaks other platforms such as Cygwin. I'll try to merge a less fragile fix today.

P.S., You should consider removing most, if not all, of those `-Wno-` flags. We've worked hard to remove things like the IFFE/Nmake specific `#pragma prototyped` directive. You shouldn't be seeing any compiler warnings that would be suppressed by those flags.

catull:
I actually am working on a small cleanup relative to the LIBICONV_PLUG.

Have a patch in preparation.

Right, will remove all `-Wno-` flags while I am preparing the patch.

catull:
Your efforts are fantastic.

Without any of the `-Wno-` flags, the build shows no warnings,`ninja` keeps counting up.

Well done!

krader1961:
Thanks. One of the things that drives me a little crazy are compiler warnings. When I make a change I should be able to just assume any warnings are the result of my change. I shouldn't have to diff the output to see if I've introduced new warnings. So earlier this year I spent a few days just fixing all the existing warnings. I'd love to see people run `bin/lint` and tackle the warnings from cppcheck and oclint. At least the ones that don't require refactoring the code to reduce function size/complexity.

PR #283, opened 2017-12-29
Remove unused functions.

catull:
This PR removes unused functions.

These are

```log
../src/cmd/ksh93/bltins/sleep.c:148:13: warning: unused function 'completed' [-Wunused-function]
../src/cmd/ksh93/bltins/typeset.c:78:18: warning: unused function 'load_class' [-Wunused-function]
../src/cmd/ksh93/edit/pcomplete.c:255:13: warning: unused function 'evaluate' [-Wunused-function]
../src/cmd/ksh93/edit/vi.c:2554:16: warning: unused function '_ismetach' [-Wunused-function]
../src/cmd/ksh93/sh/io.c:167:12: warning: unused function 'matchf' [-Wunused-function]
../src/cmd/ksh93/sh/name.c:2730:14: warning: unused function 'tableval' [-Wunused-function]
../src/lib/libast/disc/sfdcdio.c:105:16: warning: unused function 'dioread' [-Wunused-function]
../src/lib/libast/disc/sfdcdio.c:118:16: warning: unused function 'diowrite' [-Wunused-function]
../src/lib/libast/disc/sfdcdio.c:131:12: warning: unused function 'dioexcept' [-Wunused-function]
../src/lib/libast/misc/spawnvex.c:288:1: warning: unused function 'apply_flags' [-Wunused-function]
../src/lib/libast/port/astconf.c:1386:1: warning: unused function 'nativeconf' [-Wunused-function]
```

Due to the removal of the functions above in ../src/lib/libast/disc/sfdcdio.c, a few more unused functions could be removed.


krader1961:
I haven't looked at this change yet but just an FYI that when making this type of cleanup you need to consider whether any of the code is conditionally compiled. We recently found a few cases where our current build time defines were incomplete and there may be more. So you can't just remove the code because the compiler says it isn't currently used. You have to check whether it might be used if a `#define` were changed.

catull:
I actually checked and search the code base for each one of the functions, if it were "used" in any way.
Nothing turned up.

krader1961:
You missed `nativeconf()` which is used in the very file containing the definition you removed. But the use is predicated on `#ifdef _pth_getconf_a` which is presumably false on your system but not on OpenSUSE and other Linux systems.

catull:
Thanks, will fix it next.

krader1961:
Merged with a fix for the `nativeconf()` issue I noted above.

Issue #81, opened 2017-10-25
Remove support for "att" and "ucb" universes (i.e., OS personalities)


krader1961:
One way to make it easier to switch to a new build system (e.g., Cmake or Meson) is to eliminate support for long dead operating systems. Such as DYNIX/3 from Sequent Computer Systems. As evidenced by these statements emitted by `bin/package make`:

```
iffe: test: is universe a command ... no
iffe: test: is getuniverse a library function ... no
iffe: test: is sys/universe.h a header ... no
```

See my [comment](https://github.com/att/ast/issues/42#issuecomment-339210893) on issue #42.

siteshwar:
I would be mostly interested in supporting just Linux and BSD family of operating systems. I am fine with dropping support for rest.

krader1961:
I think we should try to support as many operating systems as possible that are still being actively maintained. Not just used but being maintained by which I mean that bugs are being fixed in a timely manner. There may be people still using Solaris for example but it is dead. From its wikipedia page:

> On September 2, 2017, Simon Phipps reported that Oracle had laid off the Solaris core development staff, interpreting it as sign that Oracle no longer intends to support future development of the platform.

Clearly we shouldn't expend resources trying to make ksh93 work on dead OS's. Removing support for an OS that is supported by ksh93u is a more contentious question and probably needs to be handled on a case by case basis. I'm comfortable removing support for DYNIX/3 because it's dual universe support adds non-trivial complexity to ksh and no one is going to care that we no longer support it. The same is true for UWIN. Not to mention pre-ANSI C compilers.

krader1961:
FWIW, It turns out that the att/ucb universe mechanism was found in OS's other than DYNIX/3: https://en.wikipedia.org/wiki/Universe_(Unix). But as all of those operating systems are now dead we can still safely remove support for dual universe OS's. Also, to the extent there are any SVR4 based distros being used that still have a /usr/ucb directory the commands therein should not be special-cased by ksh93. The native SVR4 variants are perfectly acceptable.

krader1961:
Removing support for att versus ucb "universe"s is going to be really
difficult without breaking things. At least inadvertantly. Consider this
block of code from function `B_echo()` in *src/cmd/ksh93/bltins/print.c*:

```
       static char bsd_univ;

       /* This mess is because /bin/echo on BSD is different */
       if(!prdata.sh->universe)
       {
               register char *universe;
               if(universe=astconf("UNIVERSE",0,0))
[...]

PR #286, opened 2017-12-29
Removed code supporting UWIN.

catull:
Support for _WIN32, __CYGWIN__ is intact.
_WINIX now is a synonym for __CYGWIN__; up-to now, it included UWIN and __EMX__ (OS/2 emulation).
Also removed 2 references to __INTERIX; had to look that one up: https://en.wikipedia.org/wiki/Interix.
Did not consider it worth keeping it.

Builds on Fedora, runs all tests OK.
Builds on mac OS 10.12.6 (with a small hack for libiconv), runs with the usual 8 test failures.
Don't have access to a CYGWIN box; the changes should not impact CYGWIN.

This PR addresses a concerned discussed in issue #81.


src/cmd/INIT/ratz.c
```diff
@@ -93,7 +93,7 @@ static const char usage[] =
 
 #else
 
-#if !defined(_WINIX) && (_UWIN || __CYGWIN__ || __EMX__)
```
krader1961: FYI, This is part of the legacy build system that we no longer support. It simply hasn't been removed as switching from Nmake to Meson is a work in progress.

src/cmd/ksh93/bltins/misc.c
```diff
@@ -506,18 +506,15 @@ int	b_universe(int argc, char *argv[],Shbltin_t *context)
 #endif /* cmd_universe */
 
 #if SHOPT_FS_3D
-#if _UWIN
-#include <sys/mount.h>
-#endif
-#   if 0
+#if 0
```
krader1961: When making these types of cleanups try to avoid superfluous cleanups such as removing the whitespace here.

src/lib/libast/vmalloc/vmbest.c
```diff
@@ -19,7 +19,7 @@
 *                     Phong Vo <phongvo@gmail.com>                     *
 *                                                                      *
 ***********************************************************************/
-#if defined(_UWIN) && defined(_BLD_ast)
+#if !defined(_BLD_ast)
```
krader1961: This is wrong. The enclosed code should not be compiled if `_UWIN` is not defined. It should therefore be removed rather than predicated on `_BLD_ast` not being defined. Notice the `a && b` form of the statement you are replacing.

src/lib/libast/vmalloc/vmclear.c
```diff
@@ -19,7 +19,7 @@
 *                     Phong Vo <phongvo@gmail.com>                     *
 *                                                                      *
 ***********************************************************************/
-#if defined(_UWIN) && defined(_BLD_ast)
+#if !defined(_BLD_ast)
```
krader1961: As I noted in my previous comment this is incorrect.

src/lib/libast/vmalloc/vmbest.c
catull: Have you built it that way ?
It doesn't build the way you describe.
This pattern will manifest a dozen times in the PR.

src/cmd/ksh93/bltins/misc.c
catull: I first though it wad a weird case of
```c
          if 0
```
w/o parentheses, then I spotted the '#' at the beginning of the line.
Code is read more often that it is written; so this optical illusion causes one to pause and think about it.
But, no biggie.
Won't change it in future.

src/lib/libast/vmalloc/vmbest.c
krader1961: > Have you built it that way ?

No, and it doesn't matter. Think about the condition you are modifying. If the condition is `a && b` and `a` can never be true then it does not matter whether or not `b` is true. The code conditioned on this statement will never be compiled. So it is incorrect to replace `a && b` with `!b`.

krader1961:
I'll take a closer look at this change tomorrow. In the future please limit the scope of a change made by a pull-request. In this PR you not only made changes related to the `UWIN` environment but also unrelated environments (e.g., `__INTERIX`). That makes reviewing, and approving, the change more difficult. Those additional changes are definitely a good idea and are likely to be merged on their merits alone. But a PR that is focused on more than a single improvement (like this one) is harder to review and thus approve for merging.

catull:
Appreciate your feedback.
The title is misleading, I admit.

src/cmd/ksh93/bltins/misc.c
krader1961: It's not that it shouldn't be changed. It's that you shouldn't mix functional changes with style changes. Mixing the two makes it considerably harder to review the change and determine if the functional changes are correct.

src/lib/libast/misc/getenv.c
```diff
@@ -21,13 +21,13 @@
 ***********************************************************************/
 #pragma prototyped
 
-#if _UWIN && __STDPP__
+#if __STDPP__
```
krader1961: You introduced a bug. The original was of the form `a && b`. Removing `a` is equivalent to setting it to false and thus the entire expression is false. I'll take care of fixing this when I merge it.

catull: And also similarly on line 30.

src/lib/libast/misc/getenv.c
```diff
@@ -21,13 +21,13 @@
...
 __STDPP__directive pragma pp:hide getenv
 #endif
 
 #include "intercepts.h"
 
-#if _UWIN && __STDPP__
+#if __STDPP__
```
catull: This is buggy too.

src/lib/libast/misc/getenv.c
```diff
@@ -45,7 +45,7 @@ Intercepts_t	intercepts
 		= { 0 };
 #endif
 
-#if _UWIN && !defined(getenv)
+#if !defined(getenv)
```
catull: So is this one ....

src/lib/libast/misc/getenv.c
```diff
@@ -93,7 +93,7 @@ default_getenv(const char* name)
 extern char*
 getenv(const char* name)
 {
-#if _UWIN && !defined(getenv) /* for ast54 compatibility */
+#if !defined(getenv) /* for ast54 compatibility */
```
catull: ... and this one.

krader1961:
Merged with the changes I said I would make and not the subsequent changes to this PR (which I didn't even look at).

PR #287, opened 2017-12-29
Removed old constants.

catull:
This PR removes constants for older OS and compilers.
These are:
- _UWIN
- __EMX__
- MSDOS
- OS2
- WINDOWS (16 bit)
- SYS16BIT
- __BEOS__
- AMIGA
- ATARI
- VACX / VMS
- MACOS (version 8, 9)
- __TURBOC__
- __BORLANDC__
- TOPS20
- RISCOS
- Prime/PRIMOS

Unsure about __MVS__ and __HO_AIX__, left them in.

### DIFF
--- a/src/cmd/INIT/ratz.c
+++ b/src/cmd/INIT/ratz.c
@@ -93,7 +93,7 @@ static const char usage[] =
 
 #else
 
-#if !defined(_WINIX) && (_UWIN || __CYGWIN__ || __EMX__)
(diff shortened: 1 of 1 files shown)


krader1961:
You're doing too much in a single PR. Something like this should be a series of very small commits each of which eliminates a single, or possibly two or three closely related (e.g., TURBOC, BORLANDC), symbols. And removing `MACOS` is definitely a problem. Please refactor.

PR #289, opened 2017-12-29
Remove conditional compilation with #if 0.

catull:
In all 3 cases, system functions are declared: exit(), malloc(), calloc() and free().


krader1961:
Please note that there is code in the AST code (both library and ksh) predicated on `#if 0` directives that we do not want to unilaterally remove. At least not solely because the code is predicated on a `#if 0` directive. That code may be present to help debug problems. So each instance needs to be evaluated on its own merits.

I'm not merging this for the same reason given for similar PRs: This only affects building the legacy Nmake build system.

PR #290, opened 2017-12-29
Remove support for OS/400 and MVS.

catull:
This  PR removes support for OS/400 and MVS.

### DIFF
--- a/src/cmd/INIT/ratz.c
+++ b/src/cmd/INIT/ratz.c
@@ -433,31 +433,6 @@ typedef uLong FAR uLongf;
 #  define z_off_t long
 #endif
 
-#if defined(__OS400__)
-#  define NO_vsnprintf
-#endif
-
-#if defined(__MVS__)
-#  define NO_vsnprintf
-#endif
-
-/* MVS linker does not support external names larger than 8 bytes */
-#if defined(__MVS__)
-#   pragma map(deflateInit_,"DEIN")
-#   pragma map(deflateInit2_,"DEIN2")
-#   pragma map(deflateEnd,"DEEND")
-#   pragma map(deflateBound,"DEBND")
-#   pragma map(inflateInit_,"ININ")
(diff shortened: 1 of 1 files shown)


KeithBierman:
Without taking a position on "blind" , etc.


   - Remove support for OS/400 and MVS. They are not a first tier platforms


I think it would be clearer if this sort of thing were stated as "These are
not supported platforms" ... tiers would suggest that there would be some
other way for them to have their code supported (some PR to some other
repo?), which would seem quite different than what these PR's seem to be
accomplishing.


Keith Bierman
khbkhb@gmail.com
303 997 2749

catull:
In FreeBSD, they distinguish 3 different tiers.

Tier 1 is supported architecture: x64, arm7, PPC and PPC64 etc
Tier 2 is development architecture.
Tier 3 is experimental architecture.

I think NetBSD and OpenBSD also use "tier N" naming.
Here tier is the CPU hardware.

Can't be directly applied to ksh93, as it is an application on those OS.

Your suggestion makes sense: "supported" / "not supported platform", will use it from now on.

krytarowski:
NetBSD uses term tier with similar meaning 1-3.

krader1961:
Closing without merging because this only affects the legacy Nmake build system which is code we're going to remove in its entirety in the near future. It does not affect building ksh using Meson. Also, we need to be careful about removing support for IBM "mainframes" that use EBCDIC which we still want to support.

krader1961:
Also, phrases like "first tier platform" should be avoided. At least at this point in time. There are lots of legacy platforms we no longer want to support. For example, platforms that only have pre-ANSI C compilers. Or dead operating systems like Sequent Computer Systems DYNIX/3 (which I worked for and supported many years ago). Similar, clear cut, legacy support will be removed. But even if this PR affected more than the Nmake build tool it is not clear it clears that bar.

PR #292, opened 2017-12-29
Remove VMS support.

catull:
This PR removes support for VMS and VAX.
 VMS is not a first tier platform.


krader1961:
This only affects building the legacy Nmake build system we are no longer using and the modified file will be removed in the near future.

PR #293, opened 2017-12-29
Remove BeOS support.

catull:
This PR removes support for BeOS.
It is not a first tier platform.

### DIFF
--- a/src/cmd/INIT/ratz.c
+++ b/src/cmd/INIT/ratz.c
@@ -363,18 +363,6 @@ static const char usage[] =
 #  endif
 #endif
 
-#if defined (__BEOS__)
-#  ifdef ZLIB_DLL
-#    ifdef ZLIB_INTERNAL
-#      define ZEXPORT   __declspec(dllexport)
-#      define ZEXPORTVA __declspec(dllexport)
-#    else
-#      define ZEXPORT   __declspec(dllimport)
-#      define ZEXPORTVA __declspec(dllimport)
-#    endif
-#  endif
-#endif
-
 #ifndef ZEXTERN
 #  define ZEXTERN extern
 #endif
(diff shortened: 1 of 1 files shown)


siteshwar:
Please stop blindly removing code. The source code under INIT directory is not used at all and will be removed altogether in future.

catull:
The Wiki gives no indication on what the focus is.

I have a general idea, and this idea forms daily.

I was instructed to bite-size a larger PR on my last review.
I was not told to leave it altogether; you did not give input then.

So, you can consider it "blind action", I do not.

siteshwar:
You should check if the files you are touching are actually being compiled or not. Also, there are some features which are currently under discussion (for e.g. https://github.com/att/ast/issues/129), it's possible that such features will require the code you remove.

If you want to focus on something, I will suggest you to look at the test cases. Current test cases test only subset of ksh functionality and needs to be improved for more thorough testing.

catull:
I am fine with the PRs not being merged if they are in the way.

My motivation is to build ksh93 on macOS natively, and being able to use it reliably.
The other tools (nmake etc) are not my primary concern.

So back to test cases it is for now.

krader1961:
> I was instructed to bite-size a larger PR on my last review.

I provided that feedback based solely on your earlier PR whose commit comment listed a long laundry list of things that were affected. I did not look at the change. The fact your original PR, and this spinoff, modifies files we intend to remove in the immediate future is why I asked that your earlier PR be split into smaller pieces. Thus making it easier to notice if a particular change is worthwhile and correct.

> The Wiki gives no indication on what the focus is.

The Github wiki is not the authoritative answer to that question. The [README](https://github.com/att/ast/blob/master/README.md) you see when browsing this project makes it reasonably clear the focus is on `ksh` (or `ksh93` if you prefer) and building it with Meson rather than the legacy Nmake build system.

extrowerk:
Will try to add support for Haiku.

PR #294, opened 2017-12-29
Remove support for various DOS, Win3.1 and similar legacy PC OS

catull:
These platforms include
- UWIN
- MSDOS
- Windows 3.x
- OS/2 (along with __EMX__)
- 16 bit small model
- Borland C, Turbo C compilers for the above platforms

They are not first tier platforms.

### DIFF
--- a/src/cmd/INIT/ratz.c
+++ b/src/cmd/INIT/ratz.c
@@ -93,7 +93,7 @@ static const char usage[] =
 
 #else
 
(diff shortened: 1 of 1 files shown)


krader1961:
Thanks but this isn't needed. It only affects the legacy Nmake build system which we are no longer using. We simply haven't gotten around to removing those files.

I recommend you pick a specific [issue](https://github.com/att/ast/issues) and fix that rather than doing cleanup work like this that isn't going to be merged.

PR #296, opened 2017-12-29
Remove 'vmalloc' feature test for libast

siteshwar:
Moved some iffe tests to Meson


src/lib/libast/vmalloc/meson.build
```diff
@@ -7,3 +7,96 @@ libast_files += [
     'vmalloc/vmsegment.c', 'vmalloc/vmset.c', 'vmalloc/vmstat.c',
     'vmalloc/vmstrdup.c', 'vmalloc/vmtrace.c', 'vmalloc/vmuserdata.c'
 ]
+
+
+if cc.has_function('mallinfo', prefix: '#include <malloc.h>')
+    libast_c_args += [ '-D_lib_mallinfo=1' ]
```
krader1961: There isn't any need for the `=1` portion since without it that is the default value assigned to the symbol. I point this out because elsewhere (e.g., `-D_lib_utimensat`) we don't explicitly include the `=1`. We should either consistently set the value or omit it to use the equivalent default.

src/lib/libast/vmalloc/malloc.c
```diff
@@ -600,7 +600,11 @@ void _vmoptions(int boot)
 
 #if _malloc_hook
 
+#if __FreeBSD__
```
krader1961: Note that macOS also puts some things needed here in *stdlib.h*. It also has a *malloc.h* that defines things like `struct mstats` but it's at *malloc/malloc.h*. I don't think we want to do it this way. There should be a meson test for these two headers. Although I think it is safe to just assume *stdlib.h* is available and unconditionally include it. We probably only need a feature test for the *malloc.h* header.

krader1961:
FYI, The `iffe` generated feature header does `#include <ast_mmap.h>` which is no longer automatically included.  I don't yet know if that is the source of this PR's failure to build on macOS. Interestingly there are two files that directly import it: *src/lib/libast/sfio/sfhdr.h* and *src/lib/libast/features/vmalloc*. What a fracking pile of caca.

src/lib/libast/vmalloc/malloc.c
siteshwar: Test for headers will not work because FreeBSD has this definition for `malloc.h` :

```
#error "<malloc.h> has been replaced by <stdlib.h>"
```

That's why I have put a separate check for FreeBSD there.

siteshwar:
@krader1961 Thanks for the hint. I managed to fix compilation on macOS. I have updated the pull request according to your comments.

krader1961:
LGTM. FWIW, after my previous comments I did a build without this change and saved a copy of the generated *FEATURES/vmalloc* file. Then I just methodically went through adding the macro symbols it defined as true until the build worked.

PR #298, opened 2017-12-31
Run 'sfinit.c' and 'signal.c' tests without iffe

siteshwar:
These tests do not have to be run through iffe and should be refactored
out later.

Also cleaned up some unneeded code.


krader1961:
LGTM

Issue #301, opened 2018-01-01
Test `locale` assumes GNU `join` flags are supported which isn't valid for BSD `join`


catull:
In comparing the difference in runtime between macOS and Fedora, I found this behaviour.

At first, on macOS I get this with `meson test -v locale`:

```log
TODO: Enable when custom builtins of external commands is working.
join: multi-character tab ‘＠’
	locale.sh[205]: join test script failed -- exit code 1
	locale.sh[208]: LC_ALL test script failed -- expected 'f1', got ''
1/1 locale                                  FAIL     0.64 s

OK:         0
FAIL:       1
SKIP:       0
TIMEOUT:    0
```

The same test run on Fedora actually gets the same error, you only see it if you look very hard for it.
The effect is though that it fails on macOS repeatedly, but not on Fedora.

These lines slightly changed make macOS catch up:

```diff
diff --git a/src/cmd/ksh93/tests/locale.sh b/src/cmd/ksh93/tests/locale.sh
index a353949..5be316f 100755
--- a/src/cmd/ksh93/tests/locale.sh
+++ b/src/cmd/ksh93/tests/locale.sh
@@ -196,9 +196,9 @@ done
 # setocale(LC_ALL,"") after setlocale() initialization
 locale=en_US.UTF-8

-printf 'f1\357\274\240f2\n' > input1
-printf 't2\357\274\240f1\n' > input2
-printf '\357\274\240\n' > delim
+printf 'f1@f2\n' > input1
+printf 't2@f1\n' > input2
+printf '@' > delim
 print "export LC_ALL=$locale
 join -j1 1 -j2 2 -o 1.1 -t \$(cat delim) input1 input2 > out" > script
 $SHELL -c 'unset LANG ${!LC_*}; $SHELL ./script' ||
```

Credit goes to https://stackoverflow.com/questions/24587955/bash-sort-and-multi-character-tab-error

The temp file `delim` above could be entirely removed in favour of

```sh
 join -j1 1 -j2 2 -o 1.1 -t '@' input1 input2 > out" > script
```

krader1961:
[...]

PR #302, opened 2018-01-01
Fix for unit test locale.sh and siblings

catull:
This PR fixes a "bug" with `join -t`.
Join is a utility that in this particular unit test is used to combine two files satisfying certain criteria.
It turns out that `join -t` accepts either a single character or a string of with blank or comma as separating characters.
The test instead uses a three-character string containing no space or comma.
It aims to use the at-sign (@) as tab character.

The PR changes the input files to use the intended tab character.
It fixes issue #301.


krader1961:
Not merging because a) it doesn't fix the problem and b) changes the semantics such that it no longer verifies that a non-ASCII field delimiter works correctly.

PR #303, opened 2018-01-01
Fix proper indentation

catull:
`gcc` complains in three cases that the indentation is misleading, and therefore the "guarding" is not evident.

The warnings look this:

```log
../src/lib/libast/path/pathcanon.c:605:3: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
   for (a = s + 2; *a == '/'; a++);
   ^~~
../src/lib/libast/path/pathcanon.c:617:4: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘for’
    *t++ = *s++;
    ^
../src/lib/libast/path/pathcanon.c:683:107: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
     if ((flags & PATH_PHYSICAL) && (t - 1) > (x ? x : canon) && (loop < 32 || (flags & PATH_EXISTS) || *s && (flags & PATH_EXCEPT_LAST)))
                                                                                                        ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The PR corrects the indentation levels appropriately.

If this PR collides with the ongoing styling campaign, feel free to reject it.

Still, only 3 C source files have improper indentation from the view-point of a modern C compiler speaks for the discipline of the original engineers at work.

### DIFF
--- a/src/cmd/ksh93/sh/nvdisc.c
+++ b/src/cmd/ksh93/sh/nvdisc.c
@@ -255,7 +255,7 @@ static void	assign(Namval_t *np,const char* val,int flags,Namfun_t *handle)
 	Namval_t	node;
 	union Value	*up = np->nvalue.up;
 	Namval_t	*tp, *nr;
-	if(val && (tp=nv_type(np)) && (nr=nv_open(val,shp->var_tree,NV_VARNAME|NV_ARRAY|NV_NOADD|NV_NOFAIL)) && tp==nv_type(nr)) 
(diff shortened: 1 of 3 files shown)


krader1961:
Thanks, but you're doing this the hard way and the cleanup is incomplete even for the three files you touched. The right way to do this is to run the module through `clang-format` and add the missing braces by hand. As I mentioned in another PR we're also not touching the libast code at this time except to switch building with Meson (i.e., removing the `iffe` uses) and to fix bugs. We're deferring restyling the libast code and silencing compiler warnings with it until we think we're pretty close to having it stripped down to the absolute minimum needed to build ksh.

PR #304, opened 2018-01-01
Fix "dangling else" cases.

catull:
There are 4 cases of "dangling else" that `gcc` points out.
Adding surrounding curly braces makes the code clearer for humans, and avoids the complier warning, too.

This PR fixes "dangling else" cases.


krader1961:
Merged but please note you still left a lot of places, even in the few lines you touched, which need braces. Even though the compiler won't warn about this any place you have this pattern

```
if (x)
    do_x();
else
    do_y();
```

The `if` and `else` blocks need to be enclosed in braces. See the *src/cmd/ksh93/sh/[a-m]&ast;* modules for the new style I'm slowly imposing on the code.

catull:
You are right about that there are far more cases which need a pair of opening and closing curly braces.
I kept the change to the smallest amount of lines.

PR #306, opened 2018-01-01
Make ksh build on Cygwin

### LINKED ISSUE #300: Ksh doesn't build on Cygwin
I finally got around to trying to build `ksh` on Cygwin. Not surprisingly it failed. But the good news is it looks like most of the problems are trivial to fix. For example, we need to include the library (`libcats`) that contains the message catalog functions when building libast. Other failures are a little more perplexing but probably simple to fix. For example, the linker complains that `cygwin_conv_to_posix_path` is undefined. Probably some conditionally compiled code that isn't compiled due to a missing `#define`.

krader1961:
The Cygwin UNIX environment on MS Windows systems has some quirks. This
changes adapts the project to those quirks.

Note that there are a huge number of test failures. But that's true for
macOS too. Also, `meson test` itself sometimes fails (showing a traceback)
for no obvious reason.

Fixes #300

### DIFF
--- a/src/lib/libcmd/iconv.c
+++ b/src/lib/libcmd/iconv.c
@@ -68,7 +68,7 @@ USAGE_LICENSE
 ;
 
 #include <cmd.h>
-#include <iconv.h>
+#include <ast_iconv.h>
 
 /*
  * optget() info discipline function
--- a/src/lib/libast/include/ast_iconv.h
+++ b/src/lib/libast/include/ast_iconv.h
@@ -9,8 +9,14 @@
 #define _lib_iconv	1	/* iconv() in default lib(s) */
 #include <ast_common.h>
 #include <ccode.h>
-#define LIBICONV_PLUG 1 /* prefer more recent GNU libiconv */
-#include <iconv.h>	/* the native iconv.h */
+
+#if !__CYGWIN__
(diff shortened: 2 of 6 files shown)


src/lib/libcmd/iconv.c
```diff
@@ -68,7 +68,7 @@ USAGE_LICENSE
 ;
 
 #include <cmd.h>
-#include <iconv.h>
+#include <ast_iconv.h>
```
siteshwar: We are not building this file yet.

krader1961: Huh? You removed it from `iffe` generation and made it a static header back in commit 6071c35c6d. And it's already being used in several other modules such as *src/lib/libast/tm/tmlocale.c*.

siteshwar: I was referring to this file `src/lib/libcmd/iconv.c`

krader1961: Oh, right, I noticed that as well. I'm simply changing this include for consistency so we don't have a "WTF?" moment if and when we do build this. And if we ultimately drop this code I don't feel bad about throwing away a mechanical one line change :smile:

krader1961:
Merged as commit d4726db07.

PR #307, opened 2018-01-02
Fix a few typos in sh.1

dannyweldon:


dannyweldon:
There is another typo but I'm not sure what it should be.  If you run:

    nroff -man sh.1 | less -R

and search for "so that wait is", you will see that it is missing some text before the closing bracket.  Can anyone comment on what that text should be?  I am guessing "interruptible", but I'm not sure.

Also, the `-D` option is duplicated with different text.  My thoughts were to replace the first `-D` description with the second.  And the man page is missing many options which are visible with `ksh --man`, which has also duplicated the `-r` option.

Would you like me to try to fix all these up in some more PRs?

krader1961:
Thanks -- applied to the beta branch. Since we renamed that file in the master branch (it's now *ksh.1*) I had to make a separate commit for it. And, yes, we'd love to see more PRs for the documentation as it never gets enough love from developers.

> I am guessing "interruptible"

I agree.

> My thoughts were to replace the first -D description with the second.

I agree.

Issue #309, opened 2018-01-03
Recently introduced macro redefinition SF_APPEND


catull:
A few days I ago, this warning appeared in compilations:
```
../src/lib/libast/include/sfio.h:194:9: warning: 'SF_APPEND' macro redefined [-Wmacro-redefined]
#define SF_APPEND       SF_APPENDWR     /* BSDI sys/stat.h              */
        ^
/usr/include/sys/stat.h:336:9: note: previous definition is here
#define SF_APPEND       0x00040000      /* writes to file may only append */
        ^
```

It appears 98 times.

krader1961:
This was introduced by commit 0c1fad6, "Remove 'common' feature test from libast". The problem is that the removed *src/lib/libast/features/common* `iffe` test script had some special tests for things like macOS defining `SF_APPEND`. This type of problem is actually expected to occur since the first step in removing a dependency on `iffe` results in a static header file. As we build on various systems we'll then reintroduce additional feature tests.

In this case the `iffe` dynamically generated header includes `#define _mac_SF_APPEND 1` while the new static header does not. What is interesting is that symbol isn't even needed. It is sufficient to just do a `#ifdef SF_APPEND` where `_mac_SF_APPEND` is tested.

catull:
Will change that accordingly.

What about the next 3 lines
```c
#if !_mac_SF_CLOSE
#define SF_CLOSE	SF_CLOSING	/* AIX sys/socket.h		*/
#endif
```

Do they need to become
```c
#ifdef SF_CLOSE
#define SF_CLOSE	SF_CLOSING	/* AIX sys/socket.h		*/
#endif
```
?

SF_CLOSE is not needed in any .c source files.
It is assigned conditionally in src/cmd/ksh93/include.io.h, which does include the header file sfio.h:
```c
#ifndef SF_CLOSING
#define SF_CLOSING	SF_CLOSE
#endif
```

catull:
In the same header - io.h - the next lines are:
```c
#ifndef SF_APPENDWR
#define SF_APPENDWR	SF_APPEND
#endif
```

This is what I have found out.

In `ast-open.README` there is this change log entry:
```log
01-02-22 pathprobe: reprobe test now checks probe.ini too
	 sfio_s.h: advertize public Sfio_t members with _ prefix
	 sfio.h,features/stdio: add <sfio_s.h> reference
	 sfhdr.h: map <sfio_s.h> _foo to foo
[...]

PR #310, opened 2018-01-03
Fix for macro re-definition of SF_APPEND

catull:
This PR fixes issue #309.

It un-defines SF_APPEND before redefining it.

On macOS SF_APPEND is defined in <sys/stat.h>, as seen in issue #309.
There its value is 0x00040000.

In sfio.h it is SF_APPENDWR - or 0000010 once resolved.
And this value must be used in the end.

### DIFF
--- a/src/lib/libast/include/sfio.h
+++ b/src/lib/libast/include/sfio.h
@@ -191,6 +191,7 @@ struct _sffmt_s
 
 /* namespace incursion workarounds -- migrate to the new names */
 #if !_mac_SF_APPEND
+#undef SF_APPEND_
 #define SF_APPEND	SF_APPENDWR	/* BSDI sys/stat.h		*/
 #endif
 #if !_mac_SF_CLOSE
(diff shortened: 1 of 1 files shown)


src/lib/libast/include/sfio.h
```diff
@@ -191,6 +191,7 @@ struct _sffmt_s
 
 /* namespace incursion workarounds -- migrate to the new names */
 #if !_mac_SF_APPEND
+#undef SF_APPEND_
```
krader1961: This is the wrong fix. See the comment I added to the associated issue. The correct fix, that mimics the old behavior, is to do `#ifdef SF_APPEND`.

P.S., Thank you for opening an issue and tying this PR to that issue. Too many people just create a PR.

catull: I am one of them.  Haven't always created an issue along with a PR.

catull: Applied your review finding's suggestion.
Awaiting further actions depending on the concern documented in the issue #309.

krader1961:
Merged with two corrections.

Issue #311, opened 2018-01-03
OpenSUSE & sysinfo


catull:
Jenkins builds for OpenSUSE have these log entries wrt. `sysinfo`:

```log
....
Checking for function "sysinfo": YES
....
../src/lib/libast/port/astconf.c: In function 'print':
../src/lib/libast/port/astconf.c:1118:9: warning: implicit declaration of function 'sysinfo' [-Wimplicit-function-declaration]
         if ((v = sysinfo(p->op, buf, sizeof(buf))) >= 0)
         ^
```
Since `meson` detects `sysinfo`, it therefore defines `-D_lib_sysinfo`.
In fact, the if-statement above is conditionally guarded:
```c
#if _lib_sysinfo
        if ((v = sysinfo(p->op, buf, sizeof(buf))) >= 0)
```
The compilation warning can be eliminated with
```diff
diff --git a/src/lib/libast/port/astconf.c b/src/lib/libast/port/astconf.c
index 353266f..67c62d8 100644
--- a/src/lib/libast/port/astconf.c
+++ b/src/lib/libast/port/astconf.c
@@ -36,7 +36,9 @@
 #include <proc.h>
 #include <ls.h>
 #include <sys/utsname.h>
-
+#if _lib_sysinfo
+#  include <sys/sysinfo.h>
+#endif
 #include "conftab.h"

 #ifndef DEBUG_astconf
```

floppym:
I'm pretty sure I already fixed this as a side effect of #393. No need another one.

krader1961:
Thanks, @floppym. It looks like that is true and this should have been closed when your PR was merged.

PR #312, opened 2018-01-03
Include <sys/sysinfo.h> if meson detects sysinfo()

catull:
This PR fixes issue #311.
The function sysinfo() must be included via <sys/sysinfo.h> if the constant _lib_sysinfo is defined.


catull:
This fix is not working at all.  Closing it.

Issue #313, opened 2018-01-03
src/cmd/ksh93/edit/edit.c:1666:12: warning: implicit declaration of function ‘ioctl’


catull:
The source file `edit.c` references `ioctl` without being properly included from <sys/ioctl.h>.

PR #314, opened 2018-01-03
Include <sys/ioctl.h>, which declares ioctl()

catull:
This PR fixes issue #313.
It makes sure the function is defined, the compiler will not issue a warning.


krader1961:
Merged. FWIW, Please read https://help.github.com/articles/closing-issues-using-keywords/ for how to format a commit message so that it will automatically close the issue when pushed upstream.

Issue #319, opened 2018-01-03
Warning: "Empty" redefined 


catull:
The file src/cmd/ksh93/sh/name.c includes the header `"defs.h"`.

Both of them define `Empty` almost identically, with some spaces/tabs differing.

Defined in `"defs.h"`:
```c
#define Empty			((char*)(e_sptbnl+3))
```

Defined in `name.c`:
```c
#define Empty ((char *)(e_sptbnl + 3))
```

The constant `Empty` is imported from `"defs.h"` in many more C source files.
In none of these C is `Empty` re-defined, except for in `name.c`.

The re-definition in `name.c` adds no benefit; it must be removed.

PR #320, opened 2018-01-03
Remove re-definition of Empty

catull:
This PR fixes the issue #319.
It makes sure the constant `Empty` is taken from "defs.h" and used without modification.


krader1961:
Merged

Issue #323, opened 2018-01-03
Missing realpath program on BSD, macOS


catull:
At least OpenBSD and macOS do not have a binary installed named `realpath`.
The system call `realpath` does exist on these platform.

Perhaps a shrink-down helper utility could be written, compiled and temporarily copied to bin/.
This utility could then be used in the scripts bin/*.sh.

catull:
Something along the lines of https://github.com/harto/realpath-osx.

krader1961:
For now I'm just going to modify the three scripts that use `realpath` to also use `grealpath` if it's available. That should be good enough for any reasonably modern BSD or Linux system. Although googling shows that old systems, such as Fedora 14, don't have it. If and when we need to build on those systems hopefully these bootstrap scripts won't be needed.

siteshwar:
According to [run_command](http://mesonbuild.com/Reference-manual.html#run_command) documentation it sets `MESON_SOURCE_ROOT` variable for source directory. We should be able to use it instead of determining path at runtime with `realpath`.

PR #324, opened 2018-01-03
Use $MESON_SOURCE_ROOT to determine source directory

siteshwar:
Meson exports $MESON_SOURCE_ROOT variable to scripts that are run
through run_command(). This commit removes need of realpath command to
determine source directory.

Resolves: #323


krader1961:
This is much better. LGTM.

Issue #333, opened 2018-01-05
Latest builds from GitHub for Fedora


siteshwar:
I have enabled builds from GitHub `master` branch in Fedora copr. Copr repository can be found [here]( https://copr.fedorainfracloud.org/coprs/g/ksh/latest/). This is the command to subscribe to the repository :

```
dnf copr enable @ksh/latest
```

For now I am triggering the builds manually, but I would like to setup a webhook to trigger the builds automatically on each commit. I am opening this issue to get some initial feedback about current setup.

krader1961:
I don't know enough about setting up a project on Fedora Copr or Suse OBS to be able to offer meaningful feedback. But I wanted to add a reminder to document this type of thing in the *CONTRIBUTING.md* document I just created.

siteshwar:
@krader1961 Good point. I will wait for feedback from community for current set up and update README.md and CONTRIBUTING.md once we have a stable set up.

siteshwar:
GitHub webhook is working now, so we should get latest builds from [copr repo](https://copr.fedorainfracloud.org/coprs/g/ksh/latest/).

PR #346, opened 2018-01-09
Add reference to Fedora copr repo in README

siteshwar:
Related: #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,20 @@ The `ksh` executable, helper libraries and man page can be installed with:
 ninja -C build install
 ```
 
+## Getting KornShell
+
+### Linux
+Latest builds for Fedora can be found in [copr](https://copr.fedorainfracloud.org/coprs/g/ksh/latest/).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,10 @@ The `ksh` executable, helper libraries and man page can be installed with:
 ninja -C build install
 ```
 
+## Latest Builds
```
krader1961: In the expectation additional platforms will be added in the future I'd recommend using the structure and verbiage from https://github.com/fish-shell/fish-shell/. See the "Getting fish" section. Obviously we'll want to name the section "Getting ksh". I'd even go so far as to include the "macOS" and "Windows" subsections with "TBD" or similar text to let people know we want to make it easy to get a bleeding edge ksh on those platforms.

krader1961:
LGTM. My suggestion doesn't have to be implemented so feel free to amend it or merge it as is.

siteshwar:
I have updated it.

PR #353, opened 2018-01-13
More sh.1 fixes and additions

dannyweldon:
cd section was appending rksh93
Added histexpand and -H
Removed duplicate -D
Some options were indented too far
Some reordering of options


hlangeveld:
LGTM
Mostly cosmetic, with a tiny change in content (for the better).

siteshwar:
LGTM too. Thanks!

dannyweldon:
Thanks.  I did not end up adding all the cli options because I realised that they were covered by:

       The remaining options and arguments are described under the set command
       above.  An optional - as the first argument is ignored.

PR #354, opened 2018-01-13
Make it possible to build `ksh` on OpenBSD 6.2

### LINKED ISSUE #276: Meson Build fails under OpenBSD-current
The project will not build under the latest version of OpenBSD-current (6.2+kernel patch number 310), here is the meson-log.txt:

[meson-log.txt](https://github.com/att/ast/files/1591139/meson-log.txt)

That attempt was using Clang (OpenBSD has switched to that as the default compiler) but it also fails with gcc.

Is OpenBSD supported?

They have the old nmake version of ksh93 in their ports tree:

http://openports.se/shells/ksh93

krader1961:
This makes it possible to build a working `ksh` binary on OpenBSD 6.2
and run the unit tests. Of course a record setting 52 tests fail but
it's a step in the right direction.

Resolves #276

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,6 +4,10 @@ run_command('bin/ksh93_prereq.sh')
 
 add_project_arguments('-D_PACKAGE_ast', language: 'c')
 add_project_arguments('-DMESON_BUILD=1', language: 'c')
+# This was added because OpenBSD installs a lot of stuff in /usr/local/{include,lib}
+# that would normally be in /usr/{include,lib}. But the compiler doesn't
+# automatically search /usr/local/included for headers so force it to do so.
+add_project_arguments('-I/usr/local/include', language: 'c')
 
 # We require the C99 (aka ISO9899:1999) standard. Note we use `gnu99` rather
 # than `c99` because the former enables language features we need. The latter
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -4,6 +4,10 @@ run_command('bin/ksh93_prereq.sh')
 
 add_project_arguments('-D_PACKAGE_ast', language: 'c')
 add_project_arguments('-DMESON_BUILD=1', language: 'c')
+# This was added because OpenBSD installs a lot of stuff in /usr/local/{include,lib}
+# that would normally be in /usr/{include,lib}. But the compiler doesn't
+# automatically search /usr/local/included for headers so force it to do so.
+add_project_arguments('-I/usr/local/include', language: 'c')
```
siteshwar: If it is only needed on OpenBSD then we should check for [host machine](http://mesonbuild.com/Reference-manual.html#build_machine-object) and put this code block under `if` condition.

krader1961: I'll do that but that still means adding the arg on other BSDs like FreeBSD. But that's probably still safer than doing it unconditionally for every platform. On the other hand there does seem to be a trend among newer distros to break from the past with respect to things like /usr/lib.

krader1961:
Merged.

Issue #177, opened 2017-12-02
Support for C_EU.UTF-8 locale in ksh93


siteshwar:
I could not find it documented anywhere, however there is one comment about this locale in the test cases :

```
# this locale is supported by ast on all platforms
# EU for { decimal_point="," thousands_sep="." }
locale=C_EU.UTF-8
```

krader1961:
From (among others) the *https://www.gnu.org/software/libc/manual/html_node/Locale-Names.html* page:

> Portability Note: With the notable exception of the standard locale names ‘C’ and ‘POSIX’, locale names are system-specific.

I am not aware of any non-AST implementation that provides support for "territory" extensions to the C or POSIX codeset. Given that Unicode encodings provide support for code points like the Euro currency symbol there should not be a need for the `C_EU.UTF-8` locale name. The `C` (or `POSIX`) locale implicitly does not support things like the `EU` territory for number formatting. The need for such support is why the misguided ISO-8859-x standards exist and were ultimately replaced by the Unicode standard.

Anyone wanting support for EU formatting of numbers can do so by setting the `LC_NUMBER` env var while leaving `LANG` set to `C`.

The bottom line is that ksh should use the same distro provided library code that almost every other program uses for localization. If only ksh93 recognizes `LANG=C_EU.UTF-8` what use is it?

kdudka:
How are ```decimal_point``` and ```thousands_sep``` used in ksh?

siteshwar:
It's used with builtins for e.g. 

```
$ echo $LANG
en_US.UTF-8
$ printf "%'d\n" 1000
1,000
$ export LANG=C_EU.UTF-8
$ printf "%'d\n" 1000   
1.000
```

krader1961:
Yes, but since it isn't a standard locale no other program will honor it. For example, if you do this on Ubuntu you get no thousands separator:

```
$ env LANG=C_EU.UTF-8 /usr/bin/printf "%'d\n" 1000
1000
$ echo $LANG
en_US.UTF-8
$ /usr/bin/printf "%'d\n" 1000
1,000
```

The `C_EU.UTF-8` locale is only meaningful inside ksh which is borderline useless and going to be the source of many bug reports due to the confusion it causes.

krader1961:
[...]

PR #355, opened 2018-01-14
Remove compatibility function for strtold

siteshwar:
Also disable a test case for C_EU.UTF-8 locale. This test case started
failing after removing strtold function.

See https://github.com/att/ast/issues/177 for discussion about
C_EU.UTF-8 locale.


krader1961:
LGTM

PR #356, opened 2018-01-14
Remove astsa functions from libast

siteshwar:
astsa provides a small collection of functions that can be used to make
a mini libast library. However we are using the full libast
implementation for ksh, so these functions are not needed.


krader1961:
I was wondering if this needed to be retained. Happy to see it go. LGTM.

PR #359, opened 2018-01-14
Fix syntax error in pty.sh

### LINKED ISSUE #358: Syntax error while running pty.sh test
The tests run on Fedora produced little diagnostic output until recently.

I ran all tests with `meson test -v`, the workspace is unmodified from commit 19d0620a72fe8c1b0ae43ae198f1f505fc9320c1.

This log statement appears:

```log
....
ast.git/src/cmd/ksh93/tests/pty.sh: line 32: syntax error at line 41: `(' unmatched
....
```

I am preparing a PR for it.

catull:
The second here-doc marker may only have tab characters preceding it.
The PR converts the space into a tab character.

It would also work if the `EOF` were at the beginning of the line.

This PR fixes #358.

### DIFF
--- a/src/cmd/ksh93/tests/pty.sh
+++ b/src/cmd/ksh93/tests/pty.sh
@@ -48,7 +48,7 @@ x=$( $SHELL <<- \EOF
         #sleep 1
         jobs
         kill $$
-    EOF
+	EOF
 )
 [[ $x == *Stop* ]] && err_exit 'monitor mode enabled incorrectly causes job to stop'
 


siteshwar:
If you want to fix your changes from previous commit, you should [amend](https://help.github.com/articles/changing-a-commit-message/) the commit instead of doing another commit. This helps in keeping git history clean. You have made 2 different commit messages for a single line change, it should be a single commit only.

catull:
Done.

siteshwar:
Thanks!

src/cmd/ksh93/tests/pty.sh
```diff
@@ -48,7 +48,7 @@ x=$( $SHELL <<- \EOF
         #sleep 1
         jobs
         kill $$
-    EOF
+	EOF
```
krader1961: FYI, A better fix would have been to leave the spaces and change the `<<-` to `<<#`. Now this file has a single invisible tab character.

catull: What is the policy on spaces vs tabs for the shell scripts ?
It is a mix of both everywhere.

siteshwar: I left tabs in `<<-` sections for readability. I was not aware about the ` <<# ` syntax. I think we can replace all occurences of `<<-` with `<<#` and replace all tabs with spaces. The only exception is the output of `typeset` command.

krader1961: Tabs cause more problems than they're worth. And mixing tabs and spaces is especially bad. We haven't established a policy for scripts. But the policy for the C source files is no tabs. I'd prefer that be the policy for the scripts as well.

catull: I switched from tabs to spaces 30 years ago and never looked back.
I am glad the consensus is to prefer spaces over tabs in this project.

Issue #335, opened 2018-01-07
What versioning scheme should we use?


krader1961:
Now that we are setting up automated builds (see issue #333) it is time to think about the format of the version string. One option is to continue with the current scheme which does not seem to be documented AFAICT. That is, strings like `93u+` (the version most commonly encountered on a distro) or `93v-` (the version at the time the project was open sourced). Presumably the letter was being incremented whenever a stable minor release point was reached. No idea what the minus or plus symbols represent.

A second option is to switch to [semantic versioning](https://semver.org). This would use a string of the form `major.minor.patch-level`; e.g., `93.15.0`. When building from any git commit without a version tag the current git commit hash would be appended to the most recent version tag. This is my preferred solution.

There are probably other options as well. Anyone with an opinion please feel free to comment.

krader1961:
Also, notice the somewhat unusual formatting of the version information:

```
$ src/cmd/ksh93/ksh --version
  version         sh (AT&T Research) 93v- 2014-12-24
$ echo $KSH_VERSION
Version BIJ 93v- 2014-12-24
```

This makes using that information to programmatically determine which version is being used, and thus which features are available, more difficult than it needs to be. Which is another reason I favor a simple semantic version string.

dannyweldon:
But if you echo them in an arithmetic context, you get just the plain number:

    $ echo $(( KSH_VERSION ))
    20141224
    $ echo $(( .sh.version ))
    20141224

In one way, this makes it easier to check for a particular feature by checking whether this number is `>=` a particular date stamp.  There is no reference available though for when features were added, so I had been making notes of a few from the old email list when I came across them and from the RELEASE file.  Perhaps I could start a wiki page for it.

It might be better to keep that as is, so that existing scripts don't break and the fact that `93 < 2014`, and create a new variable.  eg. `.sh.semver`, but then should that become an array like `$BASH_VERSINFO`?  Or, maybe use `$KSH_VERSINFO`?  (But I don't particularly like the name.)

I don't agree with starting a semver from `93` either.  That number should have been changed years ago.  There was a mention on the ast-developers lists that the internal company politics to get it changed was too hard, so they found it easier to just leave it.

I think for major versions:

ksh88 was version 1
ksh93 was version 2
next major release version 3 ?

[...]

PR #361, opened 2018-01-14
Update version number string in copr builds

siteshwar:
This string is displayed with 'ksh --version'. Use the same version
as used by rpm package.

### DIFF
--- a/.copr/prep.sh
+++ b/.copr/prep.sh
@@ -11,4 +11,6 @@ sed "s,#COMMIT#,${COMMIT},;
      s,#COMMITDATE#,${COMMIT_DATE}," \
          packaging/fedora/ksh.spec.in > packaging/fedora/ksh.spec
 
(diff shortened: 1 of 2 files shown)


.copr/prep.sh
```diff
@@ -11,4 +11,6 @@ sed "s,#COMMIT#,${COMMIT},;
      s,#COMMITDATE#,${COMMIT_DATE}," \
          packaging/fedora/ksh.spec.in > packaging/fedora/ksh.spec
 
+sed -i "s/#define SH_RELEASE.*/#define SH_RELEASE \"${COMMIT_DATE}+git.${COMMIT_NUM}.${COMMIT_SHORT}\"/" src/cmd/ksh93/include/version.h
```
krader1961: That seems needlessly verbose to me but it otherwise looks fine. And since this will probably be changed in response to issue #335 it's find for now.

PR #362, opened 2018-01-14
Change temp folder pattern to indicate which test script it is created for

### LINKED ISSUE #360: Tests create temporary folders which are difficult to associate with test scripts
The tests run on Fedora produced little diagnostic output until recently.

I ran all tests with `meson test -v`, the workspace is unmodified from commit 19d0620a72fe8c1b0ae43ae198f1f505fc9320c1.

When running all tests, these sort of lines are difficult to associate with the test affecting it:
```log
....
rm: cannot remove '/tmp/tmp.vUGNMNYPSf/a/b/c': Permission denied
....
rm: cannot remove '/tmp/tmp.11kcPtcerj/a/b/c': Permission denied
....
rm: cannot remove '/tmp/tmp.ffVNsoxJsZ/a/b/c': Permission denied
...
```

My conclusion is that the three failed removals of temporary directories happen in the EXIT `trap`.

It would be great if the temporary folders were created with a bit more context.
Such as
```sh
tmp=$(mktemp -dt ksh.${Command}-XXXXXXXX)
```

It would help in tracking problems; imagine the situation where you have dozens of temporary directories or files in /tmp/ matching `tmp.*` over time.

catull:
This PR addresses the difficulty to associate the temp folder created by the test scripts.
The temp folder names are unique to allow for repeated invocation, and for the invocation
to happen in parallel.

This PR adds a prefix "ksh." plus the test name in front of the temp folder name.
It fixes #360.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -195,7 +195,7 @@ foreach testspec : all_tests
         if locale != ''
             locale = '/' + locale
         endif
-        test(testname + locale, ksh93_exe, timeout: timeout, args: [test_path],
+        test(testname + locale, ksh93_exe, timeout: timeout, args: [test_path, testname],
              env: [shell_var, lang_var, ld_library_path])
     endforeach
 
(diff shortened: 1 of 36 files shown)


krader1961:
Merged. What I'd like to see at some point is moving all the temp files into the *build* directory so that all build/test artifacts are within that directory. But that's a very low priority change.

PR #364, opened 2018-01-14
Hyphenate the test names

### LINKED ISSUE #363: Can the test names please drop the parentheses ?
I keep forgetting to quote the parentheses on some of the tests.
It would be simpler to not have to type ....
```sh
meson test -v 'signal(shcomp)' 'signal(C.UTF-8)'
```
etc.

Why not `meson test -v signal signal-shcomp signal-C.UTF-8` ?

It would be even better if this was supported `meson test -v signal signal-s signal-c`, `-s` is for `shcomp` and `-c` is for `C.UTF-8`.

catull:
Separate test names and variant with a '/'

Make the entering of test names simpler.
Instead of entering 'signal(shcomp)' or signal\(shcomp\) etc, let the user
enter signal/shcomp.

Tis PR fixes #363.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -229,7 +229,7 @@ foreach testspec : all_tests
     foreach locale : locales
         lang_var = 'LANG=' + locale
         if locale != ''
-            locale = '(' + locale + ')'
+            locale = '/' + locale
         endif
         test(testname + locale, ksh93_exe, timeout: timeout, args: [test_path],
              env: [shell_var, lang_var, ld_library_path])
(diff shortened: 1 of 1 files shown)


krader1961:
You also need to modify how the names are formatted in the test output. I should be able to copy/paste the test name from a previous failed run. I also think using a slash as the separator is probably a better choice since it won't appear in the locale string.

catull:
Good point on using slashes between test run and variant.

This PR now allows entering `signal/shcomp`, the result is shown as:

`meson test -v alias alias/C.UTF-8 signal/shcomp` displays
```log
ninja: Entering directory `/usr/local/download/shell/korn/ast.git/build'
ninja: no work to do.
1/3 alias                                   OK       0.08 s
2/3 alias/C.UTF-8                           OK       0.08 s
3/3 alias/shcomp                            OK       0.13 s

OK:         3
FAIL:       0
SKIP:       0
TIMEOUT:    0
```

The change in `src/cmd/ksh93/meson.build` works for both entering AND displaying the result.

krader1961:
> ... both entering AND displaying the result.

Ah, yes, you're right. I could have sworn the test driver script constructed the name as well. Merged. Thx.

PR #368, opened 2018-01-16
Implement `bin/style` and `bin/lint` scripts

krader1961:
This implements the commands documented in a previous update to the
CONTRIBUTING.md document. The commands are currently implemented as fish
scripts since they are derived from the equivalent scripts I wrote for
the Fish shell project. TODO is rewriting them as ksh scripts.

On macOS `bin/lint --all` (which is shorthand for `bin/lint src/cmd/ksh93`) it produces a report over 7000 lines long. Most of which is from IWYU because the code depends too much on transitive includes.

### DIFF
--- a/.cppcheck.rules
+++ b/.cppcheck.rules
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+
+<![CDATA[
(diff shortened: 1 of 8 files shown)


siteshwar:
LGTM

PR #372, opened 2018-01-17
Stop generating bash_pre_rc.c on the fly

krader1961:
Since we're going to drop bash emulation take the first baby step.
Stop generating bash_pre_rc.c on the fly since build artifacts in the
source tree is a bad practice.


siteshwar:
LGTM

Issue #370, opened 2018-01-17
Universes and echo builtin


siteshwar:
`echo` builtin is broken because it's not parsing command line arguments. Here is a simple test:

```
$ echo -n Hello
-n Hello
```
However it is working normally under legacy `beta` branch. It is broken because it is trying to mimic behaviour of `att` universe intead of `ucb`.  Related code is [here](https://github.com/att/ast/blob/e98f9927a028e23d05821b890cfe3ac87fdb3a67/src/cmd/ksh93/bltins/print.c#L103-L109). The legacy [feature test](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/features/lib#L493-L503) for universe has been removed from `master` branch. I found this [background story](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819) behind this test. My understanding is that all the new operating systems that we want to support have the `ucb` (or `bsd`) behavior, so we should use it unconditionally and remove support for `att` behavior. Thoughts ?

krader1961:
This is a good example of why I want to remove all the universe stuff (issue #81). It shouldn't matter whether the currently selected universe is "att" or "ucb". The builtin `echo` should behave the same in either. If you want the universe to affect the behavior use the external command. Also, not this line from the stackexchange answer you linked to:

> POSIX says: if the first argument is -n or any argument contains backslashes, then the behavior is unspecified. 

And that's only the tip of the iceberg that is the mess that is the `echo` command.

Also, note that BSD and GNU echo supports `-n` to mean suppress the final newline. But GNU does not support `\c` while BSD does. On the other hand GNU supports `-e` while BSD does not. 

My vote is we modify the builtin echo to ignore the universe and support a hybrid of the two behaviors. That is, it recognizes `-e`, `-n`, and `\c`.

siteshwar:
Thanks @stephane-chazelas for the background story.

hlangeveld:
What's not mentioned in the story is that David Korn essentially gave up on `echo` and created `print`, which was not adopted elsewhere, afaik.

Later, `printf` was adopted by POSIX for the shell & utilities, and we used this to make [ryland/shpec](https://github.com/rylnd/shpec) work across all nominally POSIX conforming shells.

KeithBierman:
@henk all too true. BSD vs SysV bifurcation :<  It was one of the things
that pdksh did differently (on some platforms) than real ksh.
Keith Bierman
[...]

PR #373, opened 2018-01-17
echo builtin should always behave BSD style

siteshwar:
Older versions of ksh allow echo builtin to behave in either 'att' style
which does not recognize escape sequence and parameters, or 'ucb' (BSD)
style that can recognize escape sequences and parameters.

Since echo command behaves BSD style on the newer platforms, we should
use it unconditionally.

Resolves: #370

### DIFF
--- a/src/cmd/ksh93/bltins/print.c
+++ b/src/cmd/ksh93/bltins/print.c
@@ -93,33 +93,24 @@ static char *nullarg[] = {0, 0};
 // Builtin `echo`.
 //
 int B_echo(int argc, char *argv[], Shbltin_t *context) {
-    static char bsd_univ;
     struct print prdata;
     prdata.options = sh_optecho + 5;
-    prdata.raw = prdata.echon = 0;
+    prdata.echon = 0;
+    prdata.raw = 1;
     prdata.sh = context->shp;
     UNUSED(argc);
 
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/bltins/print.c
```diff
@@ -93,33 +93,24 @@ static char *nullarg[] = {0, 0};
...
-        char *universe;
-        universe = astconf("UNIVERSE", 0, 0);
-        if (universe) bsd_univ = (strcmp(universe, "ucb") == 0);
-        prdata.sh->universe = 1;
-    }
-    if (!bsd_univ) return b_print(0, argv, (Shbltin_t *)&prdata);
+    // This mimics BSD style echo command
+    // See https://github.com/att/ast/issues/370
     prdata.options = sh_optecho;
     prdata.raw = 1;
```
krader1961: I was worried that selecting the BSD variant would not support `-e` since the macOS external `echo` doesn't support it. Since the ksh BSD behavior does support it this change LGTM. However, I would hoist this to line 98 -- it's kind of silly to init this var twice:

```
prdata.echon = 0;
prdata.raw = 1;
```

rather than the existing

```
prdata.raw = prdata.echon = 0;
```

PR #374, opened 2018-01-18
Build the `pty` command.

krader1961:
This change enables building the standalone `pty` command that is needed
by one of the unit tests which otherwise "succeeds" because the command
can't be found.

===

@siteshwar, This breaks the Travis builds because they now find a `pty` command and try to run the associated unit tests. This won't be merged until I have either fixed those tests or explicitly commented them out with a "TODO" comment to re-enable them. Because this is already a fairly large change I wanted to get it out for review.

Note: This causes the same `pty` related test failures I've been seeing on my OpenSUSE VM (because it has a  */usr/bin/pty* command) on my other systems like macOS (which now has that command available to the relevant unit test).


src/cmd/builtin/meson.build
```diff
@@ -0,0 +1,66 @@
...
+                                     '../../lib/libast/features/',
+                                     '../../lib/libcmd/')
+
+lib_util_dep = cc.find_library('util', required: false)
+
+openpty_hdrs = '#include <stdlib.h>'
+if cc.has_header('pty.h')
+    pty_c_args += ['-D_hdr_pty']
+    openpty_hdrs += '\n#include <pty.h>'
+endif
```
siteshwar: nitpick: missing a newline

siteshwar:
LGTM

krader1961:
Merged

siteshwar:
`pty` command is installed by default. This is causing copr builds to fail with [this error](https://copr-be.cloud.fedoraproject.org/results/@ksh/latest/epel-7-x86_64/00703556-ksh/builder-live.log) :
```
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/bin/pty
```

Can we make installing `/usr/bin/pty` optional ? We do not ship this command in fedora or rhel, so I want to avoid packaging it.

krader1961:
> Can we make installing /usr/bin/pty optional ?

Absolutely, yes. The only reason I added support to build it was because it's included in OpenSuse 42.3 (Malachite) and it is the only command in that directory which is unique to this project and thus possibly deserves to be a builtin and a standalone command. Since what little I used to know about RPM is more than a decade old I'll leave it to you to figure out how to deal with this. Unless what you're suggesting is a way to build this project that doesn't result in that binary being built then I'm happy to dig into Meson/Ninja to figure that out.

siteshwar:
I am suggesting that we should make building and installing `pty`command optional in meson. I will look into how to do it.

krader1961:
@siteshwar, Unsurprisingly building just the `ksh` binary and nothing else is as simple as `ninja src/cmd/ksh93/ksh`.

siteshwar:
You can select targets while building, but `pty` will be built anyways when you try to install the targets. I have done https://github.com/att/ast/pull/378 to fix this.

PR #376, opened 2018-01-20
Fix ksh writing wrong byte sequence to terminal

siteshwar:
Description from bugzilla:

In vi mode when accessing via HP ilo4 virtual serial port, carriage
return and newfeed on input do not work properly. The cursor stays on
the same line, and sometimes the first few characters of the prompt are
missing. Things seem okay in emacs mode and when coming in via ssh or
directly via a physical serial port.

strace shows that the new ksh is writing \0\n where the old one just
writes \n.

The ksh issue is caused by it writing \0\n instead of just \n. User
wrote a kernel probe that intercepts writes and if it finds the new ksh
trying to write \0\n it replaces the \0 with a space. With that probe,
everything works fine. Without the probe, the problem returns.

Resolves: rhbz#1016611

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -1416,7 +1416,7 @@ Shell_t *sh_init(int argc, char *argv[], Shinit_f userinit) {
 #endif  // SHOPT_TIMEOUT
     // Initialize jobs table.
     job_clear(shp);
-    sh_onoption(shp, SH_MULTILINE);
+
     if (argc > 0) {
         int dolv_index = -1;
         // Check for restricted shell.


src/cmd/ksh93/sh/init.c
```diff
@@ -1416,7 +1416,7 @@ Shell_t *sh_init(int argc, char *argv[], Shinit_f userinit) {
 #endif  // SHOPT_TIMEOUT
     // Initialize jobs table.
     job_clear(shp);
-    sh_onoption(shp, SH_MULTILINE);
```
krader1961: This can't possibly be the correct fix. The statement being removed simply makes multiline editing the default behavior. Which is the default in the currently shipping version ksh93u+. Presumably if the user does `set -o multiline` when interacting via the problematic connection the problem will still occur. Is this the actual fix that RH has made to their ksh fork? Unless the person who created this change put a comment in the RH fork or the bugzilla issue that explains why this is correct I think we need to dig deeper.

siteshwar: This fix was written more than 4 years ago. But we can still ping @mhlavink who made this change. @mhlavink We would appreciate any comments about it.

krader1961:
FWIW, I fixed a similar [fish bug](https://github.com/fish-shell/fish-shell/issues/3740) involving the Linux console device and RS-232 connected terminals. The issue is that the tty's rows and cols values were zero which is obviously nonsense. I would not be surprised if something similar was at work here.

src/cmd/ksh93/sh/init.c
siteshwar: I have been told that it was part of a memory leak fix which @dgkorn added and the discussion was off list so there is no reference about it in the mailing list.

krader1961: Then I'm inclined to not merge this change. I don't see how changing the default setting of the option would fix a memory leak. There are several mentions of "leak" in *src/cmd/ksh93/RELEASE* but none that would obviously relate to this issue. But I would hope that if @dgkorn and company were aware of a leak that they fixed it in the code that was open sourced. Also, all of the `ksh` commands I have ready access to (which doesn't include any RH flavor) have the `multiline` option enabled by default. Is it really the case that `ksh` on RH distros disable this option by default?

siteshwar: This option was turned on by default with last stable release :

```
12-04-26 +Enabled multiline editing by default.  set +o multiline can disable.
```

On RHEL/Fedora it has been turned off by this patch. If you want to investigate this issue further I am fine with not merging it.

krader1961: Given that it's enabled by default in most distros and the fix is both dubious and for an obscure environment that 99.99% of ksh users will never encounter I'd prefer to determine the root cause of the problem and fix that. Anyone using ksh over a "HP ilo4 virtual serial port" can just do `set +o multiline`. Let's open an issue to remind us about the problem and the workaround.

krader1961:
Closing since I've opened issue #382 to remind us this may (or may not) still be a problem.

PR #378, opened 2018-01-22
Building pty command should be optional

siteshwar:
pty command is not shipped in RHEL or Fedora, so building and installing
it should be optional.

This fixes fedora copr builds which were broken by
74f72bde73794b7967810e4fcf136c770310c7bf

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,0 +1,1 @@
+option('build-pty', type : 'boolean', value : true)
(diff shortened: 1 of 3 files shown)


meson_options.txt
```diff
@@ -0,0 +1 @@
+option('build-pty', type : 'boolean', value : true)
```
krader1961: Cool. I see from the Meson docs we can also use `get_option()` to retrieve info like the installation prefix which might prove useful in the future.

krader1961: Also, this option name is fine for now but as other, optional, binaries are added to the build (e.g., `suid_exec`) it's not obvious if we'll want a config option for each program or options that manage groups of programs.

krader1961:
LGTM

Issue #379, opened 2018-01-23
Static or dynamic/shared libraries? Installed or not installed?


krader1961:
The legacy build system by default builds all the AST libraries as static libs. The new Meson build system builds them as dynamic/shared by default and tags them to be installed with the `ksh` binary. Of the systems I've been testing changes on macOS, FreeBSD, and Ubuntu use static libs and don't install them. OpenSuse and OpenBSD make them dynamic and install them. So the question is what should our default be and should we make it easy to choose the other option.

siteshwar:
I will vote for building static libraries by default.

dannyweldon:
> The legacy build system by default builds all the AST libraries as static libs.

My recollection with the `nmake` build system was that by default it gave me both static and shared libraries.

> I will vote for building static libraries by default.

Is there any reason?  I'm just asking.  :)

siteshwar:
@dannyweldon Because it's default on most of the systems and meson makes it very easy to change between building dynamic and static libraries.  All you have to do is change [default_library](http://mesonbuild.com/Reference-manual.html#library) command line parameter :

To build shared libraries:
```
meson -Ddefault_library=shared build; and ninja -C build
```

To build static libraries:
```
meson -Ddefault_library=static build; and ninja -C build
```

PR #381, opened 2018-01-23
Build static libraries by default

siteshwar:
Since most of the systems we care about build static libraries by
default we should do the same. For building dynamic libraries you
can manually set 'default_library=shared' user option on command
line:

meson -Ddefault_library=shared build

Resolves: #379


krader1961:
LGTM

PR #384, opened 2018-01-24
Replace several memcmp() calls with strncmp()

siteshwar:
While running code through asan I noticed that ksh uses memcmp() instead
of strncmp() to compare strings which can cause invalid reads.

Since memcmp() slightly varies in behavior (it does not stop comparing
on NULL) I have only replaced those memcmp() calls with strncmp() which
compare hardcoded ascii strings. Rest of the calls should be checked
and replaced with strncmp() if they only compare ascii strings.

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -1386,7 +1386,7 @@ static bool path_chkpaths(Shell_t *shp, Pathcomp_t *first, Pathcomp_t *old, Path
             }
             *cp = 0;
             m = ep ? (ep - sp) : 0;
-            if (m == 0 || (m == 6 && memcmp((void *)sp, (void *)"FPATH=", m) == 0)) {
+            if (m == 0 || (m == 6 && strncmp(sp, "FPATH=", m) == 0)) {
                 if (first) {
                     char *ptr = stkptr(shp->stk, offset + pp->len + 1);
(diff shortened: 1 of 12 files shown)


DavidMorano:
@siteshwar 
Thanks for finding this bug and fixing them where they might occur. If I remember, a customer was seeing crashes from |memcpy()|s running into non-mapped memory pages (a string was sitting right at the end of a mapped page). I was wondering if these fixes got into this latest clean-up effort (I do not follow this work as closely as I would like to).  I am glad that they are being included. Nice job in tracking this down and fixing these cases.

src/cmd/ksh93/sh/path.c
```diff
@@ -1386,18 +1386,18 @@ static bool path_chkpaths(Shell_t *shp, Pathcomp_t *first, Pathcomp_t *old, Path
             }
             *cp = 0;
             m = ep ? (ep - sp) : 0;
-            if (m == 0 || (m == 6 && memcmp((void *)sp, (void *)"FPATH=", m) == 0)) {
+            if (m == 0 || (m == 6 && strncmp((void *)sp, (void *)"FPATH=", m) == 0)) {
```
krader1961: Since you're touching this please remove the pointless cast to `void *` here and below.

krader1961:
Good catch. LGTM

PR #386, opened 2018-01-25
Fix an off by one write in history buffer

siteshwar:
NULL byte was written one byte beyond the buffer boundaries

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -455,7 +455,7 @@ static History_t *hist_trim(History_t *hp, int n) {
             if (newp <= oldp) break;
         }
         if (!(buff = (char *)sfreserve(hist_old->histfp, SF_UNBOUND, 0))) break;
-        *(endbuff = (cp = buff) + sfvalue(hist_old->histfp)) = 0;
+        *(endbuff = (cp = buff) + sfvalue(hist_old->histfp) - 1) = 0;
         // Copy to null byte.
         incmd = 0;
         while (*cp++) {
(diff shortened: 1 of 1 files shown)


krader1961:
LGTM given what I gleaned from a quick read of the SFIO code. I'm curious what caused you to notice this? ASAN?

siteshwar:
Yes, it was caught by ASAN.

siteshwar:
This caused builtin tests to fail because it modified the history buffer. I will probably revert this patch.

DavidMorano:
@siteshwar
OK, the tests are failing, but regardless: is the fix actually correct? Was there a bug where a NUL was being written one byte beyond the end of the buffer? With the bug in place, had you ever observed crashes from the bug, or other bad behavior (the written byte messing up some adjacent data structure)? Thanks for any additional information.

siteshwar:
> OK, the tests are failing, but regardless: is the fix actually correct? 

The fix was not correct and caused a test case to fail. So I will have to dig little more deeper into it.

> Was there a bug where a NUL was being written one byte beyond the end of the buffer? 

The bug is still there. I have reverted my commit.

> With the bug in place, had you ever observed crashes from the bug, or other bad behavior (the written byte messing up some adjacent data structure)? Thanks for any additional information.

I have not faced any crashes, but it can cause undefined behavior. I will check how it can be fixed.

DavidMorano:
@siteshwar
> The fix was not correct ...

Thanks very much for the additional information.  I very much appreciate it. Good luck in finding the source of the bug and implementing a fix.

PR #387, opened 2018-01-25
Fix an off by one write in history buffer

siteshwar:
NULL byte was written one byte beyond the buffer boundaries

My previous fix for this issue caused a regression in builtin tests.
I am fixing it now by allocating a new buffer which allocates an extra
byte to put NULL character.

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -397,7 +397,7 @@ static History_t *hist_trim(History_t *hp, int n) {
     char *cp;
     int incmd = 1, c = 0;
     History_t *hist_new, *hist_old = hp;
-    char *buff, *endbuff, *tmpname = 0;
+    char *buff, *endbuff, *tmpname = 0, *tmp = 0;
     off_t oldp, newp;
     struct stat statb;
 
@@ -455,6 +455,10 @@ static History_t *hist_trim(History_t *hp, int n) {
             if (newp <= oldp) break;
         }
         if (!(buff = (char *)sfreserve(hist_old->histfp, SF_UNBOUND, 0))) break;
+        // Reserve extra char to put NULL at end of buffer
+        tmp = malloc((sfvalue(hist_old->histfp) + 1) * sizeof(char));
+        memcpy(tmp, buff, sfvalue(hist_old->histfp));
+        buff = tmp;
         *(endbuff = (cp = buff) + sfvalue(hist_old->histfp)) = 0;
         // Copy to null byte.
         incmd = 0;
@@ -470,6 +474,7 @@ static History_t *hist_trim(History_t *hp, int n) {
         c = cp - buff;
         hist_new->histcnt += c;
         sfwrite(hist_new->histfp, buff, c);
+        free(tmp);
     }
     hist_cancel(hist_new);
     sfclose(hist_old->histfp);
@@ -486,7 +491,7 @@ static History_t *hist_trim(History_t *hp, int n) {
 // Position history file at size and find next command number.
 //
 static int hist_nearend(History_t *hp, Sfio_t *iop, off_t size) {
-    unsigned char *cp, *endbuff;
+    unsigned char *cp, *endbuff, *tmp = 0;
     int n, incmd = 1;
     unsigned char *buff, marker[4];
 
@@ -495,6 +500,10 @@ static int hist_nearend(History_t *hp, Sfio_t *iop, off_t size) {
     // with HIST_CMDNO.
     while ((cp = buff = (unsigned char *)sfreserve(iop, SF_UNBOUND, SF_LOCKR))) {
         n = sfvalue(iop);
+        // Reserve extra char to put NULL at end of buffer
+        tmp = malloc((n + 1) * sizeof(char));
+        memcpy(tmp, buff, n);
+        cp = buff = tmp;
         *(endbuff = cp + n) = 0;
         while (1) {
             // Check for marker.
@@ -527,6 +536,7 @@ static int hist_nearend(History_t *hp, Sfio_t *iop, off_t size) {
             if (n > 0) size += n;
             incmd = 0;
         }
+        free(tmp);
     }
 begin:
     sfseek(iop, (off_t)2, SEEK_SET);
@@ -541,7 +551,7 @@ begin:
 // it is followed by 0.  If followed by 0 then it cancels the previous command.
 //
 void hist_eof(History_t *hp) {
-    char *cp, *first, *endbuff;
+    char *cp, *first, *endbuff, *tmp;
     int incmd = 0;
     off_t count = hp->histcnt;
     int oldind, n, skip = 0;
@@ -557,6 +567,10 @@ again:
     sfseek(hp->histfp, count, SEEK_SET);
     while ((cp = (char *)sfreserve(hp->histfp, SF_UNBOUND, 0))) {
         n = sfvalue(hp->histfp);
+        // Reserve extra char to put NULL at end of buffer
+        tmp = malloc((n + 1) * sizeof(char));
+        memcpy(tmp, cp, n);
+        cp = tmp;
         *(endbuff = cp + n) = 0;
         first = cp += skip;
         while (1) {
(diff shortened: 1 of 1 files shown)


DavidMorano:
@siteshwar 
Thanks again for both finding this bug and studying it for a possible fix.

> I am fixing it now by allocating a new buffer which allocates an extra
byte to put NULL character.

Needless to say, but this appears to be the classical (not big enough size by one) buffer-allocation bug. This type of bug has seemed to plaque every generation of programmer for generations.  Maybe only relatively recently with the increased use of something like the C++ STL |string| object might the incidence of this type of bug finally diminish. But buffers are still allocated for many reasons other than classical string storage, so who can know how long this type of bug will be with us all into the future.

Thanks again for the good (extraordinary) work in finding these subtle bugs. I know that @krader1961  asked you this previously, but how did you ever come across this bug in the first place?

It would seem like the bug has likely been inside of KSH from its beginning. Apparently the data structure located in memory just after the buffer in question does not contain any information which upon destruction (by the writing of the NUL byte) ever became noticed.  Just curious here, but do you happen to know, by any chance, what the data element or structure is that was located just beyond the buffer in question? and what function that other element might have served within KSH?

It would be interesting to know why such a long-lived bug had not been detected, possibly by incorrect program behavior, before now. I can only guess that the memory element being corrupted by the NUL byte somehow was either low or no importance to the relatively correct operation of KSH.  But if that is the case, then the question about how you discovered this bug in the first place becomes even more intriguing?

Thanks for any additional information (when you have some free time) on either or both of how you discovered the bug, and possibly why the bug did not manifest itself in bad KSH behavior in all of these years (like by knowing what KSH memory element lied beyond the end of the buffer).

krader1961:
@DavidMorano, See [ASAN](https://en.wikipedia.org/wiki/AddressSanitizer). It's a tool that makes it more likely that invalid memory accesses, such as accessing the byte past the end of a buffer, will fail. Without such a tool such bugs tend to be extremely hard to reproduce. Padding by the memory allocator and the vagaries of the layout of the virtual address space means that such off by one errors are frequently harmless. But once in a million times it will happen the buffer ends on a page boundary with no adjacent page mapped in the address space so you get a SIGSEGV.

DavidMorano:
@krader19
Thanks for the information.
Yes, I am aware of ASAN but I was not aware of whether that is how @siteshwar found the bug. I do not remember seeing whether @siteshwar had answered your question on that (I seem to remember that you had asked him if he had used ASAN also).
I was wondering if the bug had manifested itself in some way, perhaps by breaking the correct functioning of the adjacent memory element. Either way, I am happy that one more bug is removed.

src/cmd/ksh93/edit/history.c
```diff
@@ -455,6 +455,10 @@ static History_t *hist_trim(History_t *hp, int n) {
             if (newp <= oldp) break;
         }
         if (!(buff = (char *)sfreserve(hist_old->histfp, SF_UNBOUND, 0))) break;
```
krader1961: `SF_UNBOUND` is the magic value -1. I admit I have no idea what `sfreserve()` is actually doing but since it's returning a pointer to the buffer what happens if you just tell it to reserve a buffer exactly the size we need? That is,

```
if (!(buff = (char *)sfreserve(hist_old->histfp, sfvalue(hist_old->histfp) + 1, 0))) break;
```

Note that there's no need to multiply by `sizeof(char)` since that is by definition one.

I suspect the bug is that the subsequent stuffing of null into the end of the buffer is violating the SFIO API and that there is probably a SFIO function that should be used to append a char (or string) to a stream that has been set to size `SF_UNBOUND` so that SFIO takes care of allocating the additional space. Which is why I want to see SFIO replaced with stdio. Who among us knows anything about SFIO?

DavidMorano: @krader1961 
> Note that there's no need to multiply by sizeof(char) since that is by definition one.

For historical information:
There was a time around about 1984, when AT&T UNIX (I think it was either AT&T System V Release 1 or 2) was ported to the Cray computer of that time. The C language compiler was a cheap port (meaning very simply done), and since the Cray did not have byte addressability, the compiler assigned each 'char' to its own addressable memory word on the Cray -- which at the time was 64-bits. So the sizeof(char) was 8! Hopefully that particular compiler port did not last long. I learned this at a talk given on the various ports of AT&T UNIX around then.  As an aside, AT&T UNIX SysV was (or had been) adopted by a number of other manufactures then, including HP and Amdahl.

krader1961: @DavidMorano, I'll bet you a paycheck I've used more idiosyncratic systems than anyone working on this project other than its founders. When I was a college student I used a CDC Cyber 760 with a 60 bit integer that used "truncated ASCII" (6-bit chars so no no lowercase chars and ten chars to an integer word). Which also used ones complement so it had both positive  and negative zero. I used a Harris H500 that had a three byte (24 bit) integer.

While I can appreciate the necessity of supporting such unusual platforms 20+ years ago that is no longer true. Those systems are no longer being sold. Which means we are no longer encumbered by them and can therefore simplify our code.

DavidMorano: @krader1961 
> I'll bet you a paycheck I've used more idiosyncratic systems than anyone working on this project other than its founders.

:-) :-) !! :-)

> ... CDC Cyber 760 ...

I think you made a typo and meant to say the "CDC Cyber 7600".

I worked on the CDC Cyber 6600 (the predecessor machine), almost identical w/ the 7600, w/ 60 bit words, 10 6-bit chars to a word, one's compliment! :-)  I was a great assembly language programmer for that machine! :-)

I also used the DEC PDP-8 with its 12-bit words! :-) :-)

We are -- sadly or not -- really showing our ages here on this forum! :-)

DavidMorano: @krader1961 
> Although maybe close, there is still no way I would take your bet! :-)

siteshwar: @krader1961 According to sfio man page (`src/lib/libast/man/sfio.3`):

```
After a sfreserve() call, whether or not it succeeds, sfvalue(f) gives the size of the available data block.
```
So it's not possible to call `sfavlue()` before `sfreserve()`.

siteshwar:
> Thanks again for the good (extraordinary) work in finding these subtle bugs. I know that @krader1961 asked you this previously, but how did you ever come across this bug in the first place?

@DavidMorano It was found through [ASan](https://github.com/google/sanitizers/wiki/AddressSanitizer).

src/cmd/ksh93/edit/history.c
krader1961: Okay, but the man page also says

> n < 0: sfreserve() attempts to get at least |n| bytes into the buffer.

And since `SF_UNBOUND` is -1 that means it is reserving one byte. What I suspect is happening is it returns a buffer whose size is one only if the buffer was currently zero length otherwise it returns a pointer to the existing buffer. And the person who wrote this code expected it to reserve one __additional__ byte to make room for the nul char. The only reason it's appending that nul is so that it can do

```c
        while (*cp++) {
            ;  // empty loop
        }
```

even if the old hist buffer does not contain a nul terminated string. That block of code should be rewritten so that it doesn't need to append the nul char. There isn't any reason we should pay the cost of an expensive malloc and buffer copy just to retain the existing, bad, logic for copying an existing history buffer from the old hist file to the new one.

Also, looking at the definition of `SF_UNBOUND` I don't understand why they're using that rather than just passing `-1`. Also, from reading the man page it looks like that should just be `sfreserve(f, 0, -1)` to get a pointer to the existing buffer that we can then call `sfvalue()` to get the length of it.

Finally, grep shows a lot of places that call `sfreserve()` with `SF_UNBOUND` and the first couple I looked at seem dubious. The man page implies that symbol should only be used with functions like `sfsetbuf()` and `sftmp()` to create an unbuffered file. I'm reasonably confident that all the places passing that to `sfreserve()` should either be passing a literal `-1` or they have the same bug as this block of code.

siteshwar: @krader1961 Ok, I will try to rewrite the patch.

PR #388, opened 2018-01-26
Implement the new `config_ast.h` header file

krader1961:
This is a precursor to replacing all of the compiler `-Dsome_symbol`
flags with `#define` directives in a global platform feature header.
This merely introduces the platform features config header and includes
it in every source module.


krader1961:
I've merged this since it is clearly superior to what we have been doing with respect to platform feature tests and recording the results. Whether or not details like the name of the files is optimal is something we can easily change if my choices are deemed suboptimal.

PR #389, opened 2018-01-27
Update changelog and echo builtin section in man page

siteshwar:
Related: #370

### DIFF
--- a/src/cmd/ksh93/ksh.1
+++ b/src/cmd/ksh93/ksh.1
@@ -6086,26 +6086,13 @@ or all active jobs if
 is omitted,
 when a login shell terminates.
 .TP
-\f3echo\fP \*(OK \f2arg\^\fP .\|.\|. \*(CK
-When the first
-.I arg\^
-does not begin with a \-, and
-none of the arguments contain a \e,
-then
-.B echo
-prints each of its arguments separated by a space
-and terminated by a new-line.
-Otherwise, the behavior of
-.B echo
-is system dependent
-and
-.B print
-or
-.B printf
-described below should be used.
-See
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/ksh.1
```diff
@@ -6086,26 +6086,13 @@ or all active jobs if
...
-described below should be used.
-See
-.IR echo (1)
-for usage and description.
+\f3echo\fP \*(OK \f3\-n\fP  \f3\-e\fP \*(CK \*(OK \f2 arg\^\fP .\|.\|. \*(CK
+echo builtin prints all of its arguments separated by space and terminated by new-line.
+.B -n
+will skip putting a newline character at the end of output. If
+.B -e
+is set, it will enable interpreting escape sequences.
```
krader1961: Note that the current implementation recognizes `\c` to mean ignore the rest of the text and don't output a newline but only if `-e` is used. That needs to be documented. On the other hand I feel like we should take @KeithBierman's advice and be compatible with mksh. From my comment in #370:

> I installed mksh version R50 2015/04/19 on OpenSuse. It's echo does not behave as was documented in the stackexchange post I read. And which caused me to write in a previous comment "WTF?" because it wasn't compatible with any of the common implementation. It is in fact reasonably sane. It does default to -e. It does support -n and \c. It does support -E. It does not support -- to stop parsing flags. All flags must precede non-flag arguments. The one oddity is that it recognizes \c as equivalent to -n unless you also specify the -E flag.

krader1961:
This is okay and can be merged but it is incomplete. Since it doesn't automatically resolve #370 I have no objection to merging it as is.

PR #391, opened 2018-01-28
Fix posix_spawn feature test

floppym:
Include <unistd.h> for _exit(2) and close(2).

Shorten argv since meson's run() does not pass any arguments.

### DIFF
--- a/features/posix_spawn.c
+++ b/features/posix_spawn.c
@@ -7,6 +7,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <unistd.h>
 
 // If it uses fork() why bother?
 #undef fork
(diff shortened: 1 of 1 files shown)


krader1961:
Thanks, this eliminates a bunch of compiler warnings. But on all my systems it still fails with a non-zero exit status. Are you getting a positive result and if so on which platform?

floppym:
I tested on Linux with glibc-2.26, which has a fully functional `posix_spawn` implementation. The meson output looks like this:
```
Checking if "posix_spawn exists and it works and its worth using" runs: NO (2)
```
The "NO" is deceptive; meson outputs "NO" for any exit status other than 0. However, for our purposes, the meaning is inverted.

If you review posix_spawn.c, you can see the following exit status indicators:

0: posix_spawn fails in some critical way, making it completely unusable.
1: posix_spawn works partially, producing a child process which exits with status 127 if the OS returns ENOEXEC.
2: posix_spawn is fully functional, returning an error if the OS returns ENOEXEC.

floppym:
Maybe we should take this opportunity to "fix" the exit status of this test? That would look something like:

0: fully functional
1: partly functional
>1: failure

floppym:
I'm closing this in favor of #392.

PR #392, opened 2018-01-29
Fix posix_spawn feature test

floppym:
Include <unistd.h> for _exit(2) and close(2).
Shorten argv since meson's run() does not pass any arguments.
Rework exit codes for better output from meson.

### DIFF
--- a/features/posix_spawn.c
+++ b/features/posix_spawn.c
@@ -7,6 +7,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <unistd.h>
 
 // If it uses fork() why bother?
 #undef fork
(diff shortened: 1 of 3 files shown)


src/lib/libast/misc/spawnvex.c
```diff
@@ -696,7 +696,7 @@ pid_t spawnvex(const char *path, char *const argv[], char *const envv[], Spawnve
                              | VEXFLAG(SPAWN_umask)
 #endif
                                  ))
-#if _lib_posix_spawn < 2
+#if _lib_posix_spawn > 0
```
krader1961: Instead of this just simplify it to `#if _lib_posix_spawn`. And in the *features/meson.build* set it to one if the feature test is successful.

features/posix_spawn.c
```diff
@@ -23,6 +24,10 @@ pid_t __fork(void) {
     return -1;
 }
 
+// Exits with the following status codes:
+//    0 : fully functional
+//    1 : partly functional
+// >= 2 : unusable
```
krader1961: The only place that tests the `_lib_posix_spawn` symbol only distinguishes between two and everything else. In other words  we don't need to distinguish between partly functional and unusable. So this can be simplified to zero and non-zero.

features/posix_spawn.c
```diff
@@ -36,47 +41,46 @@ int main(int argc, char **argv) {
...
     char tmp[1024];
-    if (argv[2]) {
+    if (argc > 1) {
         _exit(signal(SIGHUP, SIG_DFL) != SIG_IGN);
     }
     signal(SIGHUP, SIG_IGN);
     if (posix_spawnattr_init(&attr)) {
         printf("posix_spawnattr_init() FAILED\n");
-        _exit(0);
+        _exit(2);
```
krader1961: I like that you've added distinct exit statuses to help understand why the feature test failed.

features/posix_spawn.c
```diff
@@ -107,15 +110,15 @@ else if (!WIFEXITED(status)) {
 else {
     status = WEXITSTATUS(status);
     if (status == 127) {
-        n = 1;
         printf("ENOEXEC produces exit status 127 (GOOD)\n");
+        _exit(1);
     } else if (status == 99)
```
krader1961: Since you're touching this please also add the missing braces here and below.

src/lib/libast/misc/spawnvex.c
floppym: We actually only want this code path if posix_spawn is broken, so I changed it to `!_lib_posix_spawn` instead.

krader1961:
LGTM Merged. Looking forward to seeing more changes from you, @floppym. Especially since you're a member of one of the modern mainstream distros. We're interested in hearing from your team about problems this ksh reboot causes with respect to backward compatibility.

siteshwar:
Thanks for the fix @floppym

PR #393, opened 2018-01-29
Fix various implicit declarations

floppym:
This series of commits fixes the errors that occur when building with `CFLAGS="-Werror=implicit"`. Implicit declarations can cause nasty runtime bugs.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,6 +30,11 @@ if system == 'openbsd'
     add_global_arguments('-I/usr/local/include', language: 'c')
 endif
 
+# Error on implicit declarations.
+if cc.has_argument('-Werror=implicit')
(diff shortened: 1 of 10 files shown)


floppym:
This should fix #311.

siteshwar:
LGTM

meson.build
```diff
@@ -30,6 +30,11 @@ if system == 'openbsd'
     add_global_arguments('-I/usr/local/include', language: 'c')
 endif
 
+# Error on implicit declarations.
+if cc.has_argument('-Werror=implicit')
```
krader1961: Thanks for the cleanups. I especially like this addition.

krader1961:
Merged

PR #394, opened 2018-01-29
Call exit(3) instead of _exit(2) in posix_spawn feature test

floppym:
The exit(3) function will flush stdout before terminating the program,
which leads to more consistent output from the printf() calls.

With meson, stdout is redirected to meson-logs/meson-log.txt which means
that output is buffered in large chunks instead of being auto-flushed on
newlines when stdout is attached to a tty device.

### DIFF
--- a/features/posix_spawn.c
+++ b/features/posix_spawn.c
@@ -3,6 +3,7 @@
 #include <signal.h>
 #include <spawn.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


floppym:
I don't see any obvious reason to call `_exit(2)` instead of `exit(3)`, but please let me know if I am overlooking something.

krader1961:
There isn't any reason to use `_exit()` in this situation. The only time it should be used is in an error path where you explicitly do not want stdio buffers flushed and want to avoid similar side-effects. Such as on the child side of a fork() where there may be text buffered on the stdout stream. That doesn't apply to the *posix_spawn.c* code.

Merged

kdudka:
I cannot immediately evaluate whether it applies to this case or not, but the major difference between ```exit()``` and ```_exit()``` is that ```_exit()``` is an async-signal-safe function whereas ```exit()``` is not:

http://man7.org/linux/man-pages/man7/signal-safety.7.html

PR #395, opened 2018-01-29
Fix a memory leak in executing subshell if PATH is reset

siteshwar:
Also, added test case for same.

Resolves: rhbz#1324990

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -1613,8 +1613,13 @@ void path_alias(Namval_t *np, Pathcomp_t *pp) {
         Shell_t *shp = sh_ptr(np);
         struct stat statb;
         char *sp;
+        Pathcomp_t *old;
         nv_offattr(np, NV_NOPRINT);
         nv_stack(np, &talias_init);
+        old = np->nvalue.pathcomp;
+        if (old && (--old->refcount <= 0)) {
+            free(old);
+        }
         np->nvalue.cp = (char *)pp;
         pp->refcount++;
         nv_setattr(np, NV_TAGGED | NV_NOFREE);
--- a/src/cmd/ksh93/include/name.h
+++ b/src/cmd/ksh93/include/name.h
@@ -31,6 +31,7 @@
 #include <cdt.h>
 
 typedef int (*Nambfp_f)(int, char **, void *);
+struct pathcomp;
 
 // Nodes can have all kinds of values.
 union Value {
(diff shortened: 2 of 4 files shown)


src/cmd/ksh93/sh/path.c
```diff
@@ -1613,8 +1613,13 @@ void path_alias(Namval_t *np, Pathcomp_t *pp) {
         Shell_t *shp = sh_ptr(np);
         struct stat statb;
         char *sp;
+        Pathcomp_t *old = 0;
```
krader1961: s/0/NULL/ please (Yes, I know the existing code uses zero liberally where NULL makes the intent clearer but no sense adding more bad idioms :-)

src/cmd/ksh93/sh/path.c
```diff
@@ -1613,8 +1613,13 @@ void path_alias(Namval_t *np, Pathcomp_t *pp) {
...
         nv_offattr(np, NV_NOPRINT);
         nv_stack(np, &talias_init);
+        old = (Pathcomp_t*)np->nvalue.cp;
+        if (old && (--old->refcount <= 0)) {
+            free((void*)old);
```
krader1961: No need for the cast.

src/cmd/ksh93/sh/path.c
```diff
@@ -1613,8 +1613,13 @@ void path_alias(Namval_t *np, Pathcomp_t *pp) {
         Shell_t *shp = sh_ptr(np);
         struct stat statb;
         char *sp;
+        Pathcomp_t *old = 0;
         nv_offattr(np, NV_NOPRINT);
         nv_stack(np, &talias_init);
+        old = (Pathcomp_t*)np->nvalue.cp;
```
krader1961: Since `np->nvalue.cp` is a `const char *` why are you declaring `old` to be a `Pathcomp_t *` thus necessitating this cast?

src/cmd/ksh93/sh/path.c
krader1961: Also, no need to initialize it here since it isn't used before you assign to it below.

src/cmd/ksh93/sh/path.c
siteshwar: We need to read `old->refcount` before free'ing the pointer.

siteshwar: Done!

krader1961:
Thank you. LGTM. Now to fix the remaining 10,000 similar bogosities :smile:

PR #397, opened 2018-02-01
Skip adding -nostartfiles flag during compilation

siteshwar:
Due to a recent change in gcc, compilation breaks on Fedora rawhide. This
should be fixed by skipping -nostartfiles flag.

Resolves: rhbz#1540796
Related: rhbz#1540642


krader1961:
LGTM. What I don't understand is why this flag was ever even used by this project given its purpose. Which isn't applicable to a normal user command like `ksh` or any of the other libraries or commands built by this project.

Issue #399, opened 2018-02-05
Improve escape sequence handling in prompt


perette:
The code that calculates the prompt length takes into account some escape sequences, but not all.  One might use these for a highlighted prompt, for example.  Incorrect calculation causes vi mode (`set -o vi`) to behave erratically: with a multi-line statement, the cursor is moved to the second line earlier than it should, causing the cursor to incorrectly reflect the edit point.  Using j/k to travel back through the history also causes problems, including duplicate lines, moving back up the screen, or "eating the prompt" by moving the first line to the left.

The attached patch adds additional escape sequence recognition.  In revising the code, my citations were:

- [Xterm Control Sequences](https://www.xfree86.org/4.8.0/ctlseqs.html)
- The escape sequence from `tput sgr0`: 033   (   B 033   [   m
- The escape sequence from `tput rs2`:  033   [   !   p 033   [   ?   3   ;   4   l 033   [   4   l 033 &lt;

Additionally, the `rs2` sequence involves a '!'.  The prompt display was modified to avoid expanding '!' in an escape sequence to the command number, which breaks the escape sequence.

Please integrate this into the code base.

[ksh-prompt.txt](https://github.com/att/ast/files/1695874/ksh-prompt.txt)

siteshwar:
@perette Mind doing a pull request ? I can integrate this patch manually, but a commit on your name should give you due credit.

perette:
Not worried about credit, just hoping to fix the bug.

siteshwar:
@perette Merged requested patch. Thanks for your contribution.

PR #400, opened 2018-02-07
Improve escape sequence handling in prompt

siteshwar:
Thanks @perette for the patch
Related: #399


src/cmd/ksh93/sh/io.c
```diff
@@ -1990,14 +1991,20 @@ static int io_prompt(Shell_t *shp, Sfio_t *iop, int flag) {
 #endif /* TIOCLBIC */
             cp = sh_mactry(shp, nv_getval(sh_scoped(shp, PS1NOD)));
             shp->exitval = 0;
-            for (; (c = *cp); cp++) {
-                if (c == HIST_CHAR) {
+            for (escape_index = 0; (c = *cp); cp++) {
```
krader1961: Might as well put the declaration inside the `for` statement since the var isn't used outside the loop.

krader1961:
LGTM. I didn't actually test it but it looks more or less correct.

siteshwar:
Made requested changes and merged.

Issue #396, opened 2018-01-30
Replace AST Vmalloc with standard malloc


krader1961:
There is no good reason that `ksh` and related commands should depend on the AST Vmalloc subsystem. In theory just doing `-D_std_malloc` when building should use the standard malloc subsystem. But that obviously hasn't been tested in several years since it doesn't build.

I did all the mechanical transformations necessary to use standard malloc rather than vmalloc. Only one unit test fails with that change which was not failing before: `sh_match` (all three variants). The failure only happens sporadically. The failure is a SIGSEGV in this statement from line 3177 in *src/cmd/ksh93/sh/name.c*:

```
if (mp && (mp->nvname == NULL || *mp->nvname == 0)) mp = NULL;
```

Obviously that should never result in a SIGSEGV unless the data structure has been corrupted. Capturing a core file shows that `mp->nvname` is sometimes an invalid address such as 0x8. In fact, every core file I've examined has that pointer set to 0x8 (which is the ASCII BS character). Without a doubt there is an off by one bug somewhere that is causing memory to be corrupted. Something  likely to be papered over by the Vmalloc subsystem if it pads the blocks differently than the native malloc subsystem or otherwise alters how dynamically allocated blocks are arranged.

krader1961:
I just found core dumps with `mp->nvname` set to 0x6 and 0x7. So it's clearly not a case of stuffing a backspace char into a string buffer that is too short.

siteshwar:
You should be able to use system malloc() functions by setting `-D_AST_std_malloc=1` in `CFLAGS`.

krader1961:
> You should be able to use system malloc() functions by setting -D_AST_std_malloc=1 in CFLAGS.

Maybe, but that doesn't really address the issue. We need to eliminate the dependency on the AST Vmalloc code. No one is going to actively maintain that subsystem. It is a potential source of bugs. It is unlikely to be optimized for new architectures like ARM or PPC. It may not interact well with tools like Valgrind and Asan. It is a complicated interface that no one outside of the AST authors is familiar with. The fact that some of its APIs which mirror the UNIX malloc APIs also zeroes the allocated memory while the UNIX malloc version does not is a source of confusion and bugs. Vmalloc has to be eliminated as a ksh dependency.

[...]

PR #401, opened 2018-02-08
Enable malloc debug features

krader1961:
This enables malloc debugging features provided by various
implementations. This is only really useful when issue #396 is resolved
so that `ksh` no longer uses the AST Vmalloc subsystem. I'm recommending
merging this now to to make it obvious that any bugs involving the use
of the native malloc subsystem are caught sooner rather than later.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,27 @@ add_global_arguments('-D_GNU_SOURCE', language: 'c')
 # in an array that will be used by the feature tests.
 feature_test_args = ['-std=gnu99', '-D_GNU_SOURCE']
 
+# Enable malloc debug options for macOS, GNU, FreeBSD, and other
+# implementations. This isn't a replacement for Valgrind or Asan but is useful
+# for detecting problems without incurring the overhead of those tools.
+#
+# To use these add `--setup=malloc` to your `meson test` command.
+#
+malloc_debug_env = environment()
+# These env vars are recognized by GNU malloc.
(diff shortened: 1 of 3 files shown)


krader1961:
P.S., I tested this on several distros such as FreeBSD11. This change makes the `meson test sh_match` fail consistently whereas without the test seldom fails.

krader1961:
P.P.S., This results in FreeBSD 11 test failures going from 65 to 74. So apparently there are a lot more "use after free" or "malloc buffers should be initialized to zero", or similar, bugs in the code.

meson.build
```diff
@@ -15,6 +15,27 @@ add_global_arguments('-D_GNU_SOURCE', language: 'c')
...
 
+# Enable malloc debug options for macOS, GNU, FreeBSD, and other
+# implementations. This isn't a replacement for Valgrind or Asan but is useful
+# for detecting problems without incurring the overhead of those tools.
+#
+# To use these add `--setup=malloc` to your `meson test` command.
+#
+malloc_debug_env = environment()
+# These env vars are recognized by GNU malloc.
+malloc_debug_env.set('MALLOC_CHECK_', '3')
```
siteshwar: I think we should also set `MALLOC_PERTURB_` environment variable here. This would set all the free'd bytes to value specified by this variable, so it's easier to detect use after free cases.

krader1961: Done. Set to 165 (0xA5) to mimic the macOS behavior.

Issue #161, opened 2017-11-29
Investigate issues identified by the `bin/lint` command


siteshwar:
To run the build through ```clang-analyzer``` :

```
ninja scan-build
```

Currently there are more than 250 bugs that are identified by it.

krader1961:
This makes me so sad as I love ksh. On the other hand, only ~250 bugs? Heck we should be able to fix that many bugs in a few days. :smile:

On a more serious note this is why I want to make it trivial and routine to run the code through tools like `clang-format`, `oclint`, `cppcheck`, and `scan-build`. There should be zero warnings from the compiler or any of those tools. Let alone obvious errors being identified.

krader1961:
My PR#368 enables clang static analysis (what `ninja scan-build` does) as part of running `oclint`. As well as running the code through `cppcheck`. Just the *src/cmd/ksh93* code produces a report that is over 7000 lines long.

krader1961:
I'm closing this because it is now trivial to lint the code, including just the files in an uncommitted change or the most recently committed change. Simply type `bin/lint` or `bin/lint all`. There is no point keeping this issue open because there will always be lint warnings and errors. Which means this issue will never be closed.

PR #406, opened 2018-02-13
Fix dead increments detected by clang-analyzer

siteshwar:
Related: #161

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -2146,8 +2146,6 @@ static void env_init(Shell_t *shp) {
             nv_newattr(np, flag | NV_IMPORT | NV_EXPORT, size);
             if ((flag & (NV_INTEGER | NV_UTOL | NV_LTOU)) == (NV_UTOL | NV_LTOU))
                 nv_mapchar(np, (flag & NV_UTOL) ? e_tolower : e_toupper);
-        } else {
-            cp += 2;
         }
     }
 skip:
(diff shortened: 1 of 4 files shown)


src/lib/libast/tm/tmxmake.c
```diff
@@ -98,7 +98,6 @@ Tm_t *tmxtm(Tm_t *tm, Time_t t, Tm_zone_t *zone) {
     tm->tm_mday = x + 1;
     tm->tm_nsec = tmxnsec(t);
     tmfix(tm);
-    n += 1900;
```
krader1961: This is correct but I feel compelled to point out that the preceding logic on line 94 which subtracts one and line 95 is needlessly obtuse. Don't change those lines because the benefit doesn't justify the risk. I just felt slightly nauseous when I reviewed the code to ensure removing this line was okay.

src/cmd/ksh93/sh/init.c
```diff
@@ -2146,8 +2146,6 @@ static void env_init(Shell_t *shp) {
             nv_newattr(np, flag | NV_IMPORT | NV_EXPORT, size);
             if ((flag & (NV_INTEGER | NV_UTOL | NV_LTOU)) == (NV_UTOL | NV_LTOU))
                 nv_mapchar(np, (flag & NV_UTOL) ? e_tolower : e_toupper);
-        } else {
-            cp += 2;
```
krader1961: Please also change the statement 11 lines below to hoist the initial assignment to its own statement:

```
    if ((cp = nv_getval(SHELLNOD)) && (sh_type(cp) & SH_TYPE_RESTRICTED)) {
```

krader1961:
LGTM

PR #407, opened 2018-02-13
Fix undefined pointer cases detected by clang analyzer

siteshwar:
Related: #161

### DIFF
--- a/src/cmd/ksh93/sh/nvtype.c
+++ b/src/cmd/ksh93/sh/nvtype.c
@@ -1568,8 +1568,8 @@ Namval_t *nv_typeparent(Namval_t *np) {
 }
 
 void nv_checkrequired(Namval_t *mp) {
-    Namtype_t *dp, *dq = 0;
-    Namval_t *np, *mq, *nq;
+    Namtype_t *dp = NULL, *dq = NULL;
+    Namval_t *np = NULL, *mq = NULL, *nq = NULL;
     int i;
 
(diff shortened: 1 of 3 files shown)


src/cmd/ksh93/sh/nvtype.c
```diff
@@ -1568,8 +1568,8 @@ Namval_t *nv_typeparent(Namval_t *np) {
 }
 
 void nv_checkrequired(Namval_t *mp) {
-    Namtype_t *dp, *dq = 0;
-    Namval_t *np, *mq, *nq;
+    Namtype_t *dp = NULL, *dq = NULL;
```
krader1961: It's a shame C doesn't have something like C++11's `nullptr`. So while changing `0` to `NULL` has no practical effect we should be making such changes anywhere the value is meant to be interpreted as a pointer if only for clarity to a human reader.

krader1961: LGTM

DavidMorano:
> It's a shame C doesn't have something like C++11's nullptr. So while changing 0 to NULL has no practical effect we should be making such changes anywhere the value is meant to be interpreted as a pointer if only for clarity to a human reader.

Agreed. This may be a bit fanciful (but maybe not in the long term), but even though C does not have 'nullptr' like C++ , there could be a possible future conversion of the current code from being C into being C++ (someday).  Having the null pointer constant being represented as 'NULL' (as opposed to '0') makes it rather trivial to change those constants to 'nullptr' at some future time. It isn't as common as dirt, but I have changed a number of code modules or translation units from C to C++ in order to get the language advantages of C++ for a code or feature enhancement.

PR #410, opened 2018-02-13
Fix some use after free cases

siteshwar:
These were detected by clang analzyer.
Related: #161

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -672,6 +672,7 @@ static void *sh_coinit(Shell_t *shp, char **argv) {
     if (!(csp->coshell = coopen(NULL, CO_SHELL | CO_SILENT, argv[1]))) {
         free(csp);
         errormsg(SH_DICT, ERROR_exit(1), "%s: unable to create namespace", name);
+        __builtin_unreachable();
     }
     csp->coshell->data = (void *)csp;
     csp->name = (char *)(csp + 1);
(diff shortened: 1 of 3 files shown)


src/cmd/ksh93/sh/xec.c
```diff
@@ -672,8 +672,9 @@ static void *sh_coinit(Shell_t *shp, char **argv) {
     if (!(csp->coshell = coopen(NULL, CO_SHELL | CO_SILENT, argv[1]))) {
         free(csp);
         errormsg(SH_DICT, ERROR_exit(1), "%s: unable to create namespace", name);
+    } else {
+        csp->coshell->data = (void *)csp;
     }
-    csp->coshell->data = (void *)csp;
```
krader1961: This is only a partial fix. Notice the next statement uses the potentially freed pointer. My understanding is that calling `errormsg()` with `ERROR_exit(n)` (where `n` > 0) forces the shell to exit. So we can't reach this statement if `csp` was freed. It would be better to add a comment to that effect and which would suppress the use after free warning. The simplest would be to add `__builtin_unreachable();` after that `errormsg()` call. See https://stackoverflow.com/questions/25408795/tell-gcc-that-a-function-call-will-not-return. That, however, depends on GCC and Clang behavior that may not be present in other C compilers we need to support. So alternatively just add `exit(1);  // not reached`.

krader1961:
LGTM since I'm willing to accept the limitation that any compiler this project supports should support `__builtin_unreachable();`. At least during this transition period where we are trying to modernize the code.

PR #413, opened 2018-02-15
Check return values for setuid(), setgid() and setregid()

siteshwar:
Related: #161

### DIFF
--- a/src/cmd/ksh93/bltins/test.c
+++ b/src/cmd/ksh93/bltins/test.c
@@ -552,13 +552,24 @@ int sh_access(const char *name, int mode) {
             setreuid(shp->gd->euserid, shp->gd->userid) == 0) {
             mode = access(name, mode);
             // Restore ids.
-            if (shp->gd->userid != shp->gd->euserid) setreuid(shp->gd->userid, shp->gd->euserid);
+            if (shp->gd->userid != shp->gd->euserid) {
+                if (setreuid(shp->gd->userid, shp->gd->euserid) < 0) {
+                    // Restoring real user id failed, exit.
+                    error(ERROR_system(1), "setreuid(%d, %d) failed", shp->gd->userid, shp->gd->euserid);
+                }
+            }
             if (shp->gd->groupid != shp->gd->egroupid) {
-                setregid(shp->gd->groupid, shp->gd->egroupid);
+                if (setregid(shp->gd->groupid, shp->gd->egroupid) < 0) {
+                    // Restoring real group id failed, exit.
+                    error(ERROR_system(1), "setregid(%d, %d) failed", shp->gd->groupid, shp->gd->egroupid);
+                }
             }
             return mode;
         } else if (shp->gd->groupid != shp->gd->egroupid) {
-            setregid(shp->gd->groupid, shp->gd->egroupid);
+            if (setregid(shp->gd->groupid, shp->gd->egroupid) < 0) {
+                // Restoring real group id failed, exit.
+                error(ERROR_system(1), "setregid(%d, %d) failed", shp->gd->groupid, shp->gd->egroupid);
+            }
         }
     }
 
--- a/src/cmd/ksh93/sh/args.c
+++ b/src/cmd/ksh93/sh/args.c
@@ -35,6 +35,7 @@
 #include "path.h"
 #include "shlex.h"
 #include "terminal.h"
+#include <error.h>
 #if SHOPT_BASH
 #define BASHOPT "\374"
 #else
@@ -563,8 +564,14 @@ void sh_applyopts(Shell_t *shp, Shopt_t newflags) {
     if ((!sh_isstate(shp, SH_INIT) && is_privileged) ||
         (sh_isstate(shp, SH_INIT) && is_privileged_off && shp->gd->userid != shp->gd->euserid)) {
         if (!is_option(&newflags, SH_PRIVILEGED)) {
-            setuid(shp->gd->userid);
-            setgid(shp->gd->groupid);
+            if (setuid(shp->gd->userid) < 0) {
+                error(ERROR_system(0), "setuid(%d) failed", shp->gd->userid);
+                return;
+            }
+            if (setgid(shp->gd->groupid) < 0) {
+                error(ERROR_system(0), "setgid(%d) failed", shp->gd->groupid);
+                return;
+            }
             if (shp->gd->euserid == 0) {
                 shp->gd->euserid = shp->gd->userid;
                 shp->gd->egroupid = shp->gd->groupid;
--- a/src/lib/libast/misc/procopen.c
+++ b/src/lib/libast/misc/procopen.c
@@ -35,7 +35,7 @@
 
 #include <ast_tty.h>
 #include <ls.h>
-
+#include <error.h>
 /*
  * not quite ready for _use_spawnveg
  */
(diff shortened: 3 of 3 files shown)


src/cmd/ksh93/bltins/test.c
```diff
@@ -552,13 +553,27 @@ int sh_access(const char *name, int mode) {
...
+                    // Restoring real group id failed, exit.
+                    error(ERROR_system(1), "setregid(%d, %d) failed", shp->gd->groupid, shp->gd->egroupid);
+                }
             }
             return mode;
         } else if (shp->gd->groupid != shp->gd->egroupid) {
-            setregid(shp->gd->groupid, shp->gd->egroupid);
+            while (((status = setregid(shp->gd->groupid, shp->gd->egroupid)) < 0) && (errno == EAGAIN));
+            if (status < 0) {
+                error(ERROR_system(0), "setregid(%d, %d) failed", shp->gd->groupid, shp->gd->egroupid);
```
krader1961: Why is the failure of `setreuid()` above fatal and the failure of `setregid()` here not fatal? It seems to me the two cases are symmetrical and should both be fatal or non-fatal.

siteshwar: In the above case it fails while restoring real user id for the process.

krader1961: And here it fails restoring the real group id. How is that any less serious?

siteshwar: Fixed.

src/lib/libast/misc/procopen.c
```diff
@@ -485,12 +485,33 @@ Proc_t *procopen(const char *cmd, char **argv, char **envv, long *modv, int flag
...
         if (flags & (PROC_DAEMON | PROC_SESSION)) modify(proc, forked, PROC_sys_pgrp, -1, 0);
         if (forked || (flags & PROC_OVERLAY)) {
+            int status = 0;
             if ((flags & PROC_PRIVELEGED) && !geteuid()) {
-                setuid(geteuid());
-                setgid(getegid());
+                while (((status = setuid(geteuid())) < 0) && (errno == EAGAIN));
+                if (status < 0) {
+                    error(ERROR_system(0), "setuid(%d) failed", geteuid());
+                    return NULL;
```
krader1961: I can see reporting a non-fatal error but is it okay to return early? You're changing the behavior with the insertion of these `return` statements and it isn't obvious that is safe to do.

siteshwar: According to `src/lib/libast/include/proc.h`:
```
#define PROC_PRIVELEGED (1 << 9)  /* setuid(0), setgid(getegid())       */
```
This flag is set to execute a new process under a specific user id and group id. If we fail to honour it, I presume we should return back.

krader1961: I'm a little uncomfortable because while this change seems correct it is changing the logic.

krader1961:
LGTM given that this doesn't break anything and does potentially fix some highly unlikely corner cases.

src/cmd/ksh93/sh/args.c
```diff
@@ -563,8 +565,16 @@ void sh_applyopts(Shell_t *shp, Shopt_t newflags) {
     if ((!sh_isstate(shp, SH_INIT) && is_privileged) ||
         (sh_isstate(shp, SH_INIT) && is_privileged_off && shp->gd->userid != shp->gd->euserid)) {
         if (!is_option(&newflags, SH_PRIVILEGED)) {
-            setuid(shp->gd->userid);
-            setgid(shp->gd->groupid);
+            while (((status = setuid(shp->gd->userid)) < 0) && (errno == EAGAIN));
```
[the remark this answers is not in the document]
DavidMorano: @krader1961 
> Also, In my experience a tight loop like this is dangerous. In general there should be a usleep(100); or similar to rate limit the attempt.

Yes, I abstract this sort of thing (into a new subroutine) and rate-limit the re-try loop
(especially w/ EAGAIN) with an intervening sleep of some sort (actually one of my own).

Just a small note here. Although |usleep(3c)| is multithread-safe on the latest Linux,
it is *not* multithread-safe on many other UNIXi platforms.  The reason: It was never intended to be multithread-safe when it was first introduced.  And many platforms never went back and made it multithread-safe either.  Also, the main UNIXi standards bodies (forget exactly which now, X/Open, ?) have classified |usleep(3c)| as deprecated and not recommended for either new or continued use. They (the standards people) recommend the use of |nanosleep(3rt)| (part of the POSIX "real-time" feature set) instead of |usleep(3c)|.  Another option is to just write one's own (simple enough) using |poll(2)| for millisecond precision. In any event, the idea is to use a form of sleep that is multithread-safe.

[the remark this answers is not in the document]
krader1961: Good points, @DavidMorano, about `usleep()`. This obviously returns us to the earlier discussion about whether `ksh` should explicitly support multiple threads. I know there are people who argue that additional threads of control might be created without the knowledge of the program and thus every program has to assume it is multi-threaded. I'm not convinced by those arguments. If `ksh` and the APIs it relies on do not explicitly create threads, and its plugin API makes it clear that creating threads is not allowed, it seems okay to me to assume that there will be only one thread.

Having said that I agree that `ksh`, given it supports plugins that it has no control over and which might create threads given the plugin API does not forbid creating threads, has to assume there is more than one thread of execution. And therefore require any API it uses to be thread safe. Which makes me sad since it adds considerable complexity that would otherwise be unnecessary.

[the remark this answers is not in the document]
DavidMorano: > Which makes me sad since it adds considerable complexity that would otherwise be unnecessary.

:-)
I have been doing nothing but multithread-safe (and multithreaded) for about 36 years now, but I feel your pain! :-)  There was talk some years ago about moving KSH itself towards a multithread direction (explicitly), but little seemed to come out of that talk. I actually had some misgivings myself about that effort, knowing how difficult getting some things correct within a multithread environment can be. My own feeling is that I would like to see some other constructs within KSH be modernized and refactored before anything w/ internal multithreading moves forward.  But in the end analysis, making programs multithread-safe is not really the hardest thing (by far) out there. But one has to start somewhere.  One can start by making all new code multithread-safe, and then going though the rest of the code to make everything else multithread-safe. It has to start somewhere if one wants to achieve it. The particular difficulty with KSH is that it uses a good bit of the AST library, which itself is riddled with many multithread-UNSAFE subroutines. But the task is not insurmountable. Compiling *out* VMALLOC (itself neither multithread-safe nor fork-safe, both a real requirement for something so fundamental as a "malloc" subsystem) is a very important first step!

For myself, I have large programs (which can function either stand-alone or as ... KSH built-ins!) with unknown numbers of threads (both explicit and dynamically created background threads) performing a myriad of functions and services. The reality is (as I have stated probably several times previously) that all modern code should be written to assume unknown numbers of threads, both explicit threads and library-initiated background threads. I also have deep code stacks of thread-live modules, which even possibly call each other or even themselves in recursive loops. I have KSH built-ins that can call subroutines which eventually call back into other KSH built-ins, some of which might even be a previous built-in that ended up calling itself recursively through intervening layers of subroutines (being used as abstraction interfaces).  Further, I have a lot of dynamically loaded code modules (starting w/ |dlopen(3dl)| and ending w/ |dlclose(3dl)|) that also initiate background threads.  There is some real danger (much more than normally encountered) when using background threads from within dynamically loaded and unloaded modules, especially when all code w/ background threads also has to be fork-safe (some of which I have previously discussed in these spaces). But this is more for the brave of heart out there. Since it should *not* be assumed that KSH built-ins will not be unloaded after being loaded (unloaded like w/ a |dlclose(3dl)|). novice KSH built-in thread users should probably not attempt to create background threads (threads that continue to run past the exit of the built-in and giving control back to KSH proper) from within KSH built-ins. But we are getting ahead of ourselves. Let's start by compiling *out* AST VMALLOC (as I did years ago, but possibly with unknown hidden bugs as @krader1961 discovered); the removal of which being a prerequisite for really any threading direction people might want to take.

msekletar:
I don't think there is any value added by ```while``` loops that are also introduced in this PR. Such retry loops do appear in practice but in the different context, i.e. syscalls interrupted by signal (errno == EINTR). I'd drop the retry loops and just error out when syscall fails.

DavidMorano:
EAGAIN? 
Yes, I am a little bit curious myself on what systems can return EAGAIN on these |setxxuid(2)| calls. I think that most systems will not even return EINTR on these calls (let alone EAGAIN). EAGAIN is usually used for normal (?) busy conditions or temporary system resource wait conditions. For myself, I have never seen either EAGAIN or EINTR on these calls. Although some systems are a bit too EINTR happy, in my opinion, returning it more than really ever needed.

siteshwar:
I discussed these changes with @kdudka  and @msekletar and agree that the `EAGAIN` condition is unlikely to happen (at least on Linux) and added complexity of while loop is not justified. I will fix my changes.

kdudka:
@DavidMorano It is really a corner case but you can find a code path in up2date Linux source code where it returns ```EAGAIN```.  If ```alloc_uid()``` returns ```NULL```, then ```set_user()``` returns ```-EAGAIN```, which is subsequently returned by the ```setreuid``` syscall handler.

DavidMorano:
@kdudka
> It is really a corner case ...

Yep, I see it now. Temp resource failure on Linux (but I see their reasoning and it seems totally legit). It looks like, as far as I've searched, that only Linux is capable of this behavior. But maybe others will one day follow if they do some kernel redesign.
OK.  Thanks for the heads up. I will actually go ahead and update some of my wrapper code to handle this case (with a re-try loop and short sleep, and a time-out).

krader1961:
> ... the EAGAIN condition is unlikely to happen (at least on Linux) and added complexity of while loop is not justified.

Whether or not it is justified my main complaint is that handling it is a) not implemented consistently (even in the context of this PR) and b) where it is appropriate it needs a delay between syscalls and a limit on the number of retries. Both of which should be abstracted into a function that handles those details in a consistent manner.

Note that nearly every syscall could theoretically fail with errno==EGAIN on some UNIX like implementation (e.g., WSL, Windows 10 Subsystem for Linux). For 99.99% of those theoretical failure modes it will only occur during a full moon when a butterfly flaps its wings in China at precisely 1200 hours. Thus the added complexity to deal with EAGAIN cannot be justified by a program like `ksh` which isn't going to be used in aircraft/spacecraft avionics code. Obviously I'm being hyperbolic but the point is that some failure modes should just result in a "Sorry, I can't continue" error followed by the process exiting.

PR #414, opened 2018-02-15
Fix a buffer overflow in hist_eof() function

siteshwar:
There is an off by one write at the end of history buffer. Remove this
extra write and adjust code so that it does not require NULL character
at the end of buffer.


krader1961:
LGTM but I didn't spend as much time reviewing this as I would like too. Life is too short and I'm still trying to figure out why replacing vmalloc with malloc causes so many unit test failures. :smile:

krader1961:
>  I have been using a KSH that uses MALLOC instead of AST VMALLOC for some time now, so I am very interested to see how this turns out.

You're probably using a ksh based on the ksh93u+ (or earlier) release. I spent a couple of hours trying to make the legacy `2012-08-01-master` (the ksh93u+) version build and pass unit tests with the macOS system malloc. I gave up given how much of the project had build and test failures.

I would like to see your customizations that allow you to use the system malloc.

PR #415, opened 2018-02-19
Replace several memcmp() calls with strncmp()

siteshwar:
In character encodings supported by ksh, strings are terminated by
NULL character. So it should be safe to replace memcmp() calls with
strncmp().

### DIFF
--- a/src/cmd/ksh93/bltins/mkservice.c
+++ b/src/cmd/ksh93/bltins/mkservice.c
@@ -301,7 +301,7 @@ static char *setdisc(Namval_t *np, const char *event, Namval_t *action, Namfun_t
     Namval_t *nq;
 
     for (i = 0; (cp = disctab[i]); i++) {
-        if (memcmp(event, cp, n)) continue;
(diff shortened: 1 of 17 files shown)


krader1961:
This seems dubious to me. The `strncmp()` will be slower than `memcmp()` where we know the length. And if one of the strings happens to contain a nul before the end of the string as indicated by the length this change might produce different results. What prompted you to make this transformation?

siteshwar:
We have seen multiple cases of crashes in `memcmp()` calls due to out of bound reads. Character encodings that are supported by ksh should not have `NULL` that appears in between strings. So this change should be safe.

krader1961:
Okay, I see your point. While it's true that the limit is derived from calling `strlen()` on one of the strings it's not obvious the other pointer is guaranteed to point to a buffer at least that long.

LGTM

siteshwar:
I have updated the commit message to justify why I made this change. Referenced bugzilla is currently private. I should be able to make it public, but I need to ask reporter first.

PR #418, opened 2018-02-20
Add _ast_ prefix to glob related functions

siteshwar:
ASAN tries to replace calls to AST glob functions to glibc.
Implementation of AST glob functions slightly differs from glibc and
rest of the code depends on it. These functions should be renamed to
avoid any conflicts.


krader1961:
LGTM although I would just use a `ast_` prefix as that occurs much more often in the code than `_ast_`. The latter generally only being used for feature test symbols not function names.

PR #420, opened 2018-02-20
Fix return value from `sfraise()`

krader1961:
The `sfraise()` function must only return -1 or zero. Without this
change it can return one which breaks the `stkclose()` contract.

Partial fix for #389

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -2178,9 +2178,11 @@ void _nv_unset(Namval_t *np, int flags) {
                 }
                 dtclose(rp->sdict);
             }
-            // Note that stkclose() calls sfclose() which frees the stream.
             if (flags & NV_TABLE) {
-                while (stkclose(slp->slptr) == 1) {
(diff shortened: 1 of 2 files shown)


siteshwar:
It is causing `coprocess` test cases to fail.

krader1961:
Yes, because the coprocess test is doing something incredibly stupid. It's trapping SIGABRT and SIGIOT. The failure is not actually due to this change. It's due to the previous change which should have been innocuous but for the aforementioned bogosity. Fix in the pipeline.

krader1961:
This change reduces the test failures when using malloc from 16 to 12 on macOS and 51 to 46 on OpenSuse. So while there are still bugs in the code this fixes one failure mode. Note that the change to `sfraise()` is applicable to ksh93u+ even though the current behavior doesn't seem to cause any problems. That's because the ksh93u+ callers of `sfraise()` don't appear to check the return status. But `sfraise()` is still broken in that version it just doesn't cause any problems due to how it's used in that older release.

PR #423, opened 2018-02-23
Fix a use after free case in sh_setlist() function

siteshwar:


siteshwar:
This causes `types` test to fail. So I need to revisit this change.

PR #424, opened 2018-02-23
README.md: added details for Arch Linux packages.

Head-on-a-Stick:
I hope a PR is fine for this sort of thing, my apologies if this is wrong.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ ninja -C build install
 Latest builds for Red Hat Enterprise Linux (CentOS) and Fedora can be found in
 [copr](https://copr.fedorainfracloud.org/coprs/g/ksh/latest/).
 
-TBD packages for other Linux distros.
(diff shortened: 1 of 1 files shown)


krader1961:
> I hope a PR is fine for this sort of thing, my apologies if this is wrong.

A PR is the right thing to do when suggesting an improvement to a project whether its to the source code or its documentation (with the caveat that sometimes opening an issue first is a good idea). However, in light of the multiple, significant, use-after-free bugs we've recently found I don't think we want to encourage people to use a `ksh` built from the current `master` branch (or its progenitor `2016-01-10-beta`). The ksh93v- release simply has too many serious flaws in my opinion to be used by anyone unwilling to build from source.

I have proposed to the Red Hat representative, @siteshwar, that we rebase all the changes in the past few months on the `2012-08-01-master` branch (ksh93u+). In part because it seems very unlikely any distro will adopt the ksh93v- version even if we manage to fix all the significant bugs that have been identified. Including new builtins of common external commands where the builtin alternative is enabled unconditionally. However it's not clear anyone is really interested in doing all of that work.

Head-on-a-Stick:
OK, thanks for the information, I will go ahead and close this then.

My AUR packages are marked *-git* and so will be understood to be development versions, although I do have some people already using it.

Issue #398, opened 2018-02-05
There is a "use after free() bug"


krader1961:
This is a spin-off of issue #396. Specifically this [comment]( https://github.com/att/ast/issues/396#issuecomment-362884628). I'm opening a distinct issue because whether or not the dependency on AST Vmalloc is ever removed there is a significant "use after free()" bug in the code. Specifically, the `sh_match` unit test fails because the buffer pointed to by `np->nvenv` is freed before it is used in `nv_name()`. This bug is normally masked by the behavior of the AST Vmalloc subsystem which makes reusing the freed block unlikely before its contents are referenced again.

Addendum to the above: Per my [comment](https://github.com/att/ast/issues/398#issuecomment-368278588) made a few minutes ago there appear to be at least three use-after-free bugs even when using AST vmalloc. And those bugs are not the result of any of the changes made in the past few months since the beta (ksh93v-) branch became the master branch, the build system was changed, code was restyled, etc.

krader1961:
From issue #396 since the info really belongs here:

If I stub out the `free()` function so it is a no-op I am unable to reproduce the problem. So I am now 100% certain this is a "use after free" bug. In fact, debugging prints shows the structure whose contents are corrupted being freed:

```
WTF free 0x7fe5f4c146f0
0   nv_delete + 1037
1   array_putval + 790
2   nv_putv + 520
3   _nv_unset + 1113
4   table_unset + 637
5   sh_unscope + 74
6   sh_funscope_20120720 + 2550
7   sh_funct + 770
8   sh_exec + 12202
9   exfile + 3351
10  sh_main + 4345
11  main + 38
12  start + 1
```

And shortly thereafter the address is used in nv_name() when it is assigned to mp thusly:

```
     if (!nv_isattr(np, NV_MINIMAL | NV_EXPORT) && np->nvenv) {
        Namval_t *nq = shp->last_table, *mp = (Namval_t *)np->nvenv;
```

Because the structure pointed to by address 0x7fe5f4c146f0 wasn't freed due to my stubbing out free() that pointer (and the pointers in the Nameval_t it points to) are still valid.

krader1961:
I've found a place that defines a stack local var but doesn't initialize the contents to zero in *src/cmd/ksh93/sh/nvtree.c*, `nv_diropen()`:

```c
Namval_t *nq = 0, fake;
```

[...]

PR #427, opened 2018-03-04
Implment AST vmalloc buffer poisoning

krader1961:
This is to help debug issue #398 without going all the way of replacing
AST vmalloc with the system malloc. It causes buffers allocated (other
than via `calloc()`) and freed to be poisoned with a known pattern that
will result in a SIGSEGV if it is used as a pointer.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -22,6 +22,8 @@ feature_test_args = ['-std=gnu99', '-D_GNU_SOURCE']
 # To use these add `--setup=malloc` to your `meson test` command.
 #
 malloc_debug_env = environment()
+# These env vars are recognized by AST vmalloc.
+malloc_debug_env.set('VMALLOC_OPTIONS', 'junk')
 # These env vars are recognized by GNU malloc.
 malloc_debug_env.set('MALLOC_CHECK_', '3')
 malloc_debug_env.set('MALLOC_PERTURB_', '165')  # 0xA5
--- a/src/lib/libast/vmalloc/malloc.c
+++ b/src/lib/libast/vmalloc/malloc.c
@@ -354,6 +354,9 @@ void _vmoptions(int boot) {
                                 _Vmassert &= ~b;
                         } while ((v = strchr(v, ':')) && ++v);
                     break;
+                case 'j': /* junk */
+                    if (boot & 1) _Vmassert |= VM_junk;
+                    break;
                 case 'k': /* keep */
                     if (boot & 2) _Vmassert |= VM_keep;
                     break;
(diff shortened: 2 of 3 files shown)


krader1961:
Yes, this causes a lot of unit test failures (46 or more new failures). But that is a good thing. Because it makes it crystal clear the code involving manipulation of ksh vars is seriously flawed. I was going to make this poisoning behavior the default but decided to make it opt-in for consistency with similar debugging options. No other work should be done on this branch until the bug(s) this change brings to light are fixed.

krader1961:
Also, I looked at every single core dump produced by this change. Not a single one implicated using an uninitialized buffer from `malloc()` or `realloc()`. They all showed the poison for a freed buffer.

DavidMorano:
@krader1961 
> No other work should be done on this branch until the bug(s) this change brings to light are fixed.

Thanks very much for your attention to this matter.

krader1961:
Thanks for the positive feedback, @DavidMorano. We know there are at least two bugs because I've already fixed one (commit 4b303db2) and there are still a lot of tests failing. The stack traces suggest there are at least two more bugs. This would be a lot easier to debug if we had a fine grained set of changes between git commit dc9a3f29 (ksh93v-) and 3f54fd6 (ksh93u+). That would make it relatively easy for someone like me to pinpoint the problematic changes even without a good understanding of how source code.

@lkoutsofios, Is there any chance we can get ahold of the individual changes between those two releases? Even just the changes that touch files in the *src/cmd/ksh93/sh* directory is probably sufficient. That's because my investigation so far strongly suggests that the bug(s) are in the *src/cmd/ksh93/sh/name.c* and closely related modules.

siteshwar:
I am fine with this change, but enabling it by default will cause problems while testing other pull requests. I think we should keep it off by default.

krader1961:
Merged with a change to *.travis.yaml* to disable the malloc poisoning tests in Travis. I don't think we should merge another change until the use-after-free bugs are fixed. That is, `meson test --setup=malloc` passes on the current Travis platforms. These use-after-free bugs are so significant that pretty much any non-trivial change should be put on hold until the aforementioned bug(s) are fixed.

DavidMorano:
> These use-after-free bugs are so significant that pretty much any non-trivial change should be put on hold until the aforementioned bug(s) are fixed.

For what it is worth, this is my instinct also (find and fix these as soon as reasonably possible).  These use-after-free bugs could be doing some very bad damage under some particular use or test scenario! It is obviously too late now, but it would have been nice if Korn, or Fowler, or someone at AST or other had routinely tested with other versions of MALLOC facilities all along, both regular MALLOC facilities and special test-type MALLOC facilities (numerous enough to forego mentioning them).

PR #432, opened 2018-03-12
Don't call vfork() in OpenBSD

kernigh:
The child of vfork() uses shared memory to pass the errno from
execve() to the parent. OpenBSD's vfork() doesn't share memory, so the
errno got lost. This caused `chmod +x script; ./script` to fail if the
script had no `#! line`. The parent didn't see ENOEXEC from the child,
then didn't run the script.

Add a feature test for vfork(). Add back the define `_lib_vfork`; it
had disappeared from spawnvex.c in git commit ed6ba1d. If vfork()
doesn't share memory, then don't call vfork(). This drops the number
of failing tests in OpenBSD from above 50 to about 28.

### DIFF
--- a/features/vfork.c
+++ b/features/vfork.c
@@ -1,0 +1,45 @@
+// This is for the Meson config stage to check if vfork() acts like
+// ksh expects.  We need vfork() to pause the parent and share memory
+// with the child until the child execs or exits.
+//
+// These systems do pause the parent and share memory:
+// - FreeBSD: https://www.freebsd.org/cgi/man.cgi?vfork
+// - Illumos: https://illumos.org/man/vfork
+// - Linux: http://man7.org/linux/man-pages/man2/vfork.2.html
+// - NetBSD: http://netbsd.gw.com/cgi-bin/man-cgi?vfork
+//
+// This system doesn't share memory:
+// - OpenBSD: https://man.openbsd.org/vfork
+
+#include <errno.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int main(int argc, char **argv) {
+    pid_t pid;
+    int m;
+    volatile int exec_errno;
+    volatile int *volatile exec_errno_ptr;
+
+    // Mimic how libast/misc/spawnvex.c calls vfork().
+    exec_errno = 0;
+    exec_errno_ptr = &exec_errno;
+    pid = vfork();
+    if (pid == -1) {
+        return 1;  // fail, can't fork
+    } else if (!pid) {
+        // As child, pretend to exec something.
+        *exec_errno_ptr = ENOEXEC;
+        _exit(126);
+    }
+
+    // As parent, read errno from child, immediately after returning
+    // from vfork() and before calling waitpid().
+    m = *exec_errno_ptr;
+    while (waitpid(pid, NULL, 0) == -1 && errno == EINTR) continue;
+    if (m == ENOEXEC) {
+        return 0;  // pass
+    } else {
+        return 1;  // fail, got wrong errno
+    }
+}
--- a/src/lib/libast/misc/spawnvex.c
+++ b/src/lib/libast/misc/spawnvex.c
@@ -704,20 +704,26 @@ bad:
         int m;
         Spawnvex_noexec_t nx;
         int msg[2];
+#if _lib_vfork
         volatile int exec_errno;
         volatile int *volatile exec_errno_ptr;
+#endif
 
         if (!envv) envv = environ;
+#if _lib_vfork
         if (!vex || !(vex->flags & SPAWN_FORK))
             nx.flags = SPAWN_VFORK;
         else
+#endif
             nx.flags = SPAWN_FORK;
         n = errno;
+#if _lib_vfork
         if (nx.flags & SPAWN_VFORK) {
             exec_errno = 0;
             exec_errno_ptr = &exec_errno;
             msg[0] = msg[1] = -1;
         } else
+#endif
 #if _lib_pipe2
             if (pipe2(msg, O_CLOEXEC) < 0)
             msg[0] = msg[1] = -1;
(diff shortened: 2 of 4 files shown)


features/vfork.c
```diff
@@ -0,0 +1,45 @@
...
+// - Linux: http://man7.org/linux/man-pages/man2/vfork.2.html
+// - NetBSD: http://netbsd.gw.com/cgi-bin/man-cgi?vfork
+//
+// This system doesn't share memory:
+// - OpenBSD: https://man.openbsd.org/vfork
+
+#include <errno.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
```
siteshwar: This test is different from the [legacy feature test](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/features/lib#L242-L269) for vfork(). I presume this means legacy test was not testing desired functionality on OpenBSD.

siteshwar: Ok, I see [similar legacy feature test](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/features/lib#L271-L285) that defines `_real_vfork` macro.

kernigh: I didn't know about the old feature test. It looks simpler than mine, and might have worked.

I didn't want to bring back the old code that had too many macros (at least _lib_vfork, _real_vfork, _hdr_vfork, _sys_vfork, _lib_fork). Instead, I made _lib_vfork like _lib_posix_spawn. There's only _lib_posix_spawn = 1 if we use it, or _lib_posix_spawn = 0 if we don't.

krader1961:
I'm currently sick with the worst case of food poisoning I've had since I worked overseas 20 years ago. Hopefully I'll feel like doing more than lay on the couch in between visits to the porcelain god by tomorrow and will take a close look at this and the issues you've opened, @kernigh.

features/vfork.c
siteshwar: I think it makes sense to restore [old feature test](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/features/lib#L271-L285) than writing a new one.

src/lib/libast/misc/spawnvex.c
```diff
@@ -704,20 +704,26 @@ pid_t spawnvex(const char *path, char *const argv[], char *const envv[], Spawnve
         int m;
         Spawnvex_noexec_t nx;
         int msg[2];
+#if _lib_vfork
```
siteshwar: Legacy code keeps these code blocks under ```_real_vfork``` macro. If there are any regressions around this code, it will make debugging easier if we rename this to ```_real_vfork```.

siteshwar:
We probably also need to restore [this code block](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/misc/spawnvex.c#L143-L150).

siteshwar:
@krader1961 I am pondering if we should get rid of `vfork()` altogether and use only `fork()`. What are your thoughts about it ?

DavidMorano:
@siteshwar 
> I am pondering if we should get rid of vfork() altogether and use only fork(). 

Maybe it was not the best decision (do not really know), but that was my decision in order to get a working (non-hanging) solution on some OSes -- i.e. modifying the code to only use |fork(2)|.

As I have discussed at length elsewhere, the use of VFORK is very nuanced and behaves slightly differently on different OSes. I am rapidly coming to the opinion that one should let the OS vendor use (or not use), as they may like, VFORK in their implementation of |posix_snawn()| and |posix_spawnp()|. We, humble developers, can then decide to use either |posix_spanwnX()| or |fork(2)| when the former is not available.

KeithBierman:
For ksh's uses, isn't spawn (there's a posix flavor) "really" what was
originally intended? Or are there clever bits which do stuff between the
(v)fort/exec to do special stuff?

Keith Bierman
khbkhb@gmail.com
303 997 2749

On Wed, Mar 14, 2018 at 6:02 AM, David A.D. Morano <notifications@github.com

kernigh:
> We probably also need to restore [this code block](https://github.com/att/ast/blob/e770c77e9816e156c6df4a455e71b5f9fff79310/src/lib/libast/misc/spawnvex.c#L143-L150).

That code block was inside `#if DEBUG_spawn`. I believe that DEBUG_spawn is never enabled, so I didn't touch that code. A comment says, "DEBUG_spawn provides an emulation of ibm i5/OS spawn()". An earlier comment says, "the DEBUG_* emulations allow { cygwin ibm } debugging on linux".

----

I made another branch [kernigh-vfork-2](https://github.com/kernigh/ast/tree/kernigh-vfork-2) with a commit 39af3ef. This version uses the old feature test for `_real_vfork` that @siteshwar found.

I would have deleted 12005e8 and added 39af3ef to this pull request. I didn't do that because [the Travis build of 39af3f](https://travis-ci.org/kernigh/ast/builds/353498391) has failed. OpenSUSE passed, but Fedora failed with, `Error: Failed to synchronize cache for repo 'fedora'`.

krader1961:
According to APUE third edition p. 234 `vfork()` is obsolete and no longer appears in the SUS v4 standard. Given that we can no longer blindly depend on the child and parent sharing the same address space after calling `vfork()`, something that leaves me doing a /face-palm, I vote for simply dropping support for `vfork()`. I'd prefer to see the code simplified to just fork and the spawn family of functions.

@kernigh, Thanks for the work you've done so far. I hate to ask you to throw it away but I really think the project is better off by dropping all support for vfork. Had I realized that the original `vfork()` semantics can no longer be relied upon I would have made the same recommendation when commit ed6ba1d was proposed.

DavidMorano:
@krader1961 
>  I'd prefer to see the code simplified to just fork and the spawn family of functions.

Agreed: fork and spawn family of functions (as I previously also advocated).
VFORK is both dangerous and too nuanced across OSes to continue playing with it in the modern era.
No one has time to research all of the various semantics across OSes. Who needs that hassle in the modern era? We are all too busy! Rather, let each OS implement their own |posix_spawnX()| in their own most efficient way. If the OS coders secretly use a hidden VFORK under the hood, that is their business and they already know how their own secret VFORK works on their own OS anyway. Life is too short to put so much effort into (in the end) a simple decision -- do not use VFORK.

Time is better spent in working out the bugs in getting rid of (compiling out) AST VMALLOC!

krader1961:
Closing in favor of simply removing all support for `vfork()`.

PR #438, opened 2018-03-19
Fix a use after free case

siteshwar:
Related: rhbz#1460940


src/cmd/ksh93/sh/subshell.c
```diff
@@ -664,6 +664,7 @@ Sfio_t *sh_subshell(Shell_t *shp, Shnode_t *t, volatile int flags, int comsub) {
         shp->st.otrap = 0;
         if (nsig) {
             memcpy((char *)&shp->st.trapcom[0], savsig, nsig);
+            shp->st.otrapcom = NULL;
```
krader1961: Good catch. But this bug would have been obvious had `shp->st.otrapcom` been used instead of `savsig` and the latter symbol simply removed. There is no reason for `savesig` to even exist. I'll create a PR that does that.

Issue #437, opened 2018-03-18
Job control not enabled on initial shell


perette:
I'm experiencing trouble with job control.  In my initial shell (SHLVL=1) it doesn't work.  It works correctly if I fork a subshell.

I've tracked it down to job.jobcontrol == 0, because in job_init (defined at jobs.c:576):

:622 Tries to sets process group ID, but fails so it sets `possible = 0`
:623 Since setpgid failed with `errno == EPERM`, it enters the if
:625 And skips the while loop because the process is already in the foreground
:640 This `#ifdef` block is skipped (`NTTYDISC` is undefined on my system)
:661 `possible` is still false from earlier so it returns...
...and thus it never gets around to enabling job control, which is down at :692.

Note at :637, then `#endif` comment (`SIGTTIN`) does not match `#ifdef` checked on :621 (`SIGTSTP`).  I'm not sure if this a red herring or a sign of edit fatigue which may have introduced this problem.

It seems to me that the intervening code tries to handle the `!possible` situation, and if it can't, it returns--meaning the extra check for !possible is superfluous.  Thus, I've tried removing line :662 (`if (!possible) return;`), with good results--but considering the obscure stuff going on here, I'm not confident that's free from side effects I don't understand.

Session demonstrating problem:
```
perette@lenora ~ $ sleep 5 &
[1]     4467
perette@lenora ~ $ fg %1
ksh: fg: No job control
perette@lenora ~ $ sleep 5 &
[1]     26065
perette@lenora ~ [subshell 2]$ fg %1
sleep 5
perette@lenora ~ [subshell 2]$
```

siteshwar:
I am running `ksh-20180316+git.660.2b3bffbd-0.fc26.x86_64` on my Fedora box, but could not reproduce this issue. What OS you are using ?

perette:
```
RELEASE=18.3
CODENAME=sylvia
EDITION="Cinnamon 64-bit"
DESCRIPTION="Linux Mint 18.3 Sylvia"
DESKTOP=Gnome
TOOLKIT=GTK
NEW_FEATURES_URL=http://www.linuxmint.com/rel_sylvia_cinnamon_whatsnew.php
RELEASE_NOTES_URL=http://www.linuxmint.com/rel_sylvia_cinnamon.php
USER_GUIDE_URL=help:linuxmint
GRUB_TITLE=Linux Mint 18.3 Cinnamon 64-bit
```

[...]

PR #439, opened 2018-03-19
Fix job control if ksh is started as session leader

siteshwar:
Ksh fails to initalize job control if it is started as login shell. A
login shell is started as a session leader. Ksh tries to verify if job
control is set up properly by calling setpgid() on it's own process.
According to POSIX, calling setpgid() on a session leader should fail
with EPERM. This was causing ksh to skip initialzing job control.

We should be able to initalize job control by safely assuming that ksh
is started in a valid process group.

Resolves: #437

### DIFF
--- a/src/cmd/ksh93/sh/jobs.c
+++ b/src/cmd/ksh93/sh/jobs.c
@@ -143,7 +143,6 @@ static void job_unlink(struct process *);
 static void job_prmsg(Shell_t *, struct process *);
 static struct process *freelist;
 static char beenhere;
-static char possible;
 static struct process dummy;
 static char by_number;
 static Sfio_t *outfile;
@@ -570,7 +569,7 @@ bool job_reap(int sig) {
 }
 
 //
-// Initialize job control if possible. If lflag is set the switching driver
+// Initialize job control. If lflag is set the switching driver
 // message will not print.
 //
 void job_init(Shell_t *shp, int lflag) {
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/jobs.c
```diff
@@ -598,40 +597,17 @@ void job_init(Shell_t *shp, int lflag) {
 #endif /* NTTYDISC */
 
     job.mypgid = getpgrp();
-    // Some systems have job control, but not initialized.
-    if (job.mypgid <= 0) {
```
siteshwar: I am not sure about other systems, but on Linux `getpgrp()` should return 0 only if ksh is started as first process i.e. pid 1. Given that is it safe to remove this code @msekletar ?

siteshwar: Pid 1 has process group 1 on Linux, so my previous comment was incorrect.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -598,40 +597,17 @@ void job_init(Shell_t *shp, int lflag) {
...
-        if ((fd = sh_open(ttynam, O_RDWR)) < 0) return;
-        if (fd != JOBTTY) sh_iorenumber(shp, fd, JOBTTY);
-        job.mypgid = shp->gd->pid;
-#ifdef SIGTSTP
-        tcsetpgrp(JOBTTY, shp->gd->pid);
-        setpgid(0, shp->gd->pid);
-#endif  // SIGTSTP
-    }
 #ifdef SIGTSTP
-    possible = setpgid(0, job.mypgid) >= 0;
```
siteshwar: This call to `setpgid()` succeeds in last stable version of ksh. I am still checking what's causing it to fail now.

siteshwar: `setpgid()` is actually failing on last stable version of ksh. This bug was introduced by reformatting changes in 00d0a53b8.

This line:
```
    if(possible = (setpgid(0,job.mypgid)>=0) || errno==EPERM)     
```
was changed to:

```
   possible = setpgid(0, job.mypgid) >= 0;
    if (possible || errno == EPERM) {
```
So `possible` is not set to 1 if `setpgid()` fails due to `EPERM`. It should have been changed to:

```
   possible = setpgid(0, job.mypgid) >= 0  || errno == EPERM;
    if (possible) {
```

krader1961: Good catch and this is a good example for why the existing style that does assignment inside `if()` conditions is a really bad idea. Also, the `>= 0` test is itself confusing since `setpgid()` by definition returns zero on success, -1 on failure, and should never return a positive value. It would be clearer to write it as

```c
possible = setpgid(0, job.mypgid) != -1 || errno == EPERM;
```

Issue #8, opened 2016-04-24
ksh93: "OLDPWD=/dir cd -" doesn't take you to /dir


stephane-chazelas:
```
$ OLDPWD=/bin ksh93 -c 'OLDPWD=/tmp cd -'
/bin
```

Expected:

```
/tmp
```

Things like `CDPATH=/usr cd bin` are OK though.

(on ksh93u on Debian)

siteshwar:
Still reproducible. I will fix it.

PR #441, opened 2018-03-20
Check if OLDPWD variable is overriden while executing 'cd -'

siteshwar:
Resolves: #8

### DIFF
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -896,9 +896,16 @@ else
     err_exit 'cannot cd to ~{fd} when fd is /dev'
 fi
 
+mkdir $tmp/oldpwd
+OLDPWD=$tmp/oldpwd
+cd -
+[[ $PWD == "$tmp/oldpwd" ]] || err_exit "cd - does not recognize overridden OLDPWD variable"
+
+cd $tmp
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -896,6 +896,8 @@ else
     err_exit 'cannot cd to ~{fd} when fd is /dev'
 fi
 
+mkdir $tmp/oldpwd
+[[ $(OLDPWD="$tmp/oldpwd" cd -) == "$tmp/oldpwd" ]] || err_exit "cd - does not recognize overriden OLDPWD variable"
```
krader1961: Thx for adding a test. However, the wording of the error isn't quite correct. In fact, it's probably worth adding a related test:

```
cd /
OLDPWD=$tmp/oldpwd
cd -
[[ $PWD == "$tmp/oldpwd" ]] || err_exit "cd - does not recognize overridden OLDPWD variable"
```

That's because overriding OLDPWD does work but not if done by creating a new env context. The above test should pass without this fix.

Also, s/overriden/overridden/

krader1961:
LGTM

src/cmd/ksh93/tests/builtins.sh
siteshwar: I have added the test case that you suggested.

Issue #425, opened 2018-02-24
print -s output lost when using -f


qbarnes:
This works as expected with print -s sending the output to the history file:
```
$ print -s "cd -- '$PWD'"
$ history -2
7518    print -s "cd -- '$PWD'"
7519    cd -- '/home/qbarnes'
7520    history -2
```

This works as expected displaying the output to standard out:
```
$ print -f 'cd -- %q\n' "$PWD"   
cd -- /home/qbarnes/github/att/ast
```

However, when the command immediately above is used with -s, the output goes nowhere:
```
$ print -s -f 'cd -- %q\n' "$PWD"
$ history -2
7513    print -f 'cd -- %q\n' "$PWD"
7514    print -s -f 'cd -- %q\n' "$PWD"
7515    history -2
```

I reproduced this problem with the current master branch (77b75138) using x86_64 host running Linux RHEL 7.4 and on my MacOS El Capitan running ksh AJM 93u+ 2012-08-01.  The problem happened also with the oldest Linux ksh binary I still have around (release 2007-11-05).

This seems like an obvious bug to me, but I'm open to the idea that I'm missing something in the documentation where -s and -f are somehow incompatible.  Bug or feature?

perette:
Octal dumping the .sh_history file shows the commands are separated by \n\0 sequences (newline-null), or \n\0\0.  (It seems to use an even number of bytes for each entry.)  `print -s` implicitly inserts the newlines and gets the nulls right, so it works.

`print -s -f` puts in whatever sequence you tell it, but then inserts some cruft where the null should be.  Explicitly including the null in the format string makes it work right:
```
$ print -s -f 'cd -- %q\n\0' "$PWD"
$ history -2
1157	print -s -f 'cd -- %q\n\0' "$PWD"
1158	cd -- /home/perette
1159	history -2
```
Although the number of nulls being written does vary from the usual behavior, it seems to work fine.

Perhaps something is running off the end of a buffer, expecting the null to be there, instead encountering whatever cruft is in memory there?

perette:
```
$ print -s -f 'cd -- %q\n' "$PWD"
$ history -2
730	print -s -f 'cd -- %q\n' "$PWD"
731	cd -- /home/perette/Projects/ast/build/src/cmd/ksh93
732	history -2
```

Patch is attached.

[ksh-printf-s-f.patch.txt](https://github.com/att/ast/files/1827795/ksh-printf-s-f.patch.txt)

siteshwar:
[...]

PR #442, opened 2018-03-20
Fix syncing history when print -s -f is used

siteshwar:
Resolves: #425


krader1961:
LGTM only because the change doesn't look like it can make matters worse and appears to fix the bug. I freely admit I don't understand what the code is doing or, for that matter, why `print -s` even exists.

PR #444, opened 2018-03-22
Remove support for vfork().

### LINKED ISSUE #443: `vfork()` doesn't share memory on some systems
User @kernigh brought to our attention that `vfork()` on OpenBSD 6.2 does not share the address space between the parent and child process. The old build system has a test that `vfork()` implements the legacy behavior. When we converted to the Meson build system and greatly simplified the feature detection code we didn't port those tests and made the incorrect assumption that __surely__ there aren't any `vfork()` implementations we care about which don't share the address space.

User @kernigh originally proposed reinstating the necessary feature test in PR #432. But after some discussion the consensus appears to be that maintaining support for `vfork()` isn't worth the complexity. This issue is to track removal of all support for `vfork()`.

krader1961:
This reduces the number of test failures from 54 to 31 on OpenBSD 6.2.
That's still double the number I'm seeing on macOS but a big step in the
right direction.

Fixes #443


siteshwar:
On opensuse `signal` tests are consistently failing with this change.

kernigh:
I see [signal+opensuse failing in Travis](https://travis-ci.org/att/ast/jobs/356719385), but I don't see the testlog because

```
Full log written to /source/build/meson-logs/testlog.txt
cat: meson-logs/testlog-malloc.txt: No such file or directory
```

These are the signal failures in OpenBSD 6.2:

```
119/153 signal                                  FAIL    89.46 s

--- command ---
SHELL='/home/kernigh/park/ast/build/src/cmd/ksh93/ksh' LANG='' LD_LIBRARY_PATH='
/home/kernigh/park/ast/build/src/lib/libast:/home/kernigh/park/ast/build/src/lib
/libcmd:/home/kernigh/park/ast/build/src/lib/libcoshell' /home/kernigh/park/ast/
build/src/cmd/ksh93/ksh /home/kernigh/park/ast/src/cmd/ksh93/tests/signal.sh sig
nal
--- stderr ---
        signal.sh[289]: interrupt --d test failed -- expected '3-intr', got '3-0
000 2-0000 1-0000'
        signal.sh[289]: interrupt -td test failed -- expected '3-intr 2-intr 1-i
ntr 1-0258', got '3-0000 2-0000 1-intr 1-0000'
        signal.sh[289]: interrupt x-d test failed -- expected '3-intr 2-intr 1-0
000', got '3-0000 2-0000 1-0000'
        signal.sh[289]: interrupt xtd test failed -- expected '3-intr 2-intr 1-i
ntr 1-0000', got '3-0000 2-0000 1-intr 1-0000'
        signal.sh[558]: ${c.car[0].value.q} is  0 but should be 4
        signal.sh[559]: ${c.car[1].value.q} is  0 but should be 5
-------
```

The failures for signal/C.UTF-8 and signal/shcomp look identical. All 4 failures (--d, --td, x-d, xtd) use option 'd'. This 'd' changes how tst-2 and tst-3 are run:

```
case $1 in
*d*)    tst-2 $1 $2; status=$? ;;
*)    $2 -c "tst-2 $1 $2"; status=$? ;;
```

With 'd', it runs tst-2 or tst-3 as direct commands. Without 'd', it runs `ksh -c tst-2` or tst-3.

The shell can use vfork() or fork() to run direct commands. First, the vfork() or fork() child tries to execve() tst-[23]. This fails with ENOEXEC because tst-[23] is a shell script without #! line. If the child came from vfork(), then the child exits, and the shell does a regular fork() so the subshell can run tst-[23]. If the child came from fork(), then the child just runs tst-[23].

See vexexec() in src/cmd/ksh93/sh/path.c; the flag `SPAWN_FORK` means that the child came from fork(), not vfork(). I suspect, but don't know, that removing vfork() reveals a bug in the SPAWN_FORK code in vexexec(). This bug would hide if ksh uses posix_spawn(); it doesn't use posix_spawn() in OpenBSD.

krader1961:
@kernigh, Yes, your analysis appears to be correct.

TL;DR: The original code only works if `vfork()` can be used when `posix_spawn()` isn't available.

The long version:

If you checkout commit 56e70473 (the one prior to ed6ba1d that made fork/vfork use unconditional) you will see that it adds `-D_lib_posix_spawn=1` to the compile flags on OpenSuse and Ubuntu. Even though the feature test reports

```
Checking if "posix_spawn exists and it works and its worth using " runs: NO (1)
```

Both `_lib_vfork` and `_real_vfork` are set to true by a previous change that hardcodes them to one. The current master branch sets

```
#define _lib_posix_spawn 0
```

Which means `posix_spawn()` isn't used. It also does not define `_lib_fork` or `_lib_vfork`. Which is okay because commit ed6ba1d means we always assume `vfork()` works and can be used.

If I modify the Meson build config at commit 56e70473 to not define `_lib_posix_spawn` and set `_lib_vfork` to zero then the test fails exactly as it does with this change. In other words the code was broken before this change and before @siteshwar's change to use fork and vfork unconditionally. It only works if `vfork()` can be used when `posix_spawn()` isn't available.

DavidMorano:
Thank you for removing VFORK.
The short of it is that VFORK was being called on some OSes and it was failing because the semantics for each particular VFORK was not being properly honored by the AST code. Even more short: in general VFORK is too dangerous to try to use it properly across each OS.  Good riddance to VFORK! It is one less thing that has to be stubbed out in order to get a properly working shell! Good riddance!

kernigh:
I was wrong about the signal failures; see #449.

Issue #10, opened 2016-04-26
Wrong syntax for the "suspend" alias in ksh93


stephane-chazelas:
In ksh93 (and mksh and ksh88), suspend is defined as:

```
suspend='kill -s STOP $$'
```

Unquoted parameter expansion means to ask the shell to perform word splitting and globbing on it.

```
$ echo "$$"
24725
$ IFS=7
$ suspend
kill: 24: permission denied
kill: 25: permission denied
```

It should be:

```
alias suspend='kill -s STOP "$$"'
```

dannyweldon:
I just tried this with 2012-08-01 on ubuntu and it worked fine, and the same with a local build of 2014-12-24 (beta).  What ksh version and platform are you using?

stephane-chazelas:
Danny,

```
$ ksh -c 'echo "${.sh.version}"; type suspend'
Version ABIJM 93v- 2014-12-24
suspend is an alias for 'kill -s STOP $$'
```

Leaving a variable unquoted in list contexts has a very special meaning in Bourne-like shell other than `zsh`. It doesn't make sense to leave a variable unquoted unless you want splitting and/or globbing applied to it. See https://unix.stackexchange.com/questions/171346/security-implications-of-forgetting-to-quote-a-variable-in-bash-posix-shells for more details.

Above, because `$$` is not quoted, splitting (globbing can't apply as pids never contain wildcard characters) applies, so if `$IFS` contains digits, `$$` will be split.

Yes, that's an awful, very surprising design. It's there because David Korn didn't break backward compatibility with the Bourne shell. And I suppose the Bourne shell did that, so parameter expansion was not too different from that of the Thompson shell.

```
$ ksh -c 'echo "${.sh.version}"; IFS=1234567890; suspend'
Version ABIJM 93v- 2014-12-24
./arch/linux.i386-64/bin/ksh: kill: : Arguments must be %job, process ids, or job pool names
```

dannyweldon:
Sorry, I realise now that one must set IFS to a number that occurs within $$ for this to take effect, so I have been able to replicate it.  I doubt if anyone would change their IFS variable in an interactive shell to anything that would cause this, but I think it can't hurt to update this alias for correctness.

krader1961:
[...]

PR #445, opened 2018-03-23
Avoid word splitting process id in suspend alias

siteshwar:
Fix definition of suspend alias to avoid word splitting.
Also fix related man page entry.

Resolves: #10


krader1961:
LGTM although I wouldn't have bothered. Anyone putting a digit like `7` in `$IFS` should not be surprised if it causes their computer to halt and catch fire. The design of the IFS mechanism in POSIX shells makes it a practical impossibility to protect all statements that might be affected by a user doing something like that. This change only protects one of an infinite number of such land mines.

Issue #23, opened 2016-09-19
Regression appending to an indexed array overwrites arr[-1] if arr[0] is unset


ormaaj:
This appears to be a regression since the last release.

``` shell
 $ ksh /dev/fd/9 9<<\EOF
set -x
typeset -a a=(w x) b=(a b c)
a+=("${b[@]}")           # Correct behavior with a[0] set
typeset -p a
typeset -a a=([1]=w [2]=x)
a+=("${b[@]}")           # Incorrectly overwrites a[-1] when a[0] is unset
typeset -p a
a[${#__[@]}+1].__+=(y z) # Hack to get a reference to the correct element.
typeset -p a .sh.version
EOF

+ a=( w x )
+ b=( a b c )
+ typeset -a a b
+ a+=( a b c )
+ typeset -p a
typeset -a a=(w x a b c)
+ a[1]=w
+ a[2]=x
+ typeset -a a
+ a+=( a b c )
+ typeset -p a
typeset -a a=([1]=w [2]=a [3]=b [4]=c)
+ a+=( y z )
+ typeset -p a .sh.version
typeset -a a=([1]=w [2]=a [3]=b [4]=c [5]=y [6]=z)
.sh.version='Version ABIJM 93v- 2014-12-24'
```

PR #447, opened 2018-03-26
Fix a regression in appending to arrays

siteshwar:
Existing elements are getting overwritten while appending to arrays if an array contains empty indexes. For e.g.
```
typeset -a a=([1]=w [2]=x)
typeset -a b=(a b c)
a+=("${b[@]}")
typeset -p a
```

Actual output:
```    
typeset -a a=([1]=w [2]=a [3]=b [4]=c)
```    
Expected output:
```    
typeset -a a=([1]=w [2]=x [3]=a [4]=b [5]=c)
```
    
Current code was calculating index of the last element by counting number of elements in the array. It will not work if an array contains empty indexes. This commit fixes it.
    
Resolves: #23


krader1961:
LGTM

Issue #426, opened 2018-02-27
'command -p' broken in certain contexts


McDutchie:
`command -p` is supposed to run a command with the default system `PATH` as opposed to the current `PATH`, so it is guaranteed to get a POSIX standard utility. But:

`$ ksh -c 'command -p ls'`
`ksh: ls: not found`
`$ ksh -c '(command -p ls)'`
`ksh: ls: not found`
`$ ksh -c 'echo $(command -p ls)'`
(file listing is output)
`$ ksh -c 'command -p ls & wait'`
(file listing is output)

Bug confirmed on:
* Version JM 93t+ 2010-03-05
* Version AJM 93u+ 2012-08-01
* current git

Bug *not* present on:
* Version M 1993-12-28 r

(previously reported [on the ast-developers list](http://narkive.com/rqDYjbtM.1))

krader1961:
Any idea what this vague sentence in the documentation means?

> The -p option causes a default path to be searched rather than the one defined by the value of PATH.

What is the "default path"? Using the `-p` flag causes option `SH_DEFPATH` to be turned on. Which in turn appears to cause `shp->defpathlist` to be used instead of `shp->pathlist`.  It looks like this involves iterating over `$PATH` and passing each element to `path_chkpaths()` to decide if it is allowed in the "default path".

Instrumenting `b_command()` shows that in an interactive shell (at least on my macOS system) `shp->defpathlist` is a NULL pointer and `shp->pathlist` is a linked list that mirrors my `$PATH`. When running a non-interactive shell both members are NULL pointers whether or not the `-p` flag is used. But without that flag it seems to fallback to the value of `$PATH` deeper in the code.

Also, note that it works in an interactive shell or if you force interactive mode; e.g., `ksh -ic 'command -p ls'` works. So it seems likely this has something to do with the difference in how interactive versus non-interactive shells are initialized.

McDutchie:
> Any idea what this vague sentence in the documentation means?
>> The -p option causes a default path to be searched rather than the one defined by the value of PATH.

[POSIX spec for the 'command' utility](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html#tag_20_22): "**-p** –  Perform the command search using a default value for PATH that is guaranteed to find all of the standard utilities."

Effectively this means that, instead of `$PATH`, the path should be searched that is output by the command `getconf PATH` on the current system, or by the `confstr`(3) function (name `_CS_PATH`).

siteshwar:
[...]

PR #448, opened 2018-03-28
Fix initialization of default PATH variable

siteshwar:
Default PATH variable is getting used before it is initialized when ksh is started as a non-interactive shell. This patch fixes it.
    
Resolves: #426


src/cmd/ksh93/sh/path.c
```diff
@@ -47,10 +47,12 @@ static void funload(Shell_t *, int, const char *);
 static void exscript(Shell_t *, char *, char *[], char *const *);
 static bool path_chkpaths(Shell_t *, Pathcomp_t *, Pathcomp_t *, Pathcomp_t *, int);
 static void path_checkdup(Shell_t *shp, Pathcomp_t *);
+static Pathcomp_t *defpath_init(Shell_t *shp);
 
 static const char *std_path;
```
krader1961: Mind adding a `= NULL` here? I know it's not strictly speaking necessary since static vars are guaranteed to be initialized to zero but it's better to be explicit.

krader1961:
LGTM

PR #449, opened 2018-03-29
Fix a use after free bug

siteshwar:
There is a use after free bug which is causing signal tests to fail
randomly. This bug was introduced after the last stable release.
Since we don't have detailed history of why these changes were made,
this commit is more of a workaround to fix such random crashes.

Related: #398


krader1961:
Also, when running the tests via `meson test --setup=malloc` there doesn't appear to be any reduction in failures as a result of this PR. So I'm curious what leads to you believe, even on an empirical rather than theoretical basis, that this PR is correct.

krader1961:
Okay, I'd be happier if this had a noticeable effect on the unit tests when run with freed buffer poisoning but it looks okay so go ahead and merge it.

PR #451, opened 2018-03-30
Remove wrong call to setpgid(pid, 0)

kernigh:
When spawnvex() calls fork(), it never sets vex->pgrp, so vex->pgrp is
always 0. It then called setpgid(pid, 0) on every fork() child. This
tried to put every child into a new process group, whether or not we
had a SPAWN_pgrp command.

In OpenBSD 6.2, if the child's execve() succeeded, then setpgid()
failed with EACCES. If the child's execve() failed, then setpgid()
succeeded, and the child possibly executed a shell script in the wrong
process group. This caused 4 failures in src/tests/signal.sh because
the `kill -s INT 0` failed to signal some shell scripts in the wrong
process group. The fork() child can execute a shell script since the
removal of vfork() in PR #444 to fix bug #443.

src/tests/signal.sh still has 2 other failures in OpenBSD.

### DIFF
--- a/src/lib/libast/misc/spawnvex.c
+++ b/src/lib/libast/misc/spawnvex.c
@@ -821,11 +821,7 @@ bad:
             close(msg[0]);
         }
         if (!(flags & SPAWN_FOREGROUND)) sigcritical(0);
-        if (pid != -1 && vex) {
-            if (vex->pgrp >= 0 && setpgid(pid, vex->pgrp) < 0 && vex->pgrp && errno == EPERM)
-                setpgid(pid, pid);
-            VEXINIT(vex);
-        }
+        if (pid != -1 && vex) VEXINIT(vex);
         errno = n;
         return pid;
     }


kernigh:
Travis tested this commit (2211023) twice. First, [the branch passed Travis](https://travis-ci.org/kernigh/ast/builds/360315784), but now [the pull request fails Travis](https://travis-ci.org/att/ast/builds/360318735?utm_source=github_status&utm_medium=notification). This might be the random failure that #449 might fix, so I'm going to wait for #449, then try again.

siteshwar:
I restarted the Travis build and it works now.

krader1961:
The only signal test failures I'm seeing on OpenBSD 6.2 are these two:

```
        signal.sh[558]: ${c.car[0].value.q} is  0 but should be 4
        signal.sh[559]: ${c.car[1].value.q} is  0 but should be 5
```

Not saying your analysis is wrong but since I'm not seeing any change in behavior with this PR.

kernigh:
With current master (836d0ff) and OpenBSD 6.2, I run `meson test signal` and see 6 failures:

```
        signal.sh[289]: interrupt --d test failed -- expected '3-intr', got '3-0000 2-0000 1-0000'
        signal.sh[289]: interrupt -td test failed -- expected '3-intr 2-intr 1-intr 1-0258', got '3-0000 2-0000 1-intr 1-0000'
        signal.sh[289]: interrupt x-d test failed -- expected '3-intr 2-intr 1-0000', got '3-0000 2-0000 1-0000'
        signal.sh[289]: interrupt xtd test failed -- expected '3-intr 2-intr 1-intr 1-0000', got '3-0000 2-0000 1-intr 1-0000'
        signal.sh[558]: ${c.car[0].value.q} is  0 but should be 4
        signal.sh[559]: ${c.car[1].value.q} is  0 but should be 5
```

The first 4 failures disappear with this pull request. The first 4 failures also disappeared when I accidentally tested a ksh from before the vfork() removal.

This small example shows the wrong setpgid() call:

```
$ cat <<'EOF' >child.sh
> sleep 1
> echo child = $$
> exec perl -e 'print "child pgrp = ", getpgrp, "\n"'
> EOF
$ chmod +x child.sh
$ ksh=/home/kernigh/park/ast/build/src/cmd/ksh93/ksh
$ $ksh -c './child.sh; echo parent = $$'
child = 50526
child pgrp = 50526
parent = 79038
```

This $ksh is current master (836d0ff). The parent 79038 forked a child 50526, then the parent's setpgid(50526, 0) put the child in a new process group 50526. (For machines without perl, `python3 -c 'import os; print("child pgrp =", os.getpgrp())'` might work.)

Now I check out this pull request (2211023) and do the usual `ninja -C build` to rebuild $ksh. Now the setpgid() doesn't happen:

```
$ $ksh -c './child.sh; echo parent = $$'
child = 11090
child pgrp = 52582
parent = 52582
```

The parent pid 52582 didn't call setpgid(11090, 0), so the child 11090 is still in the parent's process group 52582. (The parent has its own process group because I entered the $ksh command in an interactive shell with job control.) At least bash, csh and zsh also leave the child in the parent's process group.

krader1961:
Okay, I'm now seeing the four interrupt failures on OpenBSD 6.2. Not sure why I wasn't the other day. And this change does make those errors go away. I'm still slightly uncomfortable with the change since I don't really understand what was the intent of the block of code being removed. Nonetheless, merging since this doesn't seem to break anything and does fix one failure mode on OpenBSD.

PR #452, opened 2018-03-31
Signal tests should not be run in parallel

siteshwar:
Running signal tests in parallel causes failures. For e.g. on my machine
running this command:

meson test --repeat=10 signal

causes most of the tests to fail.

We should avoid running these tests in parallel.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -94,7 +94,7 @@ all_tests = [
     ['io'], ['locale'], ['math', 50], ['nameref'], ['namespace'],
     ['options'], ['path'], ['pointtype'], ['pty'], ['quoting'], ['quoting2'],
     ['readcsv'], ['recttype'], ['restricted'], ['return'], ['select'],
-    ['sh_match'], ['sigchld', 100], ['signal', 100], ['statics'], ['subshell', 100],
+    ['sh_match'], ['sigchld', 100], ['signal', 100, false], ['statics'], ['subshell', 100],
     ['substring'], ['tilde'], ['timetype'], ['treemove'], ['types'],
     ['variables'], ['vartree1'], ['vartree2'], ['wchar']
 ]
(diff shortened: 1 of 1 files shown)


krader1961:
I don't think this should be merged for several reasons.

First, on general principals alone only unit tests that inherently cannot be run in parallel because they depend on access to something like a device should be excluded from being run in parallel. That is not true for the `signal` unit test. In fact, that unit test needs to be stress tested by being run in parallel to help expose race conditions or other problems with the test such as using file names that are not unique to a specific test invocation.

Second, only OpenSuse seems to have a problem running two or more instances of the `signal` test in parallel.

Third, even when running only one instance of the `signal` test at any given point in time I've seen the same failure; albeit very infrequently.

Fourth, this unit test is masking at least one significant bug (albeit unrelated to the purpose of this PR). In the construction of `tst-2` if you comment out the redirection of stderr as I've done here:

```
cat > tst-1 <<'!'
#exec 2>/dev/null
case $1 in
```

You will see errors like this written to stderr:

```
tst-1[24]: ?^P5?3?^K?_%: not found [No such file or directory]
```

The specific string changes from run to run. Which is a strong indication that an invalid pointer is being used. Which is likely a use after free bug. In fact, it appears the person who wrote that specific portion of the signal unit test recognized there was a bug by explicitly adding the stderr redirection to /dev/null and testing that the test fails with status 258 which is the magic status value for a command not found. While it is valid that the command isn't found in some situations given that the script doesn't contain a valid `#!` header the weird command name string is not valid.

krader1961:
Note that you can make the signal test fail close to 100% of the time when running a single unit test (no parallelism) via `meson test signal` simply by running enough CPU hogs (i.e., spinning processes) to keep all CPUs busy. Although in my testing you generally have to oversubscribe the CPUs by 150% to get consistent failures from this simple strategy.

The main failure mode is the watchdog process put into the background to keep the realtime signal tests from blocking forward progress:

```
( ( sleep 5; kill $$ 2> /dev/null) & ) &
```

It is a) not killed when those tests are completed which can result in subsequent tests being interrupted and b) if the system if overloaded will kill the shell even though it is otherwise making forward progress with the realtime signal tests. This is a classic case of someone incorrectly assuming that the system they are using to run these tests is typical.

I've got a small change to the *signal.sh* script which greatly improves its robustness. I'll create a PR for that change. Having said that I think we should probably just simplify the unit test by removing that watchdog process since it should be the responsibility of the process managing the unit tests to handle timeouts.

bitstreamout:
On Sun, Apr 01, 2018 at 04:52:17AM +0000, Kurtis Rader wrote:
> I don't think this should be merged for several reasons.
> 
> First, on general principals alone only unit tests that inherently cannot be run in parallel because they depend on access to something like a device should be excluded from being run in parallel. That is not true for the `signal` unit test. In fact, that unit test needs to be stress tested by being run in parallel to help expose race conditions or other problems with the test such as using file names that are not unique to a specific test invocation.
> 
> Second, only OpenSuse seems to have a problem running two or more instances of the `signal` test in parallel.
> 

The problem with openSUSE is more that the signal handler with latest ksh is missused
that is malloc() is called within signal handler.

* Fri Jun 12 2015 mls@suse.de
  - go back to ksh93u to have a stable version in SLE-12
    use 93vu as version so that the software stack sees this
    as an update. (FATE#319107, bsc#939252)
* Thu Jul 17 2014 werner@suse.de
  - Skip signal test as currently ksh93 uses malloc() within signal
    handlers and this cause deadlocks even with libast memory
    management

-- 
  "Having a smoking section in a restaurant is like having
          a peeing section in a swimming pool." -- Edward Burr

krader1961:
> The problem with openSUSE is more that the signal handler with latest ksh is missused
that is malloc() is called within signal handler.

Yes, that is definitely dangerous and needs to be fixed. Whether or not that accounts for the weird errors I noted in my previous comment is unknown but seems unlikely.

PR #453, opened 2018-04-02
Improve robustness of ksh signal test

krader1961:
The signal test as currently written is very sensitive to the load on
the system. This is because it uses a background watchdog process which
waits just five seconds before killing the parent. And the watchdog
process isn't killed when it is no longer needed. This can result in
two failure modes. So explicitly track the watchdog process and kill it
when it is no longer needed to keep it from interfering with the parent
process.


siteshwar:
LGTM

Issue #75, opened 2017-10-06
ksh93: "<>" redir operator defaults to stdout, should default to stdin


McDutchie:
The I/O redirection operator `<>` (open a file descriptor for both read and write) defaults to opening standard output (i.e. is short for `1<>`) instead of defaulting to opening standard input (`0<>`) as [POSIX specifies](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_07_07).

	case $(echo hi <>/dev/null) in
	( '' )	echo 'bug' >&2 ;;
	( hi )	echo 'ok' >&2 ;;
	esac >/dev/null

ksh93 is the only shell that outputs "bug". The behaviour is documented in the man page but in terms of POSIX compliance it's nonetheless a bug. It means cross-platform scripts always need to specify the file descriptor with this redirector, even though this is supposed to be optional.

(Stéphane Chazelas flagged this up on austin-group-l.)

PR #454, opened 2018-04-03
<> operator should redirect stdin by default

siteshwar:
Ksh redirects stdout when no fd is specified with <> operator. According to POSIX <> operator should redirect stdin by default.
   
Resolves: #75


krader1961:
LGTM. Good catch.

krader1961:
Don't forget to add an entry to *CHANGELOG.md* and *src/cmd/ksh93/ksh.1*.

PR #455, opened 2018-04-04
Turn lcgen.c into a feature test

### LINKED ISSUE #419: File *port/lcgen.c* needs to be moved to the *features* directory
While working on the use-after-free bug, #398, I noticed that my build was failing because building the libast library included the object file produced by building *port/lcgen.c* and, due to debugging statements I added, that resulted in a symbol being defined twice. I fixed that specific issue with commit b16a04f. This issue is to remind us to move that program to the recently introduced, top-level, *features* directory since the object file produced by that source module isn't meant to be included in any library or binary outside of running it to detect features provided by the OS.

krader1961:
This moves the lcgen dynamically generated artifacts out of the source
tree.

Resolves #419


krader1961:
FYI, The only reason I modified the lcgen.c source was to eliminate warnings from the compiler. Thus making it easier to notice when there is a real problem.

features/locales/lcgen.c
```diff
@@ -212,23 +212,25 @@ int main(int argc, char **argv) {
...
-        fprintf(stderr, "%s: { hdr lib tab } arguments expected\n", command);
+    command = argv[0];
+    const char *tab = _LC_TAB;
+    const char *hdr = _LC_HDR;
+    const char *lib = _LC_LIB;
+    FILE *tf;
+    FILE *hf;
+    FILE *lf;
+
+    if (!(tf = fopen(tab, "r"))) {
```
siteshwar: Assignment should be moved out of `if` condition.

features/locales/lcgen.c
```diff
@@ -212,23 +212,25 @@ int main(int argc, char **argv) {
...
+        fprintf(stderr, "%s: %s: cannot read\n", command, tab);
         return 1;
     }
     if (!(hf = fopen(hdr, "w"))) {
```
siteshwar: Assignment should be moved out of if condition.

siteshwar:
I left some comments. Rest LGTM.

PR #458, opened 2018-04-05
Check for arrays before removing discipline

siteshwar:
Ksh should check for arrays before removing discipline.


src/cmd/ksh93/sh/array.c
```diff
@@ -132,17 +132,22 @@ static int arsize(struct index_array *ap, int maxi) {
 
 static struct index_array *array_grow(Namval_t *, struct index_array *, int);
 
-// Return index of highest element of an array.
+// Return next index after the highest element in an array.
 int array_maxindex(Namval_t *np) {
     struct index_array *ap = (struct index_array *)nv_arrayptr(np);
     int i = ap->maxi;
     if (is_associative(ap)) return (-1);
```
krader1961: Mind removing the pointless parens around the return value since you're in the neighborhood? :smile:

krader1961:
LGTM

Issue #74, opened 2017-10-06
ksh93: [/test give status 1 instead of 2 on invalid operator argument


McDutchie:
[ 123 -eq 1XX ]
    echo $?

Output:

    test.sh[1]: [: 1XX: arithmetic syntax error
    1

POSIX [specifies](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html#tag_20_128_14) that `$?` should be at least 2.

PR #463, opened 2018-04-09
Fix exit status of test builtin for errors

siteshwar:
Resolves: #74

### DIFF
--- a/src/cmd/ksh93/tests/bracket.sh
+++ b/src/cmd/ksh93/tests/bracket.sh
@@ -471,4 +471,8 @@ x=10
 
 ([[ x -eq 10 ]]) 2> /dev/null || err_exit 'x -eq 10 fails in [[...]] with x=10'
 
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/bracket.sh
```diff
@@ -471,4 +471,7 @@ x=10
 
 ([[ x -eq 10 ]]) 2> /dev/null || err_exit 'x -eq 10 fails in [[...]] with x=10'
 
+# POSIX specifies that on error, test builtin should always return value > 1
+test 123 -eq 123x 2>/dev/null || [[ $? -ge 2 ]] || err_exit 'test builtin should return value greater than 1 on error'
```
krader1961: I'm not a fan of chaining `||` and `&&`. It's also not needed in this case and just makes it harder to understand the logic. Just

```
test  ...
[[ $? -ge 2 ]] || err_exit
```

Issue #460, opened 2018-04-06
GNU `dladdr()` doesn't work for addresses in the base text segment


krader1961:
It's been bugging me that on GNU Linux my `dump_backtrace()` function was only showing function names for functions in dynamically loaded libraries. So I finally spent a minute carefully reading the `dladdr()` man page and was surprised to find that it is explicit that only symbols in shared libraries will be returned. First thing to try is to see if the `dladdr1()` function behaves more sensibly (i.e., like `dladdr()` on macOS). If not I may have to resort to libbfd (ugh!) on GNU.

krader1961:
Sadly, `dladdr1()` also doesn't handle addresses in the application as opposed to shared libraries. So that's not available as an easy solution.

krader1961:
Thanks, @kernigh. Adding `-Wl,-E` does indeed help on non-MacOS platforms. However, on MacOS it causes the final link to fail with `ld: unknown option: -E`. Also, on the other platforms `static` functions still cannot be resolved. Whereas on MacOS (without `-Wl,-E`) static and global functions can be resolved by `dladdr()`. I'm no Apple MacOS "fanboy" but this brain dead behavior by other implementations of `dladdr()` is damn annoying.

kernigh:
You are correct. I only need to change handle_sigsegv() to static, then the backtrace (in OpenBSD) names the wrong function.

I blame the ELF format (in Linux and most BSDs). `readelf -s src/cmd/ksh93/ksh` shows two symbol tables, `.dynsym` and `.symtab`. The local symbols (for static functions) are only in .symtab, but the "Section to Segment mapping" in `readelf -l src/cmd/ksh93/ksh` only has .dynsym. When the dynamic linker loads the segments, it sees .dynsym but not .symtab, so dladdr() can't find static functions.

To get the correct function, one would need to open the ELF file and load the .symtab that the dynamic linker didn't load. In my system, `backtrace_symbols()` already does the work. If I disable the dladdr() call...

```
diff --git a/src/cmd/ksh93/sh/fault.c b/src/cmd/ksh93/sh/fault.c
index 8ff1ea30..0c06bb68 100644
--- a/src/cmd/ksh93/sh/fault.c
+++ b/src/cmd/ksh93/sh/fault.c
@@ -74,7 +74,7 @@ void dump_backtrace(int max_frames, int skip_levels) {
     for (int i = skip_levels; i < max_frames; i++) {
         int n;
         Dl_info info;
[...]

PR #464, opened 2018-04-09
Find libexecinfo in OpenBSD

kernigh:
OpenBSD installs libexecinfo in /usr/local, where the compiler does
not search by default. Add -I/usr/local/include to feature tests so
execinfo.h gets found. Look for all libraries (not only libiconv) in
/usr/local so libexecinfo gets found. Do this only for OpenBSD.

Issue #460 is still a problem on OpenBSD; ksh can now use libexecinfo
to print a backtrace, but can't find the right symbols.

### DIFF
--- a/features/meson.build
+++ b/features/meson.build
@@ -17,28 +17,36 @@ else
     feature_data.set('_more_long_int', 0)
 endif
 
+if system == 'openbsd'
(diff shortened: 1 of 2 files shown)


krader1961:
LGTM. Merged. Thx.

PR #465, opened 2018-04-09
Don't run wrong file with same name in PATH

### LINKED ISSUE #429: The ksh `functions` unit test runs an external `zzz` command if found in `$PATH` rather than the function of the same name
A test puts my OpenBSD machine to sleep. If I press the power button to wake my machine, it sleeps again. The test acts as if it is running `/usr/sbin/zzz` in a loop. There are several instances of `zzz` in the tests, so a test might be failing and running `/usr/sbin/zzz` by mistake.

1. My machine runs OpenBSD/amd64 6.2. I had enabled the daemon `apmd -A`. The daemon allows me to run `zzz` to sleep.
2. My login shell is OpenBSD's pdksh in `/bin/ksh`.
Its KSH_VERSION is `@(#)PD KSH v5.2.14 99/07/13.2`.
3. I had installed OpenBSD's ksh93 package. It's at `/usr/local/bin/ksh93`.
Its KSH_VERSION is `Version AJM 93u+ 2012-08-01`.
4. I cloned git master at commit 77b7513.
5. I ran `meson build --prefix=$HOME/prefix` and `ninja -C build`.
6. I ran `ninja -C build install`, but then in ~/prefix/bin, I renamed ksh to ksh93. I also renamed the manual page. My path has ~/prefix/bin before /usr/local/bin, so `whence ksh93` says `/home/kernigh/prefix/bin/ksh93`, and `whence ksh` says `/bin/ksh`.
7. I cd to ast/build and run `meson test`. The shell where I type `meson test` is OpenBSD's pdksh. I have `SHELL=/bin/ksh` in the environment.
8. I wait for tests. The "basic" tests take more than 60 seconds each.
9. The computer goes to sleep.
10. If I wake the computer, it goes to sleep again.
11. I switch off the power to the computer.

kernigh:
Tests would run the wrong `dotscript`, `tst`, `tst-1`, `tst-2`,
`tst-3`, or `zzz` if PATH had another script or command with the same
name. For example, functions.sh would run /usr/sbin/zzz (which puts
some BSD machines to sleep) instead of the test's own `zzz`.

Move the directory with the test's `dotscript` or `tst*` to the front
of PATH.  Remove most directories from PATH when searching for `zzz`,
because the test searches PATH before finding its `zzz` in FPATH.

I checked this change by putting scripts or commands with these names
in my own PATH before running `meson test`.

Resolves #429

### DIFF
--- a/src/cmd/ksh93/tests/functions.sh
+++ b/src/cmd/ksh93/tests/functions.sh
@@ -268,9 +268,11 @@ chmod +x $tmp/foobar
 FPATH=$tmp
 cd $tmp
 print 'function zzz { return 0;}' > zzz
-PATH=$PATH:
+# Remove other commands named zzz from PATH.
+old_path=$PATH
+PATH=.
 zzz  2> /dev/null || err_exit 'function not found in . when $PWD is in FPATH'
-PATH=${PATH%:}
+PATH=$old_path
 cd ~-
 autoload foobar
 if [[ $(foobar 2>/dev/null) != foo ]]
@@ -525,7 +527,7 @@ then
 fi
 
 print 'set a b c' > dotscript
-[[ $(PATH=$PATH: $SHELL -c '. dotscript;print $#') == 3 ]] || err_exit 'positional parameters not preserved with . script without arguments'
+[[ $(PATH=:$PATH $SHELL -c '. dotscript;print $#') == 3 ]] || err_exit 'positional parameters not preserved with . script without arguments'
 cd ~- || err_exit "cd back failed"
 function errcheck
 {
--- a/src/cmd/ksh93/tests/signal.sh
+++ b/src/cmd/ksh93/tests/signal.sh
@@ -266,7 +266,8 @@ chmod +x tst tst-?
 
 # end standalone test generation
 
-export PATH=$PATH:
+# Find our tst* scripts before any commands with the same name.
+export PATH=:$PATH
 typeset -A expected
 expected[---]="3-intr"
 expected[--d]="3-intr"


kernigh:
[An earlier run of Travis](https://travis-ci.org/kernigh/ast/jobs/364353951) got this failure:

```
 39/153 coprocess/shcomp                        FAIL    27.31 s
--- command ---
SHELL='/source/build/src/cmd/ksh93/ksh' LANG='' SHCOMP='/source/build/src/cmd/ksh93/shcomp' LD_LIBRARY_PATH='/source/build/src/lib/libast:/source/build/src/lib/libcmd:/source/build/src/lib/libcoshell' /source/build/src/cmd/ksh93/ksh /source/src/cmd/ksh93/tests/shcomptest /source/src/cmd/ksh93/tests/coprocess.sh coprocess
--- stderr ---
	ksh.coprocess.yPKLYnrb[259]: cat coprocess 2 hung
-------
```

This was my commit d4a99e3 in Fedora. When Travis redid the test for this pull request, it passed. My commit only edits two other tests (function.sh and signal.sh) and shouldn't change the result of coprocess.sh. I don't know if this "cat coprocess 2 hung" is a real bug in ksh or a mistake by Travis. Either way, I can't reproduce the failure.

siteshwar:
`coprocess` tests are fragile and known to break at times.

krader1961:
Yeah, the `cat process 2 hung` failure is quite common. Someone should look into why that test is flakey :smile:

src/cmd/ksh93/tests/functions.sh
```diff
@@ -268,9 +268,11 @@ chmod +x $tmp/foobar
 FPATH=$tmp
 cd $tmp
 print 'function zzz { return 0;}' > zzz
-PATH=$PATH:
+# Remove other commands named zzz from PATH.
+old_path=$PATH
```
siteshwar: I think it will be better to just change name of the function then saving and restoring old path.

kernigh: I had posted in #429 a diff that renames `zzz` to `zzf`. That doesn't fix the problem, because there might be other commands named `zzf` in PATH.

src/cmd/ksh93/tests/functions.sh
```diff
@@ -525,7 +527,7 @@ then
 fi
 
 print 'set a b c' > dotscript
-[[ $(PATH=$PATH: $SHELL -c '. dotscript;print $#') == 3 ]] || err_exit 'positional parameters not preserved with . script without arguments'
+[[ $(PATH=:$PATH $SHELL -c '. dotscript;print $#') == 3 ]] || err_exit 'positional parameters not preserved with . script without arguments'
```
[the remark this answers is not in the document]
kernigh: "Most people" includes me; I didn't know that empty was `.` until I saw it in these tests.

src/cmd/ksh93/tests/functions.sh
krader1961: I agree with @kernigh. The unit tests have to assume there could be arbitrary external commands with names that conflict with ksh functions created by a test. However, I just checked and neither ksh93u nor this project's version runs the external command if I create a function with the same name in an interactive shell. Which is the expected, correct, behavior. It shouldn't be necessary to manipulate `PATH` in this manner. So what is different about the unit test environment that requires this hack?

krader1961: I looked at *functions.sh* and see that the problem isn't due to the function being named `zzz`. It's because the function is put in a file named `zzz` which is an external command found later in `PATH` than the system provided command of the same name. So there isn't any need to save/restore `PATH`. We simply need to ensure the temp dir that is created, and which is the CWD when the test runs, is at the front of `PATH`. And that should be done for every unit test. Just as every unit test creates a temp dir and cd's to it. So rather than merge this change, and more generally do this sort of thing in each individual unit test, we should create a standard way to have a preamble and postscript for each unit test. I'll take care of that as the current situation bugs me.

Issue #32, opened 2017-03-15
VAR=value followed by nested function call fails to put VAR into environment


sckendall2:
The following code should print V=1, but doesn't:

```ksh
function f2 { env | grep '^V='; }
function f1 { f2; }
V=1 f1
```

However, `V=1 f2` behaves correctly.

Every ksh93 I tried has the same behavior, including the following:

- Version M-12/28/93e (AIX 6.1 /bin/ksh93)
- Version AJM 93u+ 2012-08-01 (Red Hat Enterprise Linux Server release 7.2 /bin/ksh93)

zakukai:
There's a problem in your example ("env" is an external command, so for it to show $V, $V must be not only set in f2's scope, but also exported)

```
$ function f2 { echo "f2: VF1=$VF1"; }
$ function f1 { f2; }
$ VF1=valkyrie f1
f2: VF1=valkyrie
```

However, local variables are another story:
```
$ function f1 { typeset var VF19; VF19=excalibur; echo "f1: VF1=$VF1, VF19=$VF19"; f2; }
$ function f2 { echo "f2: VF1=$VF1, VF19=$VF19"; }      
$ VF1=valkyrie f1
f2: VF1=valkyrie, VF19=excalibur
f2: VF1=valkyrie, VF19=
```

$VF19 is local to function f1, so it isn't accessible from f2. (Korn Shell implements static scoping: Since $VF19 is not local to f2, $VF19 in f2 always refers to the variable in _global scope_)

```
$ VF19=YF19
gcaswell-vm:/home/onelink/misc_src/shell_bugs $ f1
f1: VF1=, VF19=excalibur
f2: VF1=, VF19=YF19
# f1 has a local binding for VF19 which overrides global scope in f1. But f2 still gets the copy from global scope.
```

However, it seems Korn Shell's "static typing" can be broken by invocation-level bindings:
```
$ unset VF19
$ VF19=YF19 f1      # VF19 is an "invocation-level" binding on f1
f1: VF1=, VF19=excalibur
f2: VF1=, VF19=excalibur
$ echo ${VF19-'{unset}'}
{unset}
```

Basically, when f1 is invoked, it inherits VF19 as an "invocation-level" binding - a temporary global-scope binding (I guess). As a result, f1 can't declare VF19 as local, and when f1 changes the value of VF19, the binding affects f2 as well.

(My tests are on version 93u+ 2012-08-01, RHEL 7)

sckendall2:
[...]

PR #467, opened 2018-04-10
Fix exporting variables from function local scope

siteshwar:
While copying variables from function's local scope to a new scope,
variable attributes were not copied. Such variables were not marked
to be exported in the new function. For e.g.

function f2 { env | grep -i "^foo"; }
function f1 { env | grep -i "^foo"; f2; }
foo=bar f1

prints 'foo=bar' only once, but it should print be twice. This patch
fixes it.

Resolves: #32

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -2848,6 +2848,7 @@ static void local_exports(Namval_t *np, void *data) {
     if (cp && (mp = nv_search(nv_name(np), shp->var_tree, NV_ADD | HASH_NOSCOPE)) &&
         nv_isnull(mp)) {
         nv_putval(mp, cp, 0);
+        mp->nvflag = np->nvflag;
     }
 }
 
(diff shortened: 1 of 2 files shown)


krader1961:
LGTM

siteshwar:
@krader1961 Thanks for the review. However, I am bit skeptical about merging this change. I will let this pull request stay open for few more days and see if there are any objections. @sckendall2 it would help if you can explain where did you encounter this issue.

sckendall2:
I don't remember how I encountered this issue.

But at the risk of talking your ear off, I'm not exaggerating much to say that it keeps me up at night. I work on the underpinnings of a large code base of KornShell (among other languages). This issue means that innocuous changes can break working scripts, and in a way that's subtle and hard to find by inspection.

Here's a simplistic example. We start with a working code snippet:
```Shell
function print_ids { ... | sort; }
LC_ALL=C print_ids > ids # C locale ensures consistent order
```
Now a well-meaning developer inserts a function wrapper to protect against variations in PATH:
```Shell
function sort { /usr/bin/sort "$@"; }
```
The wrapper breaks the snippet, because /usr/bin/sort no longer sees LC_ALL. If you imagine that these three lines of code are far away from each other, you can see how difficult it is to spot the problem.

This issue occurs only in ksh93,  not in mksh or other KornShell implementations I've checked.

siteshwar:
@sckendall2 You have a fair point, so I am going to merge this change.

Issue #468, opened 2018-04-11
Shall we switch back to fork()/exec() ?


siteshwar:
Forking a new process with `posix_spawn()` can result in a race condition while setting terminal foreground process group. It was discussed earlier in this [thread](https://www.mail-archive.com/ast-developers@research.att.com/msg00717.html). This is dgk's response from the thread:

```
ksh93 uses posix_spawn() rather than fork/exec to run a simple command for
performance reasons.  However, posix_spawnv() doesn't have a way to
set the terminal group before the exec().  The parent process sets the
terminal group which leads to the race.

There is a simple change you can make to a program to guarentee that
the terminal will get set at the beginning of the program.
You can add two lines.  One line is
        #include        <termios.h>

and for the other line  
                tcsetpgrp(2,getpgrp());
Alternatively, you can write a program that does these two lines and
then execs the original program.
```

This problem has been seen by multiple users and reported in [Red Hat bugzilla 1295563](https://bugzilla.redhat.com/show_bug.cgi?id=1295563). Only reliable way to fix this problem is to switch back to fork()/exec(). I propose that we should enable `_AST_no_spawnveg` flag by default.

siteshwar:
`posix_spawn()` is used due to performance reasons. We should analyze how much performance impact we will have if we stop using it. We can enable `_AST_no_spawnveg` as an experiment and see if it causes performance degradation or any other regressions.

kernigh:
I reproduced this bug by adding a sleep(1) so ksh loses the race with the child.

```diff
diff --git a/src/cmd/ksh93/sh/xec.c b/src/cmd/ksh93/sh/xec.c
index 9d037351..4574ce6e 100644
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -3358,6 +3358,7 @@ static pid_t sh_ntfork(Shell_t *shp, const Shnode_t *t, char *argv[], int *jobid
         if (grp == 1) job.curpgid = spawnpid;
 #ifdef SIGTSTP
         if (grp > 0 && !(otype & FAMP)) {
+            sleep(1);
             while (tcsetpgrp(job.fd, job.curpgid) < 0 && job.curpgid != spawnpid) {
                 job.curpgid = spawnpid;
             }
```

Then I entered this Python command in an interactive ksh:

```
$ python3 -c 'import os; print(os.getpgrp(), os.tcgetpgrp(0))'
32695 98273
```

[...]

PR #470, opened 2018-04-12
Use fork()/exec() instead of posix_spawn() by default

siteshwar:
Using posix_spawn() may result in a race condition while setting
terminal foreground process group. Copying dgk's response from related
discussion in ast-developers mailing list:

"
ksh93 uses posix_spawn() rather than fork/exec to run a simple command
for performance reasons.  However, posix_spawnv() doesn't have a way to
set the terminal group before the exec().  The parent process sets the
terminal group which leads to the race.

There is a simple change you can make to a program to guarentee that
the terminal will get set at the beginning of the program.
You can add two lines.  One line is
        #include        <termios.h>

and for the other line
        tcsetpgrp(2,getpgrp());

Alternatively, you can write a program that does these two lines and then
execs the original program.
"

Only way to reliably fix this problem in ksh is to switch back to fork()/
exec().

If there are any performance regressions or bugs caused by this change,
you can switch to posix_spawn() by setting '-D_AST_no_spawnveg=0'.

Related: #468
Related: rhbz#1160482

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -66,6 +66,7 @@ endif
 
 shared_c_args = [
     '-DUSAGE_LICENSE=""',
+    '-D_AST_no_spawnveg=1',
 ]
 
 # Compiler arguments used when building `ksh93`, `shcomp`, and related AST
--- a/src/cmd/ksh93/include/fault.h
+++ b/src/cmd/ksh93/include/fault.h
@@ -97,10 +97,19 @@ struct checkpt {
 #endif
 };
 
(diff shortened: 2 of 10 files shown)


siteshwar:
Test cases have started failing with this change. They are not failing if I build with ```-D_AST_no_spawnveg=1``` flag in last stable release. I suspect these failures are related to spawnvex family of functions that were introduced after the last stable release.

krader1961:
LGTM

Issue #472, opened 2018-04-15
Ksh's .sh.file contains garbage after returning from "sourcing" another file.


kwwall:
These results are based on the lastest AST ksh93, built from sources pulled from GitHub as of a few days ago. In a nutshell, initially ksh's value for .sh.file starts out correctly and is correct while inside a "sourced" (i.e., "dotted" script; e.g., ". ./foo.ksh"), one the script returns from the "sourced" file, the value of .sh.file now contains garbage. I discovered this while attempting to debug @rocky kshdb.

Attached is a script (try.ksh) to reproduce this. This is the output of this script is shown here:

$ ./try.ksh    # Renamed to try.ksh.txt in order to attach to GitHub
This is KornShell - Version: Version ABIJ 93v- 2014-12-24
Linux Latitude-E4300 4.13.0-38-generic #43-Ubuntu SMP Wed Mar 14 15:20:44 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 17.10
Release:	17.10
Codename:	artful
1: .sh.file = '/home/wallk/try.ksh'
1: .sh.file -- base name = 'try.ksh'
1: PASS
This is foo.ksh, $0=./try.ksh, .sh.file=/home/wallk/foo.ksh
2: .sh.file = 'Pñd'
2: .sh.file -- base name = 'Pñd'
FAIL

[try.ksh.txt](https://github.com/att/ast/files/1913143/try.ksh.txt)

krader1961:
This is almost certainly another manifestation of use-after-free issue #398.

PR #473, opened 2018-04-16
Correctly restore value of .sh.file after sourcing

siteshwar:
Commit e3dd5b1 introduced a regression which was causing .sh.file to be
set to corrupted value after sourcing. This commit reverts part of it.

Resolves: #472


siteshwar:
Add a test case.

PR #474, opened 2018-04-16
Start building builtins for some standard commands from libcmd

siteshwar:
Start building builtins for these commands :

- basename
- chmod
- dirname
- getconf
- head
- mkdir
- logname
- cat
- cmp
- uname
- sync

These builtins are disabled by default and can be enabled by running `builtin <command name>`.

### DIFF
--- a/src/cmd/ksh93/data/builtins.c
+++ b/src/cmd/ksh93/data/builtins.c
@@ -137,7 +137,7 @@ const struct shtable3 shtab_builtins[] = {
     {CMDLIST(dirname)},
     {CMDLIST(getconf)},
     {CMDLIST(head)},
-//    {CMDLIST(mkdir)},
+    {CMDLIST(mkdir)},
 //    {CMDLIST(logname)},
 //    {CMDLIST(cat)},
 //    {CMDLIST(cmp)},
--- a/src/cmd/ksh93/data/builtins.c
+++ b/src/cmd/ksh93/data/builtins.c
@@ -133,7 +133,7 @@ const struct shtable3 shtab_builtins[] = {
 #include SHOPT_CMDLIB_HDR
 #else   // SHOPT_CMDLIB_HDR
     {CMDLIST(basename)},
-//    {CMDLIST(chmod)},
+    {CMDLIST(chmod)},
 //    {CMDLIST(dirname)},
 //    {CMDLIST(getconf)},
     {CMDLIST(head)},
--- a/src/cmd/ksh93/data/builtins.c
+++ b/src/cmd/ksh93/data/builtins.c
@@ -135,7 +135,7 @@ const struct shtable3 shtab_builtins[] = {
     {CMDLIST(basename)},
     {CMDLIST(chmod)},
     {CMDLIST(dirname)},
-//    {CMDLIST(getconf)},
+    {CMDLIST(getconf)},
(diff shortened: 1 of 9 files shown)


src/cmd/ksh93/data/builtins.c
```diff
@@ -41,18 +41,13 @@
...
 #define SHOPT_CMDLIB_DIR SH_CMDLIB_DIR
-#endif  // SHOPT_CMDLIB_DIR
+#else
 #ifndef SHOPT_CMDLIB_HDR
 #define SHOPT_CMDLIB_HDR <cmdlist.h>
-#endif  // SHOPT_CMDLIB_HDR
+#endif // SHOPT_CMDLIB_HDR
+#endif // SHOPT_CMDLIB_DIR
 
-#if _AST_VERSION < 20121001L
```
krader1961: This seems dubious to me. There are other places in the code that compare `_AST_VERSION` against various magic values. Shouldn't removing this test be done separate from enabling these builtins?

src/lib/libcmd/meson.build
```diff
@@ -1,8 +1,12 @@
-libcmd_files = ['context.c', 'cmdinit.c', 'stty.c']
+libcmd_files = ['context.c', 'cmdinit.c' ]
+
+libcmd_files += [ 'basename.c', 'stty.c', 'head.c' ]
```
krader1961: It's not obvious why these modules aren't just part of the previous list. Which means it probably needs a comment.

src/lib/libcmd/cmdext.h
```diff
@@ -72,7 +72,7 @@ extern  int	b_mkdir (int, char**, Shbltin_t*);
 //extern  int	b_sha512sum (int, char**, Shbltin_t*);
 //extern  int	b_stty (int, char**, Shbltin_t*);
 //extern  int	b_sum (int, char**, Shbltin_t*);
-//extern  int	b_sync (int, char**, Shbltin_t*);
+extern  int	b_sync (int, char**, Shbltin_t*);
```
krader1961: I'd really prefer we not include this one at this time. The only way this can be justified is if ksh is meant to be an alternative to something like [busybox](https://busybox.net/about.html). Which might be a reasonable objective but I think it needs to be discussed. The `sync` command is unlike every other builtin.

src/cmd/ksh93/data/builtins.c
siteshwar: I actually left a comment in the description of https://github.com/att/ast/pull/474/commits/f047655c02d943551a76911c4181f3b02f3a0652 but it may not be clear enough. So I will leave a comment in the source code.

src/lib/libcmd/cmdext.h
siteshwar: I have mixed feelings about it. It was built by default in the last stable and the implementation looks trivial, so it should not be too much of a burden to maintain. I think we should ship it in the next release.

krader1961: > It was built by default in the last stable.

I confirmed it's in ksh93u+ as shipped with macOS. So, okay, include it. But this is another example of the lack of clear thinking (such as including math functions like tgamma) leading to including a feature seemingly for no reason other than "we can so why not".

krader1961:
LGTM

PR #481, opened 2018-04-18
Unify the unit test framework

### LINKED ISSUE #429: The ksh `functions` unit test runs an external `zzz` command if found in `$PATH` rather than the function of the same name

### LINKED ISSUE #477: Building ksh on opensuse wipes out home directory
There are some serious issues with how test cases were run in the last stable release. I installed rpm sources for `ksh` on opensuse with `zypper source install ksh` and did `rpmbuild -ba ksh.spec`. Test cases were run during  the build and I saw errors like :

```
/tmp/ksh-build.hZy52m/tmp_1d.huj
test alias begins at 2018-04-17+19:44:20
/tmp/ksh-build.hZy52m/tmp_3a.92s
alias.sh[90]: /foo: cannot create [Permission denied]
chmod: cannot access '/foo': No such file or directory
	alias.sh[93]: hash -r -- foo failed, expected foo=/foo, got 
alias.sh[90]: /-foo: cannot create [Permission denied]
chmod: cannot access '/-foo': No such file or directory
	alias.sh[93]: hash -r -- -foo failed, expected -foo=/-foo, got 
alias.sh[90]: /--: cannot create [Permission denied]
```

At the end, my home directory was wiped out and filled with files like:

```
a  a-b  abc  abd  abe  aXb  b  bb  bcd  bdir  Beware  c  ca  cb  d  dd  de  man
```
These files are created by `ksh` test script. @bitstreamout Are you aware about this issue ? I am running a vanilla openSUSE Leap 42.3 system.

@krader1961 I have seen `.sh_history` file being modified when I run test cases on my fedora system. We should run the test cases in a sandboxed environment.

krader1961:
This modifies how ksh unit tests are run to have a consistent
environment (i.e., preamble) and post test behavior (i.e., postscript).
There was at least one unit test that did not correctly report the
number of errors via the exit status.

Each test is now run in a unique temporary directory. If the test
doesn't complete without error the temp dir is not removed. This makes
it easier when running the tests on a personal system (rather than, say,
a CI environment like Travis) to explore the state of the unit test.

The `HOME` env var is now set to the temp dir to help keep the test from
modifying the user's environment (e.g., their ~/.sh_history file).

Resolves #429
Resolves #477


krader1961:
Note that this is just one step to making the ksh unit tests more consistent and hermetic. Not to mention making errors more useful be reporting expected and actual values in a consistent manner. I tried to keep this change to the minimum needed for the initial refactoring to use a test runner script that supplies a known preamble and postscript to each unit test.

This includes changes to a few sub-unit tests because those changes were necessary to understand why the original preamble/postscript refactoring broke the test. Similar changes will be made to all ksh unit tests in a subsequent commit.

siteshwar:
@krader1961 Path tests are consistently failing on fedora on Travis with this change.

krader1961:
Actually, they're not consistently failing. It's somewhere around 20% of the time. If you look closely at the test results you'll see that it's always (as far as I've observed) just one of the three path runs. And it's not always the same variant. Too, it is a single test that is failing:

```
(
    rm -rf noexec
    print 'print cannot execute' > noexec
    noexec > /dev/null 2>&1
)
actual=$?
expect=127
[[ $actual == $expect ]] || err_exit "exit status of non-executable is wrong" "$expect" "$actual"
```

This test should be 100% deterministic. There are no sleeps, no races, and it doesn't matter how overloaded the system might be. But once in a while the exit status is 126 which is supposed to mean the command could not be found. See this line in *lib/package/ast-open.html*:

> 12-10-16 spawnvex.c: spawnvex_open(SPAWN_EXEC) =&gt; posix_spawn() exit ENOENT=126 ENOEXEC=127 ok

Yet the command is in fact present and in `$PATH`. So it should always return 127 (the ENOEXEC case).

There is definitely a bug in the code. The test behavior, coupled with this happening more often after my recent refactoring, suggests the use of an uninitialized stack var or something similar.

Issue #478, opened 2018-04-17
Backticks nested in `$( )` result in misdirected output


sckendall2:
In the following, either the /bin/echo command or the first print command must be sending their output directly to the script's stdout:
```
$ foo=$(print `/bin/echo bar`)   # should print nothing
bar
$ print foo=$foo                 # should print "foo=bar"
foo=
$ 
```
This bug is in the latest code, fetched earlier today. It's a regression from older versions of ksh93, such as the one that ships with RHEL 7.

krader1961:
Note that this bug exists in the `2016-01-10-beta` branch. It's not something introduced in the `master` branch after we forked from the beta branch and changed the build system.

sckendall2:
By "the one that ships with RHEL 7" I mean `Version AJM 93u+ 2012-08-01`, RPM `ksh-20120801-22.el7_1.2.x86_64`, on RHEL 7.2. That version of ksh93 doesn't have the bug. Nor do ksh93 versions and other KornShells on a wide variety of Linux and other OSes.

siteshwar:
@sckendall2 We have built current development version from the last beta release by changing the build system, reformatting code etc. @krader1961's comment means that this bug was not introduced by those changes. It helps us in narrowing down how the regression got introduced.

sckendall2:
@siteshwar Whoops, I didn't mean my comment as a response to @krader1961. I started my comment before his appeared. I was just clarifying my initial post.

siteshwar:
fwiw this works as expected:

```
x=$(print $(/bin/echo bar))
```

ksh handles input output for $() and `` differently. We have seen multiple issues with each of these syntaxes in past. Unfortunately the code to handle input/output in subshells is too complicated and it's very likely that fixing each bug will create another regression.

McDutchie:
An effective workaround is forcing ksh93 to fork off the command substitution subshell in a separate process by including a dummy redirection in it:

    $ foo=$( : 1>&1; print `/bin/echo bar`)
    $ echo "$foo"
    bar

PR #482, opened 2018-04-18
Check for invalid file descriptor in sh_close()

siteshwar:
There was a regression introduced after the last stable release which
causes mishandling of redirections in subshells. For e.g.

foo=$(print `/bin/echo bar`)
print foo=$foo

prints `bar` on stdout and the variable foo is empty.

Handling of redirections in $() style of command substitution was
changed after the last stable release. sh_subshell() function replaces
sfstdout with a newly created temporary stream with this line :

sp->saveout = sfswap(sfstdout, NULL);

The newly created stream has it's fd value set to -1. If `` style of
command subtitution happens inside $(), it tries to use this stream with
invalid fd and fails to close it. So redirections for stdout do not
work.

We should check if the stream that we are closing actually corresponds
to the file descriptor that we want to close. If not, directly invoke
close() on the fd.

Resolves: #478

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -445,7 +445,7 @@ int sh_close(int fd) {
         return -1;
     }
 
-    if (!(sp = shp->sftable[fd]) || (sp->_file != fd) || (sfclose(sp) < 0)) {
+    if (!(sp = shp->sftable[fd]) || (sffileno(sp) != fd) || (sfclose(sp) < 0)) {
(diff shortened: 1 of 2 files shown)


krader1961:
LGTM

siteshwar:
I have made one minor change in the pull request to use `sffileno()` instead of directly accessing `_file` member of `Sfio_t` struct. There may be test case failures on Travis, but they are caused by recent changes in framework for testing (#481). 

Merging.

PR #486, opened 2018-04-20
Fix tvtouch() implementation

### LINKED ISSUE #459: Function `tvtouch()` is broken
The `tvtouch()` function is broken if the system doesn't have the `utimensat()` function (i.e., `_lib_utimensat` isn't true). I noticed this today when I happened to leave an interactive ksh running for more than a few minutes. It core dumped when `tvtouch()` attempted to dereference the `av` parameter which was NULL. If the system has a usable `utimensat()` that is handled correctly. But the two alternative implementations do not handle that case correctly. This bug is present in the beta branch (i.e., predates all the work that's been done in the past year).

krader1961:
When the code was refactored as part of switching to the Meson build
system some statements were put inside the wrong `#if` blocks.

Also, the `timets()` support is only for IRIX which is a long dead OS so
remove that support.

Fixes #459

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -219,7 +219,6 @@
 #mesondefine _lib_utime
 #mesondefine _lib_utime_now
 #mesondefine _lib_utimensat
-#mesondefine _lib_utimets
 #mesondefine _lib_wcrtomb
 #mesondefine _lib_wcscmp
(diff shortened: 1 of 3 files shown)


siteshwar:
LGTM

PR #487, opened 2018-04-21
Remove the suid_exec code

### LINKED ISSUE #366: Does any distro ship the `suid_exec` program?
At this time we don't build the `suid_exec` program (*src/cmd/ksh93/sh/suid_exec.c*). It is built by the legacy build system. That program is not on any of the distros I use or have setup for testing changes to this project. Including, macOS, OpenSuse, FreeBSD, and Ubuntu. Should we add it to the Meson config or remove that source module?

krader1961:
The suid_exec program has not been included by any distro as far as I
can tell in a very long time. Certainly predating the ksh93u+ release
that is the basis of all ksh implementations as I write this. So jetison
that dead weight.

Resolves #366


src/cmd/ksh93/sh/path.c
```diff
@@ -1178,20 +1177,14 @@ static void exscript(Shell_t *shp, char *path, char *argv[], char *const *envp)
...
         }
     }
-    savet = *--argv;
     *argv = path;
-    if (err == EACCES && sh_access(e_suidexec, X_OK) < 0) {
-        errno = EACCES;
-        return;
-    }
-    path_pfexecve(shp, e_suidexec, argv, envp, 0);
+    if (err == EACCES) return;
```
siteshwar: How does this function handle executing a script if `/etc/suid_exec` is not present ? I would appreciate a comment in the source code here.

krader1961: I'm going to revert this part of the change. Not because it's wrong. As far as I can tell via code inspection and testing this part of the change is itself okay. But looking more closely at the surrounding code I think the original code is broken. Take a look at the block inside

```
if ((euserid = geteuid()) != shp->gd->userid) {
```

It creates a temp file and assigns fd 10 to it. It then falls thru and calls `path_pfexecve()` to execute */etc/suid_exec* without any reference to that temp file. If the execve fails it then falls through to the `fail:` label and the temp file created earlier isn't used.

So instead I'm going to leave a "TODO" comment to remind someone to rewrite this function.

krader1961:
Merged with the change I mentioned above.

siteshwar:
Thanks!

Issue #485, opened 2018-04-20
Test for command file without exec bits set in paths.sh is flakey


krader1961:
From PR#481:

There is a single `path.sh` test that is flakey:

```
(
    rm -rf noexec
    print 'print cannot execute' > noexec
    noexec > /dev/null 2>&1
)
actual=$?
expect=127
[[ $actual == $expect ]] || err_exit "exit status of non-executable is wrong" "$expect" "$actual"
```

This test should be 100% deterministic. There are no sleeps, no races, and it doesn't matter how overloaded the system might be. But once in a while the exit status is 126 which is supposed to mean the command could not be found. See this line in *lib/package/ast-open.html*:

> 12-10-16 spawnvex.c: spawnvex_open(SPAWN_EXEC) => posix_spawn() exit ENOENT=126 ENOEXEC=127 ok

Yet the command is in fact present and in `$PATH`. So it should always return 127 (the ENOEXEC case).

There is definitely a bug in the code. The test behavior, coupled with this happening more often after my recent refactoring of the unit tests, suggests the use of an uninitialized stack var or something similar.

siteshwar:
From [section 2.8.2](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) of POSIX:

```
If a command is not found, the exit status shall be 127. If the command name is found, but it is not an executable utility, the exit status shall be 126. Applications that invoke utilities without using the shell should use these exit status values to report similar errors.
```
So the expected exit value of above test should be `126` and not `127`.

krader1961:
> So the expected exit value of above test should be 126 and not 127.

So the ksh implementation is wrong. Is that something we dare change? Fixing that is probably the right thing to do since POSIX conformance on this point is probably more important than backward compatibility on this obscure point. Too, either way the test randomly returns the wrong value (whether that is 126 or 127) and thus there is something seriously wrong.

P.S., Note that I can't find anywhere in the code where it explicitly returns 126 or 127 as the status. There are places that return -1 for the status. And that would get mapped to 127 due to how twos-complement ints and UNIX status values work. Presumably there is someplace that returns -2 as the status which would correspond to 126.

siteshwar:
> So the ksh implementation is wrong. Is that something we dare change? Fixing that is probably the right thing to do since POSIX conformance on this point is probably more important than backward compatibility on this obscure point. Too, either way the test randomly returns the wrong value (whether that is 126 or 127) and thus there is something seriously wrong.

[...]

PR #488, opened 2018-04-23
Keep exit status consistent when command is not executable

siteshwar:
Ksh performs command searching by trying to execute command in each
element of PATH. If a command is found, but is not executable, ksh
continues to search for it further in other paths.

According to POSIX, if shell finds a command which is not executable,
it should return with status 126. Current behavior of ksh is non
determinstic as it depends on if command is found in the last element
of PATH or not:

1. If a command is found in the last element of PATH but is not
executable, ksh returns with status 126.

2. If a command is found before last element of PATH but is not
executable, ksh continues to search  for it further in other paths. It
returns with status 127 when command could not be found.

Ksh should behave consistently and return immediately after a command is
found but is not executable.

Resolves: #485

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -901,15 +901,19 @@ void path_exec(Shell_t *shp, const char *arg0, char *argv[], struct argnod *loca
                 opath = arg0;
             }
             path_spawn(shp, opath, argv, envp, libpath, 0);
+            // Command exists but failed to exec
+            if (shp->path_err != ENOENT) {
+                // Exit with status 126
+                // https://github.com/att/ast/issues/485
+                ((struct checkpt *)shp->jmplist)->mode = SH_JMPEXIT;
+                errormsg(SH_DICT, ERROR_system(ERROR_NOEXEC), e_found, arg0);
+            }
             while (pp && (pp->flags & PATH_FPATH)) pp = path_nextcomp(shp, pp, arg0, 0);
         } while (pp);
-    // Force an exit.
+
+    // Command not found, force an exit.
     ((struct checkpt *)shp->jmplist)->mode = SH_JMPEXIT;
-    if ((errno = shp->path_err) == ENOENT) {
-        errormsg(SH_DICT, ERROR_exit(ERROR_NOENT), e_found, arg0);
-    } else {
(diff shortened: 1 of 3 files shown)


siteshwar:
@kdudka I would appreciate your comments on this.

krader1961:
I'll review ASAP but this PR caused me to notice a bogosity. One reason I couldn't readily find how the exit status 127 and 126 were defined is because they are defined in an oblique manner that obscures the actual values. From *src/lib/libast/include/ast.h*:

```
#define EXIT_BITS 8 /* # exit status bits       */

#define EXIT_USAGE 2                               /* usage exit code   */
#define EXIT_QUIT ((1 << (EXIT_BITS)) - 1)         /* parent should quit        */
#define EXIT_NOTFOUND ((1 << (EXIT_BITS - 1)) - 1) /* command not found */
#define EXIT_NOEXEC ((1 << (EXIT_BITS - 1)) - 2)   /* other exec error  */
```

Would you, pretty please, as part of this PR change those definitions to the simple numerical values of 128, 127, and 126 respectively? Since `EXIT_BITS` can never be anything other than 8 there is negative value in defining the values like this. A comment that `EXIT_BITS` must remain 8 is also a good idea. Longer term the other definitions related to these values need to be replaced by those found in the *sys/wait.h* system header.

src/cmd/ksh93/sh/path.c
```diff
@@ -901,15 +901,19 @@ void path_exec(Shell_t *shp, const char *arg0, char *argv[], struct argnod *loca
                 opath = arg0;
             }
             path_spawn(shp, opath, argv, envp, libpath, 0);
+            // Command exists but failed to exec
+            if (shp->path_err != ENOENT) {
+                // Exit with status 126
```
krader1961: I would replace this line with something like "// Assume that any other error (e.g., EACCES) means the file was found but could not be executed."

krader1961:
LGTM

siteshwar:
Let's hold this until we get clarification about POSIX requirements in https://github.com/att/ast/issues/485#issuecomment-384258496

siteshwar:
I think we should stay close to what bash and other shells do, so I am going to rewrite this patch.
I have already cherry picked 1be9edcbe6115e47d1c4f24c086324c34be3eaa1 and will open a separate pull request to fix #485.

PR #489, opened 2018-04-24
In case statement, stop trying pattern as literal string

### LINKED ISSUE #476: case: '[0-9]' matches [0-9]
$ ksh -c 'case "[0-9]" in [0-9]) echo BUG;; *) echo ok ;; esac'
    BUG

Apparently, if the pattern fails to match as a pattern, ksh checks if it matches as a literal string as a fallback.

As Geoff Clare [pointed out on austin-group-l](https://www.mail-archive.com/austin-group-l@opengroup.org/msg02127.html), this breaks a common input validation use case:

    case $input in
    [0-9]) ;;
    *) echo "Input must be a single digit"; exit ;;
    esac

This will fail to catch an error if `"$input"` equals `'[0-9]'`.

This is historical behaviour going back to ksh88 and the historical Bourne shell, but it's completely broken and no other current POSIX shell of note implements it.

It directly contravenes the text of the POSIX standard, which says unquoted `[0-9]` is a bracket pattern and not a literal string.

kernigh:
`case "[0-9]" in [0-9]) echo match;; esac` now stops matching. When a
case statement doesn't match a pattern -- as the string "[0-9]"
doesn't match the pattern [0-9] for a single digit -- then it no
longer tries to use the pattern as a literal string.

Now `case word in pattern)` in ksh behaves like `[[ word = pattern ]]`
in ksh and case statements in most other shells.

Resolves #476

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ was subsequently created to document the work being done.
 None at this time.
 
 ## Notable non-backward compatible changes
+
+- `case "[0-9]" in [0-9]) echo match;; esac` has stopped matching. When a case
+  statement doesn't match a pattern, it no longer tries to use the pattern as
+  a literal string (issue #476).
 - echo builtin now interprets escape sequences and parses command line options
   on all platforms. (issue #370)
 - Support for the UWIN environment has been removed (issue #284).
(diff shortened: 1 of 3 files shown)


krader1961:
Brilliant. LGTM. And you even thought to update *CHANGELOG.md*. Merged as 9b8391ab9.

kernigh:
@siteshwar pointed me to CHANGELOG.md in #476.

PR #491, opened 2018-04-25
Fix compiler warnings in xec.c

kernigh:
clang warned that variable `restorevex` and function `io_usevex()` are
unused.  Fix by adding `#if ... #endif`.


siteshwar:
LGTM

PR #493, opened 2018-04-27
Keep exit status consistent for non-executable commands

siteshwar:
POSIX has 2 different statements about searching commands in PATH:

If a command is not found, the exit status shall be 127. If the command
name is found, but it is not an executable utility, the exit status
shall be 126. Applications that invoke utilities without using the shell
should use these exit status values to report similar errors.

and

The list shall be searched from beginning to end, applying the filename
to each prefix, until an executable file with the specified name and
appropriate execution permissions is found.

It means that shell should not stop searching if a command is found, but
not executable. And it should return 126 only when executable command is
not present in other elements of PATH.

Ksh already complies with this behavior. However if a command is found,
but not executable, it was returning with wrong exit code in some cases:

1. If a command is found in the last element of PATH but is not
executable, ksh returns with status 126.

2. If a command is found before last element of PATH but is not
executable, ksh continues to search for it further in other paths. It
returns with status 127 when executable command could not be found in
other paths.

It should behave consistently and return with status 126 when a command
is found in PATH but is not executable.

Resolves: #485

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -876,6 +876,7 @@ void path_exec(Shell_t *shp, const char *arg0, char *argv[], struct argnod *loca
     const char *opath;
     Pathcomp_t *libpath, *pp = 0;
     int slash = 0;
+    int not_executable = 0;
 
     sh_setlist(shp, local, NV_EXPORT | NV_IDENT | NV_ASSIGN, 0);
     envp = sh_envgen(shp);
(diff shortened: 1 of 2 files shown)


DavidMorano:
@siteshwar 
> It should behave consistently and return with status 126 when a command
is found in PATH but is not executable.

For what it is worth, this was my opinion also.  I refrained from mentioning this earlier in fear of causing more confusion.  But if I understand your proposal above, you are advocating:

+ as KSH searches the PATH, if it finds a name match that is also executable, it executes it

+ if KSH finds a match in the PATH but the match is *not* executable, KSH records (perhaps by setting a bit) that it found a non-executable match and  *continues* searching for more matches

+ if KSH fails to find in the whole of PATH any matches that are also executable, KSH sets the exit status as:
-- 126 - KSH found a match but it was not executable
-- 127 - KSH did not find any matches at all, executable or not

Is this what you are proposing?

siteshwar:
> Is this what you are proposing?

Yes.

Issue #483, opened 2018-04-19
test failures in OpenBSD


kernigh:
These are the other test failures in OpenBSD, after the recent fixes of #429 (zzz problem) and #443 (vfork problem). I attach the test log from commit f5b9c64 in OpenBSD 6.3 on amd64.

Attachment: [testlog.txt](https://github.com/att/ast/files/1926539/testlog.txt)

### arith, comvario, treemove: bug in OpenBSD

```
  7/153 arith                                   FAIL     0.57 s
--- stderr ---
<E> arith[592]: printf '%g\n' Nan failed -- expected nan, got -nan
<E> arith[594]: print -- $((Nan)) failed -- expected nan, got -nan
<E> arith[592]: printf '%g\n' -Nan failed -- expected -nan, got nan
<E> arith[594]: print -- $((-Nan)) failed -- expected -nan, got nan
<W> arith[-1]: error_count = 4

 34/153 comvario                                FAIL    10.15 s
--- stderr ---
/tmp/ksh.comvario.uR9mZ3s/comvario.sh: warning: -lm does not support signbit(-Na
N)
<E> comvario[241]: Expected ...my_nan=nan... == ...my_nan=-nan...

137/153 treemove                                FAIL     0.29 s
--- stderr ---
<E> treemove[328]: typeset -m not working when passed a reference to an local argument from a calling function
<E> treemove[67]: typeset -C -a c.ar=((typeset -l -E i=-nan) (typeset -l -E i=inf) (typeset -l -E i=24) (typeset -l -E i=4) (typeset -l -E i=2) (typeset -l -E i=1) (typeset -l -E i=-1) (typeset -l -E i=-inf)) is not idempotent
<W> treemove[-1]: error_count = 2
```

*Not a bug in ksh.* This is a bug in OpenBSD/amd64, where libc functions like [strtod(3)](https://man.openbsd.org/strtod) flip the sign so "nan" becomes -nan and "-nan" becomes nan. I hacked my libc to fix strtod(), but I didn't fix strtold(), so I still see failures. I sent a bug report to OpenBSD: https://marc.info/?l=openbsd-bugs&m=152391654313171&w=2

### directoryfd: no /proc, no /dev/fd/11/*

```
 43/153 directoryfd                             FAIL     2.11 s
--- stderr ---
<E> directoryfd[116]: ...got 'mkdir: /proc/53773/fd/11//test1: No such file or directory'
[...]

PR #495, opened 2018-04-28
fix or skip some tests in OpenBSD

kernigh:
I reported in #483 that some tests fail in OpenBSD, because of differences between OpenBSD and other systems. These commits edit the tests sh_match.sh, signal.sh, and variables.sh to give "OK" on OpenBSD.

The remaining tests that "FAIL" are these:

- pty.sh might be ksh's bug.
- arith.sh, comvario.sh, treemove.sh fail because of OpenBSD's bug in strtold().
- coprocess.sh, sigchld.sh are often "OK" but have a chance to "FAIL".


krader1961:
LGTM. Much thanks, @kernigh. I'll merge later tonight when I'm home.

Issue #417, opened 2018-02-19
Fix to parameter substitution case modification operators


perette:
Attached is a patch to fix the case modification operators `${name^pattern}` and `${name,pattern}`.  It now uses shell patterns, and replaces the matched portion with upper or lower case.  These are consistent with the behaviour of other parameter substitution forms, and closer to bash's operator, although bash exhibits some arbitrary limitations.  Further down in this thread is a comparison of prior ksh, this patch, and bash behaviors.  [ksh-case-modification.patch.txt](https://github.com/att/ast/files/1759513/ksh-case-modification.patch.txt)

The original code's behavior was (for ^ and ,) to modify the first character only, and then only if the first character of the pattern was a '?'; and (for ^^ and ,,) to modify the whole string, regardless of the pattern.

Included is an update to the man page for these operators, to clarify and bring the text into line with similar operators.

>   Case  modification.   This expansion modifies the case of alphaetic characters in parameter.  The pattern is expanded to  produce a  pattern just as in pathname expansion.  The  ^  operator converts  lowercase  characters  matching  pattern to uppercase.  The  , operator converts matching uppercase characters to lowercase.   The ^^ and ,, expansions  convert  all  matched  character  in  the expanded  value.  The  ^ and , expansions match and convert only the first character in the expanded value.  If pattern is  omitted it is treated like a  ?,  which  matches  every character.  If parameter is @ or *, or an array subscripted by @ or *, the case modification operation is  applied to  each  element.

The proposed new text reads:

>    Case modification.  If the shell _pattern_ matches the value of _parameter_, then this expansion is the value of the _parameter_ with alphabetic characters in the matched portion changed to uppercase for **^** and **^^**, or to lowercase for **,** and **,,**.    For **^** and **,** the transformation is applied only if the match is at the beginning of the string.  For **^** and **,,** all matches are transformed.  If _pattern_ is unmatched, then _parameter_ is substituted unmodified. When _parameter_ is **@**, __*__, or an array variable with subscript **@** or __*__, the transformation is applied to each element in turn.

krader1961:
This looks like a backward incompatible change to me. Am I mistaken? Also, while I only glanced at the test cases the last example looks incorrect to me:

```
[array]          ${arr[@],,*([A-Ma-m])} -->arrr thar be PIRATES!
```

Why isn't the final `PIRATES` converted to lowercase? It matches the pattern and therefore should have been converted to lowercase it seems to me.

perette:
[...]

PR #496, opened 2018-04-30
Disable case modification in parameter substitution

siteshwar:
It seems this feature is inspired bash and it is a backward incompatible
with the last stable release. We should keep it disabled in the next
release.

Related: #417


krader1961:
LGTM

PR #497, opened 2018-05-01
Enable cut and wc builtins

siteshwar:
I am using implementation of these builtins from last stable release.
Related: #129


krader1961:
LGMT although this part of your commit comment worries me:

> ... different API for handling multibyte characters.

Do this mean we now have two different APIs for dealing with multibyte encodings? If so we probably need an open issue to remind us to pick one.

siteshwar:
@krader1961 There was change in macros that handle multibyte characters. [This](https://github.com/att/ast/blob/89efa5fba792b510a37505819ead6ba99d28702f/src/lib/libast/include/ast.h#L419-L453) is the related code block. Afaik libast code uses new macros and rest of the code still uses old macros.

PR #498, opened 2018-05-01
Remove support for SHOPT_ACCT and SHOPT_ACCTFILE

krader1961:
These two features are not enabled by default, are not enabled in any distro I have access to, and are huge security holes. No sane distro  would ever enable these features so remove them.


siteshwar:
I verified that this feature is not used by Fedora/RHEL, Opensuse or Ubuntu, so it should be safe to remove it.

siteshwar:
LGTM

PR #500, opened 2018-05-02
Enable test cases for standard commands

siteshwar:


siteshwar:
I think this should be fixed in Meson. I have opened an issue for it https://github.com/mesonbuild/meson/issues/3519

krader1961:
Fails with the following one or both of the following errors on macOS, opensuse and ubuntu 16.04:

```
<E> builtins[811]: basename bound to /bin/basename but should be bound to /usr/bin/basename
<E> builtins[812]: cmp bound to /bin/cmp but should be bound to /usr/bin/cmp
```

siteshwar:
@krader1961 Fixed.

krader1961:
LGTM, in as much as it no longer introduces new test failures on the platforms I'm testing against :smile: Also, I have no objection to the text of the change. But I'm far from happy about how builtin commands of this sort are exposed and tested. That, however, is a discussion for another day.

PR #502, opened 2018-05-04
Remove SHOPT_ATFUN

krader1961:
Build time symbol `SHOPT_ATFUN` was not present in the last stable
release. It is in the `2016-01-10-beta` branch but is not enabled by
default. The `fd` member of `struct pathcomp` is initialized to 0 and set
to -1 in in `path_checkdup()`. It never has any other value (which I've
proven to my satisfaction via instrumenting the code). Since neither the
build time SHOPT_ATFUN symbol or `struct pathcomp` member `fd` serves a
purpose remove both.

Obviously the code predicated on `SHOPT_ATFUN` is a work in progress.
Whomever added that code intended it to serve some purpose in the
future.  But since we have no idea what the intended purpose is, and
removing it doesn't break anything, remove it.

P.S., Note that the `if (pp->fd) close(pp->fd);` statement in function
`path_delete()` is broken given that `pp->fd` can have the value -1 to
mean the fd is invalid. Which means that test is true and it attempts
to close fd -1. Furthermore, why is a fd of zero not considered valid
at that point in the code? In other words, it should be set to -1, not
zero, when the structure is first allocated and that test should be `if
(pp->fd >= 0) close(pp->fd);`.

Part of resolving #210


siteshwar:
It should be safe to remove.

LGTM.

PR #505, opened 2018-05-05
Remove SHOPT_AMP and its associated code

krader1961:
This symbol hasn't been enabled by default at least as far back as
ksh93u+. Enabling it today (along with the other required options to
use it) results in compilation errors due to code rot. For example, it
still calls `spawnveg()` which no longer exists (it has been replaced by
`spawnvex()`). It also uses `shp->shpath` which is no longer a valid member
of that structure (it's now `shp->gd->shpath`). There are probably other
issues with the code as well. There is zero chance anyone will try to
figure out what this code is intended to do and complete the work to
make it usable. So remove it.

Part of resolving #210


siteshwar:
LGTM

krader1961:
Merged

Issue #501, opened 2018-05-02
Build code related to dynamically loadable builtins


siteshwar:
Last stable release of ksh supported dynamically loading builtins from shared object files. We should start building code related to it.

PR #506, opened 2018-05-05
Enable support for dynamically loaded builtins

siteshwar:
Resolves: #501


src/cmd/ksh93/bltins/typeset.c
```diff
@@ -914,8 +914,11 @@ int b_builtin(int argc, char *argv[], Shbltin_t *context) {
     Shbltin_f addr;
     Stk_t *stkp;
     char *errmsg;
+    void *library=0;
 #ifdef SH_PLUGIN_VERSION
+    unsigned long ver;
     int list = 0;
+    char path[1024];
```
krader1961: s/1024/PATH_MAX/ ?

src/lib/libdll/dlldefs.h
```diff
@@ -0,0 +1,170 @@
...
+*                                                                      *
+*               Glenn Fowler <glenn.s.fowler@gmail.com>                *
+*                                                                      *
+***********************************************************************/
+
+/* : : generated by proto : : */
+/* : : generated from /home/situ/test/ast/src/lib/libdll/features/dll by iffe version 2013-11-14 : : */
+                  
+#ifndef _def_dll_dll
+#if !defined(__PROTO__)
```
krader1961: What happens if you remove everything from here to line 81? AFAICT this is just boilerplate for dealing with pre-ANSI C compilers.

src/lib/libdll/dlldefs.h
```diff
@@ -0,0 +1,170 @@
...
+#      endif
+#    endif
+#  endif
+#endif
+#if !defined(__LINKAGE__)
+#define __LINKAGE__		/* 2004-08-11 transition */
+#endif
+
+#define _def_dll_dll	1
+#define _sys_types	1	/* #include <sys/types.h> ok */
```
krader1961: All of these defines need to be meson feature tests rather than being hardcoded here.

src/lib/libdll/meson.build
```diff
@@ -0,0 +1,14 @@
+libdll_files = ['dlfcn.c', 'dllcheck.c',  'dllerror.c',  'dllfind.c', 'dll_lib.c',  'dlllook.c',  'dllnext.c',  'dllopen.c',  
+                'dllplug.c', 'dllscan.c']
+
+libdll_c_args = shared_c_args + [
+    '-D_BLD_Dll=1',
```
krader1961: The `=1` isn't needed and isn't consistent with all the other true/false symbols we define in the build system.

src/lib/libdll/dllscan.c
```diff
@@ -314,7 +314,7 @@ int dllsclose(Dllscan_t *scan) {
 Dllent_t *dllsread(register Dllscan_t *scan) {
     register char *p;
```
krader1961: Might as well drop the other `register` keywords while you're touching this.

krader1961:
FYI, This introduces a SIGSEGV in the `fts_maxarglen()` function during the `builtins` test. Also, an `abort()` call in the `exit` test on macOS and openBSD but not FreeBSD:

```
/var/folders/6t/t624jkmx7cbb9t35vzvjwj180000gn/T/ksh.exit.XXXXXXX.kW341mLW/exit.sh: line 153: 14595: Abort
<E> exit[78]: ksh plugin exit failed -- was ksh built with CCFLAGS+=$(CC.EXPORT.DYNAMIC)?
```

I'll try to get more details about what is wrong.

src/lib/libdll/dllscan.c
```diff
@@ -355,7 +355,7 @@ Dllent_t *dllsread(register Dllscan_t *scan) {
             if (scan->flags & (DLL_MATCH_NAME | DLL_MATCH_VERSION)) {
                 sfstrseek(scan->tmp, scan->off, SEEK_SET);
                 if (!(t = sfstruse(scan->tmp))) return 0;
-                if ((scan->fts = fts_open((char **)t, FTS_LOGICAL,
+                if ((scan->fts = fts_open((char **)&t, FTS_LOGICAL,
```
krader1961: This is the cause of the SIGSEGV. The first arg to `fts_open()` is a NULL terminated list of strings (like main()'s `argv` parameter). You can't just take a pointer to the pointer to the string. You have to construct a two element array where the first element is `t` and the second is `NULL`.

krader1961: And, of course, you didn't introduce this bug, @siteshwar, since it looks like it was probably a day one bug.

krader1961: Changing line 358 to this fixes the SIGSEGV:

```
char *argv[2] = {t, NULL};
if ((scan->fts = fts_open(argv, FTS_LOGICAL,
```

krader1961: And it also fixes the failure in the `exit` unit test.

siteshwar: > And, of course, you didn't introduce this bug, @siteshwar, since it looks like it was probably a day one bug.

libast version of `fts_open()` supported this non-standard flag:

```
#define FTS_ONEPATH     0x00000020      /* pathnames arg is one const char*     */
```

so this bug got introduced when we switched to standard ```fts_open()``` function.

krader1961:
LGTM

PR #508, opened 2018-05-06
Enable test case for cd when HOME is unset

siteshwar:


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -919,15 +919,15 @@ cd $TEST_DIR
...
-#    cd 2> /dev/null
-#    [[ $(pwd) == "$home" ]]
-#EOF
-#[[ $? == 0 ]] || log_error 'cd with no arguments fails if HOME is unset'
+$SHELL <<- \EOF
+    # $HOME is set to a temporary directory by test framework
+    # Get actual home directory of the user
+    home=~$USER
+    unset HOME
+    cd 2> /dev/null
```
krader1961: I recommend dropping the redirect. If the cd succeeds nothing is written to stderr. If it fails it would be helpful to see the error message. In fact, you probably would have figured this out sooner had this stderr redirect not been present.

krader1961:
LGTM modulo the one suggestion to improve the test.

Issue #503, opened 2018-05-04
Should `SHOPT_HISTEXPAND` be removed or made the default


krader1961:
While working on issue #210 I noticed build time symbol `SHOPT_HISTEXPAND`. It is present in the last stable release (version ksh93u+) and is enabled by default in that build. The `hist -p` flag it enables is not documented in the `man ksh` output. It is, however, available in that version and using it produces no error (unlike, say, `hist -z`).

That build time symbol is also present in the current beta release (the `2016-01-10-beta` branch) but is no longer enabled by default. In fact, if you type `hist -p` you get dropped into your preferred editor and upon exiting it you're told that `hist -a` is invalid.

If I explicitly enable `SHOPT_HISTEXPAND` then the build fails because the `hist_expand()` function now takes three arguments but the invocation hidden behind that build symbol only passes two parameters.

P.S., I have no idea what this feature is intended to do since the `csh` command doesn't seem to have a `-p` flag for its `history` command. The code for the ksh93u+ release says:

```
SHOPT_HISTEXPAND == 1            /* csh-style history file expansions */
```

I vote we drop this build time symbol and the code predicated on it.

krader1961:
> P.S., I have no idea what this feature is intended to do since the csh command doesn't seem to have...

Yes, I could dig deeper into the ksh code to understand what enabling this "feature" does. That isn't the point. The point is that the feature isn't documented in the last stable release. So it should be safe to remove. Too, it isn't enabled in the current branch and doesn't build if enabled.

siteshwar:
It is documented under `history --help` and is used to see output of history expansions (when histexpand is turned off). For e.g. `history -p !2` will show second command from history. Despite of it's questionable usefulness, I will prefer to keep it.

@krader1961 I will take care of fixing it.

krader1961:
FWIW, my comment about it being undocumented pertains to the man page. The fact that internal documentation differs from the man page is itself another, big, problem. Note that `hist --help` also lists a `-N` flag that is not in the man page.

PR #509, opened 2018-05-06
Build code under SHOPT_HISTEXPAND by default

siteshwar:
Remove this macro and build code related to it unconditionally.
Resolves: #503

### DIFF
--- a/src/cmd/ksh93/bltins/hist.c
+++ b/src/cmd/ksh93/bltins/hist.c
@@ -29,9 +29,7 @@
 #include "io.h"
 #include "name.h"
 #include "variables.h"
-#if SHOPT_HISTEXPAND
 #include "edit.h"
-#endif
 
 #define HIST_RECURSE 5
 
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/meson.build
```diff
@@ -26,6 +26,7 @@ shared_c_args += [
     '-DSHOPT_COSHELL',
     '-DREAD_TIMEOUT=' + '@0@'.format(get_option('read-timeout')),
     '-DAUDIT_FILE=' + '"@0@"'.format(get_option('audit-file')),
+    '-DSHOPT_HISTEXPAND',
```
krader1961: Isn't this unnecessary with the above changes?

krader1961:
LGTM modulo the seemingly unnecessary meson.build change.

src/cmd/ksh93/meson.build
siteshwar: Fixed.

PR #511, opened 2018-05-07
Fix pathdev() handling of symlinks

### LINKED ISSUE #492: Something is wrong with `pathdev` and thus `pathcanon()`
I started debugging the `io.sh` unit test failures on macOS/BSD. On Linux the `io_usename()` function returns a path like `/tmp/ksh.io.RpZsmVl/.<#d_12648{;.tmp` when the CWD is `/tmp/ksh.io.RpZsmVl`. On macOS/BSD it returns `vOOD/.<#d_51261{;.tmp` when the CWD is `/var/folders/6t/t624jkmx7cbb9t35vzvjwj180000gn/T/ksh.io.XXXXXXX.7GlrvOOD`. Notice that the path is relative and only contains the last four characters of CWD. That's why the `>; not updating file on success` occurs. The temp file it tries to create is invalid. This is doubly annoying because the diagnostic message is produced by this statement:

```
errormsg(SH_DICT, ERROR_system(1), ((o_mode & O_CREAT) ? e_create : e_open), fname);
```

Why is that annoying? It's because the syscall that failed used the var `tname`, not `fname`. Which makes the error message extremely misleading and why the problem wasn't obvious until I added debug prints to the `sh_redirect()` function.

krader1961:
This fixes the io.sh unit test on any platform which has two consecutive
symlinks in the temp directory pathname that points to a path longer than
the symlink. It also introduces a regression test which fails on Linux
(where /tmp isn't normally a symlink) without this change.

Note that I do not claim this change is 100% correct. The original code
is borderline impossible to understand. But this does fix the problem
and doesn't introduce any new unit test failures.

Partial fix for #483
Resolves #492

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -932,8 +932,10 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
     static char io_op[7];  // used for -x trace info
     int trunc = 0, clexec = 0, fn, traceon;
     int r, indx = shp->topfd, perm = -1;
-    char *tname = 0, *after = "", *trace = shp->st.trap[SH_DEBUGTRAP];
-    Namval_t *np = 0;
+    char *tname = NULL;
+    char *after = "";
+    char *trace = shp->st.trap[SH_DEBUGTRAP];
+    Namval_t *np = NULL;
     int isstring = shp->subshell ? (sfset(sfstdout, 0, 0) & SF_STRING) : 0;
     int herestring = (flag & IOHERESTRING);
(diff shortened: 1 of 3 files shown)


siteshwar:
@krader1961 Thanks for taking time to fix this. But I think it would be better to remove `pathdev()` code.

krader1961:
>  I think it would be better to remove pathdev() code.

I agree. I'm going to merge this anyway since we need to allow at least a couple of weeks to discuss that proposal. In the meantime this reduces test failures on macOS from 11 to 9 and fixes a bug that is visible on every OS.

Issue #512, opened 2018-05-07
Investigate sigchld test failures 


siteshwar:
`sigchld` test sometimes fails on travis with this message:

```
<E> sigchld[56]: test loop failed
<E> sigchld[60]: expected '10 foreground' -- got '' (DELAY=0)
<E> sigchld[61]: expected '2 background' -- got '' (DELAY=0)
```

PR #513, opened 2018-05-07
Use non-zero delay in sigchld test

siteshwar:
This test was specified as time sensitive in the legacy test script and
should be run with a non-zero delay. Since we have changed how test
cases are run, '$1' now contains test name. We should always use the
default timeout of 0.2 seconds in this script.

Resolves: #512


krader1961:
LGTM

Issue #350, opened 2018-01-11
Enable libast unit tests


krader1961:
Since we're stuck for the foreseeable future using subsystems like SFIO we should probably enable running their unit tests.

siteshwar:
+1. We should also enhance current test suite and write unit tests for other components of libast that we will continue to use.

fpmurphy:
@krader1961.  You most certainly should such be running the existing libast tests.  To do otherwise would demonstrate gross ineptitude.

krader1961:
@fpmurphy, Does your comment indicate you will be providing fixes so those tests pass? Because at the moment a huge number of tests fail on the `2016-01-10-beta` branch; i.e., before we started all the recent cleanup activity and changed the build system.

PR #514, opened 2018-05-07
Enable libast unit tests

siteshwar:
I have enabled most of the unit tests and left TODO note for the rest.

Resolves: #350


siteshwar:
[tmmap2read.c](https://github.com/att/ast/blob/6229ec5797f6720a7032b0722d05a8f57db1aa7d/src/cmd/tests/sfio/tmmap2read.c) test case fails to compile on opensuse due to this error:

```
/source/build/../src/cmd/tests/sfio/../terror.h:454: undefined reference to `___mmap'
/source/build/../src/cmd/tests/sfio/../terror.h:462: undefined reference to `___mmap'
```

It has this comment on top of it:

```

/*	This test causes mmap() to fail so that read() must be used.
        On a system such as BSDI, malloc uses mmap() so if mmap()
        fails, not much else will work. In such a case, we make this
        test automatically success.
*/
```

@krader1961 Is it something that should be tested on newer systems ?

krader1961:
Yes, I think that comment should be removed since `mmap()` should be available on every system we care about. I would leave the test module in the project but remove it from the list of tests that are run. Perhaps add a comment at the top of the module explaining that we're not running it because we expect a usable `mmap()` to be present.

kernigh:
tmmap2read.c doesn't compile unless the compiler optimizes away the unused tstshared() function in `src/cmd/tests/terror.h`. tstshared() calls mmap(), but the test changes mmap to ___mmap in all header files, so tstshared() calls ___mmap(), but ___mmap() doesn't exist.

siteshwar:
@krader1961 This is a test for `sfio` API and it delibrately causes `mmap()` to fail to check if fallback to `mmap` (which according to the comment is `read`) continues to work. But I will disable this test for now and leave a `TODO` comment for it.

krader1961:
LGTM

Issue #516, opened 2018-05-08
Which Linux distros should be tested in CI ?


siteshwar:
We have already setup CI for fedora and opensuse on Travis. I am thinking about enabling Travis builds for Ubuntu and Debian too. What other Linux distros should be tested in our CI ?

krader1961:
It's not clear there is enough value in testing multiple mainstream Linux distros to justify adding more than one or more two more (e.g., Ubuntu). And beyond Linux we should be testing on macOS and either FreeBSD or OpenBSD.

siteshwar:
I have not enabled builds for Debian yet, as `signal` tests are failing on it. Possibly due to a use after free issue:

```
/tmp/ksh.signal.S2Qottz/sigtst1: line 115: Â·Ã²: not found
/tmp/ksh.signal.S2Qottz/sigtst1: line 114: Ã—Â”z6: not found
/source/build/src/cmd/ksh93/ksh[2]: exec: kill: not found
<E> signal[148]: subshell ignoring signal does not send signal to parent
<E> expect: |done SIGUSR1|
<E> actual: |done should not get here|
<E> signal[149]: parent unexpectedly consumed fifo go signal
/source/build/src/cmd/ksh93/ksh[2]: exec: kill: not found
<E> signal[159]: subshell catching signal does not send signal to parent
<E> expect: |done SIGUSR1|
<E> actual: |done should not get here|
<E> signal[160]: parent unexpectedly consumed fifo go signal
/tmp/ksh.signal.S2Qottz/signal.sh: line 424: 27526: User signal 2
```

siteshwar:
Actually it seems `procps` package is missing in my debian builds.

siteshwar:
After #519 is merged, we should be testing on Ubuntu and Debian already.

On a second thought, one of my long pending issues is to set up OBS builds for different distros. It's very likely that I will run test cases on all package builds. That should take care of most of the distros.

krader1961:
@siteshwar It seems like this can be closed as we're now testing on several mainstream Linux based distros. Yes, we could add more but if we do so it should be something bleeding edge like [NixOS](http://nixos.org/). But it's not obvious that even that adds value since it is still fundamentally based on the Linux kernel and GNU user space. The project would be better served figuring out how to do CI testing on a non-macOS BSD, Cygwin, and a non-UNIX OS like Haiku.

siteshwar:
[...]

PR #519, opened 2018-05-09
Enable travis builds for Debian

siteshwar:
Related: #516

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,10 @@ env:
         - OS_TYPE=ubuntu
           INSTALL_REQUIREMENTS="apt-get update; apt-get install -y gcc python3 sudo locales ed vim meson; locale-gen en_US.UTF-8"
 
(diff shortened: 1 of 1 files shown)


krader1961:
LGTM.

Pro Tip: Instead of

```
sed -i 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/'
```

you can do the following to avoid having to repeat the text you want to keep:

```
sed -i '/# en_US.UTF-8 UTF-8/s/..//'
```

siteshwar:
Travis is having some performance issues and a specifiec API test is failing due to timeouts, despite of increasing timeout value. I am going to ignore this failure.

Merging with pro tip.

siteshwar:
@krader1961 I think you have introduced a side effect with bb57b0b012a965a07f4099ccdffab1f182968356 that is causing `tsignal.c` tests to fail.

Issue #348, opened 2018-01-11
What to do about builtins `mkservice` and `eloop`?


krader1961:
Builtins `mkservice` and `eloop` are commented out of *src/cmd/ksh93/include/builtins.h* and *src/cmd/ksh93/data/builtins.c* (i.e., hidden behind `#if 0`). There is no mention of them in the existing change log files. What do we want to do with these commands? Should they be enabled or removed?

dannyweldon:
Has anyone subscribed to this repo tested these to see if they are working?

I think it would be useful to be able to run services in `ksh`.

catull:
I have selected to "Watch" this repo.

The notifications I receive are for commits and issue activities, such as a new issue opened, an issue closing or comments added to issues.

Works for me.

krader1961:
> Works for me.

Huh? The `mkservice` and `epoll` commands are hidden inside `#if 0...#endif` blocks. They effectively don't exist:

```
$ mkservice
ksh: mkservice: not found [No such file or directory]
```

> Has anyone subscribed to this repo tested these to see if they are working?

Seems unlikely other than a member of the AT&T that worked on the project. The embedded documentation is below. The ID string implies this was implemented in 2001 but there is no mention of it in any of the RELEASE files.

```
[-?\n@(#)$Id: mkservice (AT&T Research) 2001-06-13 $\n]" USAGE_LICENSE
[+NAME? mkservice - create a shell server ]
[+DESCRIPTION?\bmkservice\b creates a tcp or udp server that is
implemented by shell functions.]
[+?The \aservice_path\a must be of the form \b/dev/tcp/localhost/\b\aportno\a
or \b/dev/udp/localhost/\b\aportno\a depending on whether the
\btcp\b or \budp\b protocol is used.  \aportno\a is the port
number that the service will use.]
[+?The shell variable \avarname\a is associated with the service.  This
variable can have subvariables that keeps the state of all
active connections.  The functions \avarname\a\b.accept\b,
\avarname\a\b.action\b and \avarname\a\b.close\b implement the
service as follows:]{
[+accept?This function is invoked when a client tries to connect
to the service.  It is called with an argument which
is the file descriptor number associated with the
accepted connection.  If the function returns a non-zero
value, this connection will be closed.]
[+action?This function is invoked when there is data waiting
to be read from one of the active connections.  It is
called with the file descriptor number that has data
to be read.  If the function returns a non-zero
value, this connection will be closed.]
[...]

PR #523, opened 2018-05-09
Remove code related to mkservice and eloop builtins

siteshwar:
These builtins were not shipped with the last stable release, so we
should avoid building them.

Resolves: #348


krader1961:
LGTM. Thx.

PR #524, opened 2018-05-10
Don't declare strlcat() unconditionally

krader1961:
Recent changes cause the unconditional declarations of `sprintf()`,
`strlcat()` and `strlcpy()` to be used (in a few files) after the relevant
system header that declares them has been imported. On macOS this causes
compilation errors due to differing declarations. So check if the system
provides those functions and only declare them if it doesn't.


siteshwar:
LGTM

krader1961:
Merged as 474395820.

Issue #227, opened 2017-12-12
bracket test fails due to update to Trusty images in Travis


siteshwar:
Below test is consistently failing on Travis since it has updated Trusty images :
```
	bracket.sh[232]: test -N $tmp/*: st_mtime>st_atime after read
```

afaik it has nothinig to do with the recent changes I have merged into master. All the test cases are passing in docker under Fedora.

krader1961:
I assume you're aware there are a large number of test failures on various platforms. Including the ones I have been using to verify changes such as macOS 10.13, FreeBSD 11.1, and Ubuntu 16.10. I have simply been verifying that a particular change does not increase the number of failures of those platforms. Without regard to whether the change simply replaced one failure with a different failure. My goal is to reformat the code and address compiler and lint tool warnings before spending more time on the failures seen on a specific platform. I did not want to expend energy investigating test failures before that cleanup was completed.

siteshwar:
Yes, I am aware about the test failures. I want to get rid of all the ```iffe``` tests before diving into bugs.

krader1961:
> Yes, I am aware about the test failures. I want to get rid of all the iffe tests before diving into bugs.

My comment was rhetorical but I'm happy to see we are on the same page. Being focused on eliminating a lot of legacy baggage that is no longer relevant is a good thing IMHO. Primarily in the hope that doing so will make it easier to fix bugs and add new features. What I don't understand is the perspective of some commenters (e.g., "jelmd") that modernizing the code is unacceptable.

DavidMorano:
@krader1961 

> What I don't understand is the perspective of some commenters (e.g., "jelmd") that 
> modernizing the code is unacceptable. 

I am only guessing here, but there may be people who have written built-ins which use 
some (or more) functionality from either LIBSHELL or LIBAST (or even some other AST libs)
which might get either broken or missing due to the clean-up efforts.

In a dream world, I think everyone would welcome cleaner and more maintainable code,
but perhaps people fear that some old (and maybe occur) function might get compromised.

Just a guess here (I could be wrong).

siteshwar:
With latest Ubuntu images, file access timestamps are not updated on reads and remounting all partitions with `mount -o remount,atime,strictatime -a` does not fix this failure.

siteshwar:
[...]

PR #526, opened 2018-05-10
Fix test failures on latest Ubuntu image

siteshwar:
Latest Ubuntu images are configured to not update atime on reads. This
is causing one of the brackets tests to fail. Since this is a
recommended performance optimization, we should do a feature test before
running this test.

Resolves: #227


krader1961:
Nack. As I mentioned in the issue these tests are important to ensure the implementation is using the correct `struct stat` fields. It would be better to just replace the implicit timestamp updates due to read/write activity with explicit modification using the `touch` command.

siteshwar:
@krader1961 I did not disable the test unconditionally. I am running it only when file system supports updating `atime` on reads. But I will try to rewrite the test as you have asked.

krader1961:
I know but it's now fairly common for filesystems to not update the access time of a file that is read so it would be better to simply not rely on it. Also, by using `touch` to explicitly set the timestamp we can use fairly clear values like `touch -a -t 01020304.06 the-file` which might be useful if the test were to display the values it found. It also means we don't need to sleep or use other hacks to ensure the timestamps differ.

siteshwar:
> I know but it's now fairly common for filesystems to not update the access time of a file that is read so it would be better to simply not rely on it.

Feature test depends on it, so in the worst case the test will not be run on such systems.

krader1961:
@siteshwar I am confused by your update to this PR after my previous comment. You replaced the `read` with `touch` but still retained the feature test for whether reads modify the atime of the file. It should not be necessary for this type of test to rely on whether or not the filesystem updates the access time of a file when it is read. The whole point of the test is to verify that ksh correctly executes the `stat()` (or related) syscall and correctly uses the values in the returned structure. The test is not about validating the filesystem correctly updates the access time when the file is read. That just happened to be true 98% of the time a couple of decades ago when this test was written. And the percentage of the time that it is true since then has dropped. Hence the need for this pull-request.

The above is a long winded way of saying I don't understand why there is still a feature test for whether the filesystem updates the atime of a file when it is read. Especially since a given system may have multiple filesystems. Some of which update the atime and some which don't.

Issue #528, opened 2018-05-10
Remove poll builtin


siteshwar:
This builtin was added after `ksh93u+` release:

```
12-08-21  The poll builtin from Roland Mainz has been added with the option                                           
      SHOPT_POLL.
```

I will prefer not to include it in the next release.

krader1961:
I'll admit my first reaction was to say "hold on, this might be useful." Then I read the documentation embedded in *src/cmd/ksh93/bltins/poll.c*. Gah! It's just exposing the low-level `poll()` syscall with no attempt to provide a higher level abstraction more appropriate for a shell script. Furthermore, I can't imagine anyone writing the type of application that needs `poll()` (e.g., a web server) as a ksh script. Other than as a "hey, look what I managed to do" type of experiment. So, yes, let's ditch this.

PR #529, opened 2018-05-10
Remove poll builtin and associated test

siteshwar:
This builtin was added after the last stable release and should not be
added to the next release.

Resolves: #528


KeithBierman:
Humm, Roland was the OpenSolaris guy for ksh93 for several years. I can
appreciate that he may no longer be working on ksh; but did anyone check
with him (or the opensolaris community?) about *poll* usage?



Keith Bierman
khbkhb@gmail.com
303 997 2749

siteshwar:
@KeithBierman If it is used on OpenSolaris, it should be fairly trivial to get it back through a downstream patch.

DavidMorano:
@siteshwar 
> If it is used on OpenSolaris, ...

I checked Illumos, it is certainly well integrated into Illumos. It is there in the their library sources, under the 'libshell/common/bltins' directory within the Illumos library source tree. It is about nine years now that it has been in that source tree. How much it is being used within Illumos, I do not know. Like Solaris proper and OpenSolaris afterwards, it looks like KSH forms an integral part of the Solaris-like descendants lives, and continuing w/ Illumos and its distros. Solaris made KSH its default shell for everything, including boot-up scripts.  It also looks like it uses the built-ins extensively. It looks like OpenSolaris and Illumos have continued with all of that.

I think that Illumos, and the distros built off of it (OpenIndiana and a few others), have a vested interest in maintaining KSH as much as they need to, being as KSH is so integrated into them.  If the POLL builtin gets dumped and either Illumos or one of its distros ever need to update, I think they can take care of themselves without the help of any new KSH release. Just a guess on my part. So I would not necessarily hesitate to drop POLL if you would like to.

Illumos and its distros do not treat KSH like a program or a command, they treat it like a basic system library component that is used by what otherwise appears to be the KSH and SH shells as well as the various built-ins that may also run as standard commands (otherwise programs). I am not really sure how many of the standard commands are actually secretly KSH built-ins, but the frame-work for all of that seems to be in place within the Illumos base system.

krader1961:
@DavidMorano How did you check? Because the ksh93u+ release did not have a `poll` builtin. Not even hidden behind an optional build-time `#define` symbol. So either you're just picking up on the fact that ksh uses the `poll()` syscall internally __or__ the Solaris derivatives have been shipping the beta version. Which, given all the bugs we've already identified in the code added after the ksh93u+ release, is extremely scary.

DavidMorano:
@siteshwar 
> How did you check? Because the ksh93u+ release did not have a poll builtin.

It is in the Illumos source code tree on GitHub. Hold a moment and I will try to get the URL to it on at GitHub. 
...
Got it (or have something):
https://github.com/illumos/illumos-gate/blob/master/usr/src/lib/libshell/common/bltins/poll_solaris.c

> ... or the Solaris derivatives have been shipping the beta version. 
> Which, given all the bugs we've already identified in the code added 
> after the ksh93u+ release, is extremely scary.

Yes, something is up (strange), because Solaris did all of this KSH integration like at least 10+ years ago, or even before. I am not sure, but it might even be 15+ years ago when Solaris first decided to do a major new KSH integration. From the date on the Illumos version of the POLL builtin, it looks like it was added almost exactly nine years ago (this month).

I do remember talk about adding a POLL builtin and that was at least 10+ years ago (if I remember correctly). It looks like Roland Mainz wrote it way back then. If it was not in recent KSH source, somehow it was either previously there and dropped, or maybe it was not there and only added there using the Solaris version as the source for it (I have not checked on this though).

Anyway, YES, with all of the bugs  you and others have found, it does seem a bit scary how much the original Solaris and subsequent child OSes (and child distros) have relied on KSH. Somehow KSH must be working well enough in that environment so as not to have caused any major problems for them in this last decode or more. And the kicker is, that Solaris and its descendants tend to be in the large-enterprise space, which is generally a lot less tolerant of bugs than the common desktop or regular server spaces. 

For what it is worth  (if anything), it looks like (possibly) that the new center of gravity for Solaris-like descendants is now in Europe rather than in America, but that may just be some superficial appearance on my part.

DavidMorano:
@krader1961 
> Notice the name of that file in the Illumos source link you provided: poll_solaris.c.

This did not escape my notice. :-)
Yes, That is why I mentioned "... only added there [AST code base] using the Solaris version as the source for it "

> This is a variation of the "If all you have is a hammer then every problem looks like a nail" scenario.

Yes, I think that is exactly what they (the original Solaris people, Sun Microsystems at the time) were going for. They were previously relying heavily on the older Bourne shell, as well as some KSH, and they wanted to eliminate the Bourne shell altogether. So they chose KSH as being their POSIX 1003 replacement. Of course, KSH serves as itself also, hence their consolidation effort (so many years ago now).

I remember reading both the AST mailing lists and some Solaris mailing lists at the time; the Sun people were very happy at the time to update many of their system shell scripts using the many newer features of KSH, even though they did not strictly have to (since KSH is POSIX 1003 compatible like the old Bourne shell). It was their decision to make (consolidating on KSH), and (for me) I think eliminating the old Bourne shell from their system was not a bad decision. Also, Solaris used to be very sensitive to trying to follow many -- or almost all -- of the major standards. This probably influenced their decision, seeing as they felt they had to continue to provide a POSIX 1003 shell anyway for their customers.

PR #532, opened 2018-05-13
Fix test failures on latest Ubuntu image

siteshwar:
It is a recommended filesystem optimization to disable atime updates on
reads. One of the bracket tests breaks due to this optimization. To
workaround this, we should use 'touch -a` to mimic the file was read.

Resolves: #227

### DIFF
--- a/src/cmd/ksh93/tests/bracket.sh
+++ b/src/cmd/ksh93/tests/bracket.sh
@@ -305,15 +305,14 @@ done
 ) || log_error 'errors with {..}(...) patterns'
 [[ D290.2003.02.16.temp == D290.+(2003.02.16).temp* ]] || log_error 'pattern match bug with +(...)'
 rm -rf $file
-{
+touch -t 01020304.05 $file
 [[ -N $file ]] && log_error 'test -N $TEST_DIR/*: st_mtime>st_atime after creat'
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/bracket.sh
```diff
@@ -305,15 +305,15 @@ done
 ) || log_error 'errors with {..}(...) patterns'
 [[ D290.2003.02.16.temp == D290.+(2003.02.16).temp* ]] || log_error 'pattern match bug with +(...)'
 rm -rf $file
-{
+touch $file
 [[ -N $file ]] && log_error 'test -N $TEST_DIR/*: st_mtime>st_atime after creat'
 sleep 2
```
krader1961: Please either change this to `sleep 1.01` (since we should be able to count on at least second resolution of file timestamps). Even better would be to replace the first touch with an explicit timestamp: `touch -t 01020304.05 $file`. Then replace this sleep and the next print with `touch -m -t 01020708.09 $file`. Same thing a few lines below. That will speed up this test by four seconds.

siteshwar: Fixed.

krader1961:
LGTM

Issue #21, opened 2016-08-01
Missing license


mperry2:
The software is missing a license. Please add one by committing a file to the root of your repo titled LICENSE. See https://github.com/blog/1530-choosing-an-open-source-license if you need help choosing a license.

ormaaj:
The license is EPL 1.0

Yes, AT&T made a weird license choice (again). The EPL is particularly ambiguous about the reusability of "modules", and this is a project that consists largely of libraries.

I would have liked to use AST's regex code but avoided this because I couldn't guarantee license compatibility everywhere I might have wanted it.

dannyweldon:
It is a little tricky because this repo contains many independent packages, which could each technically have their own license, should AT&T decide to do so I suppose, and the license is stated at the top of every source file, though I have only checked a few.

Perhaps a LICENSE file could be created indicating this information rather than just stating EPL 1.0?  Though it makes more sense to me to have a single LICENSE file and have the source files point to that location rather than all of them declare what the license actually is, to make it easier to change.

krader1961:
The [Fish](https://github.com/fish-shell/fish-shell) project does not have a LICENSE file. The [Elvish](https://github.com/elves/elvish) project does. Those are the other shell projects I've contributed to. I am not a lawyer so I have no idea what the right thing to do here is. We can certainly, and probably should, create a *LICENSE* file that describes which portions of the code are under which licenses.

This may be difficult because parts of the code (e.g., *src/libbz/crctable.c*) are from other projects without an explicit license mentioned in the file but which do have a *LICENSE* file. While other pieces of code put the license in each source file. And, of course, there may be other variations I haven't noticed.

kernigh:
These files have different licenses:

- src/lib/libast/misc/fastfind.c has a BSD license
- src/cmd/ksh93/tests/treemove.sh mixes EPL 1.0 with CDDL
- perhaps other files that I haven't found yet

siteshwar:
Although there are some portions of code which use different licenses, this project as a whole was released under EPL-1.0. I think it should be safe to put EPL-1.0 text in the LICENSE file.

PR #534, opened 2018-05-15
Add LICENSE file with text from EPL-1.0

siteshwar:
License text has been taken from
https://www.eclipse.org/legal/epl-v10.html

Resolves: #21


krader1961:
IANAL but LGTM

Issue #327, opened 2018-01-05
Setup Coverity Scan for this project


krader1961:
The [Coverity Scan](https://scan.coverity.com/) static analysis service is invaluable for identifying problems with code. We should enroll this project with that service.

siteshwar:
We have been using coverity within Red Hat to scan ksh package builds. I agree that we should use it to scan upstream code too.

catull:
How frequent do you want to run it ?
Is it a manual task, automated ?

If automated, is it run monthly, weekly ?
If it scans on each merge/commit, it would create an entire report that is likely not read.

I support the scanning, it is miles better than relying on clang / gcc to issue warnings.

siteshwar:
Our `.travis.yml` file already started to look strange, so I am going to avoid mixing up configurations for coverity builds into it. Since there are API limits in coverity and it is recommened by coverity documentation to use separate branch to trigger builds, I am going to write a separate `.travis.yml` file for it. This file can be copied to `.travis.yml` file each time we want to trigger coverity scan.

siteshwar:
For the record, coverity builds do not work with macOS due to this error:

```
[ERROR] System Integrity Protection is enabled. In order to use cov-build,
this must be disabled. If you are confident that SIP has been
disabled, pass --no-sip-check to bypass this check.
```

PR #538, opened 2018-05-17
Add reference to coverity scan builds in README.md

siteshwar:
Resolves: #327


bin/trigger-coverity-scan.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Get current branch name
+current=$(git rev-parse --abbrev-ref HEAD | tr -d "\n")
+# Coverity scan is triggered only for coverity_scan branch
+git branch -D coverity_scan
```
krader1961: I'm not a fan of redirecting stderr to /dev/null but for this specific command it's justified since the branch may not exist which is harmless. No sense showing an error message to the user.

krader1961:
LGTM

PR #539, opened 2018-05-17
Stop creating the conftab files in the source tree

krader1961:
The legacy (IFFE based) build system generated several files in the
source tree. That is not a best practice since it requires, among other
things, that the VCS be configured to ignore those dynamically generated
files. This practice was carried forward when the project switched to
the Meson build tool out of expediency. This change takes one step
towards fixing that undesirable practice by moving the generated conftab
source files into the build tree.

This change slightly simplifies the conf.sh script. It also introduces a
new helper script to convert a template file into a target file using
Python. Since we already depend on Python indirectly via our dependency
on Meson this isn't introducing a new dependency. Using it to translate
a template into a concrete file is preferable to the alternatives such
as using shell constructs like

    eval `cat some_file`

I don't think I need to explain why `eval` in that context is dangerous
and fragile. Similarly using `sed` is fragile because substitutions that
have embedded newlines require escaping the newline and potentially
other characters.

It also introduces a helper script to sort the unknown C preprocessor
symbols to deal with a bug in the macOS `grep` command.


krader1961:
The bug in the macOS `grep` command I alluded to in the commit comment is as follows. I've already reported this to Apple. Create a file named *p* with these lines:

```
PID_MAX
SYSPID_MAX
```

Create a file named `x` with these lines:

```
SYSPID_MAX
PID_MAX
```

Run `fgrep -x -v -f p x`. The expected result is no output. The actual output is the line `SYSPID_MAX`. GNU and FreeBSD grep do the right thing. Changing the order of the lines in file "p" also results in correct behavior.

bin/libast_prereq.sh
```diff
@@ -1,26 +1,30 @@
 #!/bin/sh
-# This script is used for feature detection and setting up other
-# configurations that are required to build libast.
+# This script is does not due feature detection in the usual sense. In fact it
```
siteshwar: I did not understand this line.

siteshwar:
LGTM

PR #540, opened 2018-05-18
Make dump_backtrace() more useful

### LINKED ISSUE #460: GNU `dladdr()` doesn't work for addresses in the base text segment

krader1961:
This causes more function names to be visible to the `dladdr()` function
on GNU Linux and other systems.

Fixes #460 

P.S., The changes I made to *nvinit.c* are to demonstrate what needs to be done to all the code for this to be truly useful. If this is deemed acceptable I'll make the same mechanical transformation to the other ksh modules.

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -25,6 +25,7 @@
 // These are atypical platform symbols with respect to how they are
 // determined. They are either hardcoded or not determined via the usual
 // platform feature detection and naming conventions.
+#mesondefine DEBUG_BUILD
 #mesondefine MESON_BUILD
 #mesondefine HOSTTYPE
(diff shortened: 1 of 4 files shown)


siteshwar:
LGTM

PR #542, opened 2018-05-18
Add template for bug report

siteshwar:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+name: Bug report
+about: Short description of bug
+
+---
+
+**Description of problem:**
+
+
+**Version-Release number:**
+(If built from Git, output of `git rev-parse HEAD`)
```
krader1961: I would drop this. We should be including the git commit number in the ksh version string like most projects do. That way all we have to do is tell the user to type `ksh --version`.

krader1961: And perhaps change the section heading to something like `Ksh version:`.

krader1961:
LGTM

PR #544, opened 2018-05-20
Generate the version number from git state

### LINKED ISSUE #543: non-release ksh builds should include the git commit ID in the version string
It should be sufficient to ask the user to run `ksh --version` or `print $KSH_VERSION` to tell us if they built ksh from source from a non-release commit and which commit they used.

krader1961:
With this change the ksh version string is based on git state.
Specifically the most recent annotated tag and any subsequent changes.

At the time I made this commit the dynamically generated version string
looked like this:

Which means my git workspace was 914 commits past the commit tagged
"93v", the commit id was "03c799ab", and I had uncommitted changes.

Resolves #543


siteshwar:
Please also remove this line from `.copr/prep.sh`:
```
sed -i "s/#define SH_RELEASE.*/#define SH_RELEASE \"${COMMIT_DATE}+git.${COMMIT_NUM}.${COMMIT_SHORT}\"/" src/cmd/ksh93/include/version.h
```

siteshwar:
I am putting the differences here for comparision:

With last stable version:

```
$ ksh --version
  version         sh (AT&T Research) 93u+ 2012-08-01
[root@qeos-23 ~]# ksh
$ echo $KSH_VERSION
Version AJM 93u+ 2012-08-01
$ echo ${.sh.version}
Version AJM 93u+ 2012-08-01
```


With changes in this pull request:

```
$ src/cmd/ksh93/ksh --version
  version         sh (AT&T Research) 93v-924-g67744577
$ echo $KSH_VERSION
Version ABIJ 93v-924-g67744577
$ echo ${.sh.version}
Version ABIJ 93v-924-g67744577
```

@krader1961
I presume the changes in `$SH_VERSION` and `${.sh.version}` variables are only temporary and we will restore them back to values in last stable release once we agree on new versioning scheme.

saper:
Can we do this in a way that `git` executable is not mandatory for building (i.e. the commit number can be provided externally)?

krader1961:
> I presume the changes in $SH_VERSION and ${.sh.version} variables are only temporary...

Yes (assuming you meant `$KSH_VERSION`). Whether or not we reinstate the date stamp in the version string is TBD. Personally, I'm not a fan of including that datum in the version string. Too, once we switch to a semantic version number the `nget_version()` function should return that as a normalized int (e.g., 93.1.2 => 930102) rather than converting a date stamp like 2012-01-02 to 20120102. Note that if backward compatibility is a concern the easiest solution is to make the release major number the year rather than 93 (e.g., 2018.0.0). Then a test like `if (( .sh.version > 20120801))` to test if the ksh version is newer than the ksh93u release will still work.

> Can we do this in a way that git executable is not mandatory for building (i.e. the commit number can be provided externally)?

Sure. Probably the simplest is something like the following script installed as `git` in `$PATH`:

```
#!/bin/sh
if test "$1" = describe
then
    echo 2018.0.0
    exit 0
fi
exit 1
```

I just tested that and it works. Whether we need a different mechanism that leverages Meson features (e.g., an entry in *meson_options.txt*) is an open question.

krader1961:
I'm going to merge this in a couple of hours when I get home. The discussion about whether or not we should continue to include a date stamp in addition to a semantic version ID can continue in #335. This change at least makes it crystal clear that anyone running a ksh built from this project is no longer running 93v released four years ago. We might want to add a "2018.0.0wip" tag or something similar to one of the commits made last year when active development resumed.

siteshwar:
@saper https://github.com/att/ast/pull/571

PR #545, opened 2018-05-22
Reinstate the ksh93u `pathcanon()` implementation

krader1961:
This also removes three unit tests that depend on the simulated 3DFS
behavior provided by `pathdev()`. We'll remove `pathdev()` in  separate
commit.

Partial fix for #510

PR #547, opened 2018-05-22
Restore definition of subshell_table_unset

siteshwar:
This function was causing lot of use after free crashes, so I have
restored it back to it's definition from ksh93u+ release.

Related: #398

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -2200,16 +2200,7 @@ void _nv_unset(Namval_t *np, int flags) {
                 }
                 dtclose(rp->sdict);
             }
-            if (flags & NV_TABLE) {
-                // Drop all references and close the stream.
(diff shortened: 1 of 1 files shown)


krader1961:
`bin/style` please; otherwise LGTM.

P.S., This is basically what I'm doing for issue #545 which will involve reinstating the ksh93u `sh_open()` implementation.

krader1961:
I'm surprised that `meson test --setup=malloc` on macOS has the same number of test failures with and without this change. Shouldn't I have seen a decrease in test failures?

siteshwar:
It decreases number of failures from `44` to `26` on my fedora machine.

krader1961:
I tested on Ubuntu 16.10 and with `--setup=malloc` failures went from 36 to 17 so this is most definitely a huge improvement. TBD is why things are so much worse on macOS.

Issue #546, opened 2018-05-22
Test failures on Windows WSL and Cygwin


krader1961:
This issue is similar to #483 for OpenBSD but is instead for ksh on Windows Subsystem for Linux (WSL, aka "bash on windows") and Cygwin.

I booted Windows 10 today and updated that OS and my Cygwin installation on it. I then installed Meson and Ninja in WSL (aka “bash on windows”) and Cygwin and built and unit tested ksh. On WSL there are 7 test failures. Mostly involving signal handling which isn’t too surprising given the imperfect emulation of UNIX by WSL. There were 71 failures under Cygwin. This is the first time I’ve built/tested ksh on the two Windows platforms we need to support. Obviously there is some work to be done by someone who cares about ksh in those two Windows environments.

Note that the libast API timeouts noted [here](https://github.com/att/ast/issues/483#issuecomment-389302300) on OpenBSD also occur under Cygwin but not WSL and thus require running the unit tests with a large timeout multiplier (e.g., `meson test -t5`) on Cygwin to avoid a lot of false positive test failures.

krader1961:
I took another look at the test failures on WSL. Three of them are due to the `arith` unit test failing the equality comparison of floating point values.

```
<E> arith[712]: 4.**3/10!=6.4
<E> arith[960]: mean of index array  is 5.11999999999999922 should be 5.12
<E> arith[961]: mean of associative array  is 2 should be 5.12
<E> arith[964]: dotprod of two index arrays  is 22.1999999999999993 should be 22.2
<E> arith[965]: dotprod of two identical index arrays  is 163.599999999999994 should be 163.6
<E> arith[966]: dotprod of index and associative array  is 28.1999999999999993 should be 28.2
<E> arith[967]: dotprod of two identical associaive arrays  is 163.600000000000023 should be 163.6
<W> arith[-1]: error_count = 7
```

It's actually surprising this test doesn't fail more often on other platforms. Why? Because you can't calculate a floating point value and reliably compare it for equality to a floating point constant. You have to compute the absolute difference of the two values and test that it is less than some epsilon value.  Or, in this case, force the string representation to be rounded to the expected significant digits using `printf`.

krader1961:
With the PR I just sent out for review there are zero failures on WSL modulo the usual flakey failures.

krader1961:
[...]

PR #548, opened 2018-05-22
sched_yield rather than sleep

krader1961:
Several unit tests utilize very short duration sleeps (e.g., 100 usec)
while waiting for something to happen. The problem with this is that
on some platforms the minimum sleep interval might be 1, 10, even
100 msec. Which means the test can take one or two orders of magnitude
longer to run on those platforms compared to a platform where sleeps are
more granular. So instead of sleeping simply yield execution to another
runable thread.

This reduces the unit test run time for me from 49 minutes to 20 minutes
on OpenBSD in a virtual machine. And from 35 minutes to 16 minutes on
Cygwin.

Partial fix for #483
Partial fix for #546


krader1961:
Merged since it is obviously better on some platforms and doesn't introduce any issues on the platforms that were already running the affected tests quickly. The Travis failures are the usual flakey failures having nothing to do with the PR.

PR #549, opened 2018-05-23
Fix a use after free issue caused by _nv_unset function

siteshwar:
_nv_unset function runs stkclose() on a stack in a while loop until
it's reference count is less than one and it can free the stack. But
the same stack is getting free'd by sh_freeup() function later. This
may cause a crash due to memory corruption. In ksh93u+ release,
stakdelete() macro was used to close the stack. This macro calls
stkclose() function and it decreases reference count for the stack by
only one. It should be later free'd by sh_freeup() function.

Related: #398

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -2200,16 +2200,7 @@ void _nv_unset(Namval_t *np, int flags) {
                 }
                 dtclose(rp->sdict);
             }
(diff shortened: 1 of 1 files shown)


siteshwar:
It brings down test failures from `44` to `23` with `meson teset --setup=malloc` on my fedora machine.

krader1961:
LGTM

PR #550, opened 2018-05-25
Fix a use after free issue while unseting arrays

siteshwar:
There is a use after free issue that may occur while unseting elements
of an array. This patch works around it.

Related: #398

### DIFF
--- a/src/cmd/ksh93/sh/array.c
+++ b/src/cmd/ksh93/sh/array.c
@@ -501,7 +501,10 @@ static_fn void array_putval(Namval_t *np, const char *string, int flags, Namfun_
                     if (mp != np) {
                         array_clrbit(aq->bits, aq->cur, ARRAY_CHILD);
                         aq->val[aq->cur].cp = 0;
-                        if (!xfree) nv_delete(mp, ap->table, 0);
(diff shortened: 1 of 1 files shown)


siteshwar:
This brings down number of test failures with `meson test --setup-malloc` from `23` to `8` on my system.

krader1961:
LGTM.

For the benefit of others I'll mention a few things from the conversation @siteshwar and I had on the Gitter dev channel. Even if this change introduces a memory leak that is preferable to a use-after-free bug which can cause random data corruption or, if you're lucky, an immediate SIGSEGV fault. Eliminating the use-after-free bugs will make it possible to switch from Vmalloc (which is masking these bugs) to the system malloc. Thus eliminating a big chunk of code that has been the source of problems.

DavidMorano:
@krader1961 
> Eliminating the use-after-free bugs will make it possible to switch from Vmalloc (which is masking these bugs) to the system malloc. Thus eliminating a big chunk of code that has been the source of problems.

Thanks very much for keeping your eye on the ball -- namely, getting rid of VMALLOC!
Once KSH is on the system MALLOC, more memory related bugs can be found and eliminated.
Also, KSH will be legitimately on a thread-safe memory allocator, a necessity for thread-safe built-ins and background threads.

PR #551, opened 2018-05-25
Fix several use after free issues

siteshwar:
Related: #398

### DIFF
--- a/src/cmd/ksh93/sh/nvtree.c
+++ b/src/cmd/ksh93/sh/nvtree.c
@@ -975,7 +975,7 @@ static_fn char **genvalue(char **argv, const char *prefix, int n, struct Walk *w
 static_fn char *walk_tree(Namval_t *np, Namval_t *xp, int flags) {
     Shell_t *shp = sh_ptr(np);
     static Sfio_t *out;
-    struct Walk walk;
+    struct Walk walk = {0};
     Sfio_t *outfile;
     Sfoff_t off = 0;
     int len, savtop = stktell(shp->stk);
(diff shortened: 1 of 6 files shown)


src/cmd/ksh93/sh/nvtree.c
```diff
@@ -1021,12 +1021,13 @@ static_fn char *walk_tree(Namval_t *np, Namval_t *xp, int flags) {
     dir = nv_diropen(mp, name, (void *)shp);
     walk.root = shp->last_root ? shp->last_root : shp->var_tree;
     if (subscript) name[strlen(name) - 1] = 0;
-    while ((cp = nv_dirnext(dir))) {
-        if (cp[len] != '.') continue;
+    cp = nv_dirnext(dir);
+    while (cp) {
```
krader1961: I really dislike the `while ((x = func())` pattern so thanks for this change. But rather than do it this way and introduce a goto label how about this:

```
while (true) {
    cp = nv_dirnext(dir);
    if (!cp) break;
```

siteshwar: This will reintroduce the memory corruption. We have to get result of `nv_dirnext` before the pointer it tries to use gets released by call to `nv_delete`.

krader1961: Right, I hadn't scrolled down to see that there are other complications. But the attached diff (applied on top of your change) gets rid of the gotos and I think makes the relationship between `nv_dirnext()` and freeing the other vars slightly clearer. I've tested this on all my VMs and it appears to produce identical results.
[goto.txt](https://github.com/att/ast/files/2041284/goto.txt)

krader1961: Also, I just noticed that declaration `struct Walk walk;` does not initialize that structure. It looks like that is okay given the member assignments done later in the function but it bothers me. We've found far too many bugs due to using stack allocated data structures that are not fully initialized. Might be a good idea to add a `= {0};` to that declaration just out of paranoia.

siteshwar: Your patch is simpler so I have added it to the pull request.

krader1961:
LGTM

Issue #510, opened 2018-05-06
Should the post-ksh93u support for AST magic paths be retained or removed?


krader1961:
A new feature was introduced after the ksh93u+ release that is, AFAICT, only documented in the source (and cryptically at that). Primarily file *src/lib/libast/path/pathcanon.c*. This new code is the cause of regression issue #492. Below are some of the relevant source code comments. The question is whether we want to retain this new code as there is at least one significant bug, the `pathdev()` function is too long and hard to understand, and the feature is poorly documented. Issue #492 would be most easily fixed by simply reinstating the older `pathcanon()` implementation.

```
 * check for ast pseudo dev/attribute paths and optionally canonicalize
```

```
 * path name canonicalization -- preserves the logical view
 *
 *      remove redundant .'s and /'s
 *      move ..'s to the front
 *      /.. preserved (for pdu and newcastle hacks)
 *      if (flags&PATH_ABSOLUTE) then pwd prepended to relative paths
 *      if (flags&PATH_PHYSICAL) then symlinks resolved at each component
 *      if (flags&(PATH_DOTDOT|PATH_PHYSICAL)) then each .. checked for access
 *      if (flags&PATH_EXISTS) then path must exist at each component
 *      if (flags&PATH_VERIFIED(n)) then first n chars of path exist
 *
 * longer pathname possible if (flags&PATH_PHYSICAL) involved
 * 0 returned on error and if (flags&(PATH_DOTDOT|PATH_EXISTS)) then canon
 * will contain the components following the failure point
 *
 * pathcanon() and pathdev() return pointer to trailing 0 in canon
 * pathdev() handles ast specific /dev/ and /proc/ special files
 * pathdev(PATH_DEV) returns 0 if path is not a valid special file
 * pathdev(PATH_AUX) Pathdev_t.flags|=PATH_AUX if AUX fd was created
 * see pathopen() for the api that ties it all together
```

```
 * NAMED_XATTR
 *
 *      0:disabled
 *      1:enabled if O_XATTR!=0
 *
 * /dev/file/xattr@relative-path/file//@//?(attr)
 * /dev/file/xattr@/absolute-path/file//@//?(attr)
 *
 * names the xattr file "attr" for "file" or the xattr dir for "file"
 * neither "relative-path/file" nor "absolute-path/file" may contain "//@//"
 * so pathcanon() "relative-path/file" and "absolute-path/file" before
 * constructing a /dev/file/xattr@/...//@// path
```

krader1961:
[...]

PR #552, opened 2018-05-26
Remove `pathopen()`

krader1961:
If the magic "PATH_DEV" support that simulates features of 3DFS (3D file
system) is disabled there isn't any need for `pathopen()` so replace it
with `openat()`.

This removes one unit test that depends on the simulated 3DFS behavior
and which does not pass using ksh93u.

Part of resolving #510 

## Please Note

This change does not actually remove the `pathopen()` function. It merely makes it a synonym for `openat()`. In a subsequent change I will replace all the callers of `pathopen()` with `openat()` and actually remove this code. This just eliminates one of the `pathdev()` call sites on the road to removing that function.


siteshwar:
LGTM

krader1961:
Merged. Only thing left is to remove `pathdev()` and the couple of call sites but that will be a bit harder.

PR #554, opened 2018-05-29
Replace Vmalloc with the system malloc

### LINKED ISSUE #396: Replace AST Vmalloc with standard malloc

krader1961:
This temporarily disables the signal test until the bug that this
unmasked is resolved.

Mostly resolves #396


siteshwar:
Please also remove this line from `meson.build` file:

```
malloc_debug_env.set('VMALLOC_OPTIONS', 'junk')
```

siteshwar:
Remove `src/lib/libast/man/vmalloc.3` too.

siteshwar:
Interesting observation, a change from libast `RELEASE` file:

```
10-01-25 vmalloc/vmprivate.c: fix seg ptr initialization bug (24 years old!!)
```

Sort of shows age of vmalloc and why we are getting rid of it.

DavidMorano:
@siteshwar  wrote:
> Sort of shows age of vmalloc and why we are getting rid of it.

And, a little bit disappointingly (OK, very disappointingly), in all of those years it was never enhanced to be made:
+ (properly and tested) thread-safe
+ fork-safe

krader1961:
The only way to make it smaller is to do the actual delete of the Vmalloc files in a separate commit. However doing that wouldn't actually help future debugging because all of the relevant change (switching to system malloc calls) would still be in a single change. And, assuming I didn't introduce a typo when converting `newof()` calls to `calloc()` and the like, this change by definition should not introduce any regressions.

There is exactly one change that might introduce a bug. In *src/lib/libast/cdt/dtnew.c* I changed

```
static void *memoryf(Dt_t *dt, void *addr, size_t size, Dtdisc_t *disc) {
// This was the original Vmalloc based implementation:
//   return vmresize(((Dc_t *)disc)->vm, addr, size, VM_RSMOVE);
// That vmresize() call initializes the new bytes to zero which a simple realloc() won't.
return realloc(addr, size);
}
```

However, AFAICT the code that uses `memoryf()` doesn't actually depend on the Vmalloc behavior.

siteshwar:
> The only way to make it smaller is to do the actual delete of the Vmalloc files in a separate commit. However doing that wouldn't actually help future debugging because all of the relevant change (switching to system malloc calls) would still be in a single change. And, assuming I didn't introduce a typo when converting newof() calls to calloc() and the like, this change by definition should not introduce any regressions.

I leave the decision up to you, but in future please avoid making such big changes in a single commit.

krader1961:
Merged. I refactored it into two commits. The first replaces all the Vmalloc API calls with standard libc malloc calls. The second deletes all the Vmalloc source files.

kdudka:
On Tuesday, May 29, 2018 7:35:32 PM CEST Kurtis Rader wrote:
> The only way to make it smaller is to do the actual delete of the Vmalloc
> files in a separate commit.

I disagree.  Things like replacing:

    return 0;

.. with:

    return NULL;

.. could have been done in a separate commit.  Similarly with:

@@ -819,8 +819,9 @@ Namval_t *nv_create(const char *name, Dt_t *root, int 
flags, Namfun_t *dp) {
                 if (c) *sp = c;
                 top = 0;
                 if (np && !nv_isattr(np, NV_MINIMAL) && shp->oldnp && !np-
>nvenv &&
-                    shp->oldnp != np && !(flags & NV_ARRAY))
+                    shp->oldnp != np && !(flags & NV_ARRAY)) {
                     np->nvenv = (char *)shp->oldnp;
+                }
                 shp->oldnp = np;
                 if (isref) {
 #if NVCACHE

.. which is a coding style change not related to the actual switch at all 
(and yes, even such changes are known to cause regressions in ksh code from 
time to time).

Also the following change could have been safely applied on its own:

@@ -149,6 +149,7 @@ void *nv_diropen(Namval_t *np, const char *name, void 
*context) {
     dp = calloc(1, sizeof(struct nvdir) + len + 1);
     if (!dp) return NULL;
 
+    memset(&fake, 0, sizeof(fake));
     dp->data = (char *)(dp + 1);
     if (name[len - 1] == '*' || name[len - 1] == '@') len -= 1;
     name = memcpy(dp->data, name, len);

There are dozen changes like that in your mega commit.

siteshwar:
@extrowerk I remember you were having troubles compiling on haiku due to `vmalloc`. I would suggest you to try again as we have removed dependency over it.

extrowerk:
Thx, I will do.

PR #557, opened 2018-05-30
Fix svar_init()

### LINKED ISSUE #556: Bug in sh_setsiginfo() corrupts memory
I was testing my change to replace Vmalloc with the system malloc. That change works fine on my 64 bit systems. However, on Ubuntu 16.04 32 bit, the very first test in the `sigchld` unit test causes a SIGSEGV. That's because of this statement in `sh_setsiginfo()`:

```
strncpy(signame, p, SIGNAME_MAX);
```

The problem is that `signame` points to a buffer just 16 bytes before the structure pointed to by `SH_SIG->nvfun` (aka `(shgd->bltin_nodes + 66)->nvfun`). And because `SIGNAME_MAX` is 32 that `strncpy()` modifies the first four words of the `Namfun_t` structure that follows it. Which sets the `disc` member of that structure to NULL. Which leads to a SIGSEGV when it is dereferenced by this statement in `sh_setsiginfo()` the next time it runs:

```
while (fp->disc->createf != create_svar) fp = fp->next;
```

TBD is whether the target buffer pointed to by `signame` should be at least `SIGNAME_MAX` in length or this is a misuse of `strncpy()`.

krader1961:
This fixes the initialization of `struct Svars` by making the amount of
space allocated match the recorded `dsize`. Whether or not this is correct
is unclear because I don't understand the inclusion of `sizeof(int)`
in the calculation of `dsize`. But it does keep the `strncpy()` in
`sh_setsiginfo()` from modifying data past the end of the allocated buffer.

Fixes #556

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -1683,10 +1683,10 @@ static_fn Namfun_t *clone_svar(Namval_t *np, Namval_t *mp, int flags, Namfun_t *
     Sfdouble_t d;
 
     // cppcheck-suppress pointerSize
-    dp = malloc(fp->dsize + sp->dsize + sizeof(void **));
+    dp = malloc(fp->dsize + sp->dsize + sizeof(void *));
     memcpy(dp, sp, fp->dsize);
     dp->nodes = (char *)(dp + 1);
-    dp->data = (void *)((char *)dp + fp->dsize + sizeof(void **));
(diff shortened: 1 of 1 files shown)


krader1961:
FYI, It looks like the coprocess tests are failing at a higher rate after the switch to the system malloc. Which is not surprising given the nature of this fix.

siteshwar:
LGTM

PR #559, opened 2018-05-30
Fix unit test handling of va_list taking functions

### LINKED ISSUE #558: Failed to build on armv7l
**Description of problem:**
Build fails on armv7

**Ksh version:**
Git master 2018-05-29

Here is the build log
http://pkgsubmit.mageia.org/uploads/failure/cauldron/core/release/20180529213558.kekepower.duvel.47174/log/ksh-20180529_gitebe3b505-0.1.mga7/build.0.20180529213610.log

I don't know if it's related to the removal og vmalloc or not, but the build before this one worked fine.

Cheers.

krader1961:
The unit test incorrectly assumes that 0 == NULL == va_list. That
happens to be true on x86 but is not necessarily true on other
architectures like ARM.

Fixes #558

### DIFF
--- a/src/cmd/tests/sfio/tbadargs.c
+++ b/src/cmd/tests/sfio/tbadargs.c
@@ -21,7 +21,7 @@
 
 #include "sftest.h"
 
-int Code_line = 30; /* line number of CALL(sfclose(0)) */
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #560, opened 2018-05-30
Increase timeout for coprocess tests

siteshwar:
coprocess tests have started timing out after recent change of removing
vmalloc. We should increase timeout for these tests.


krader1961:
Nak. It doesn't matter how long the timeout is. I just ran `make -t4 coprocess/C` in a loop and it timed out on the fourth run. That was with a timeout of 200 seconds. One of the tests is deadlocking.

PR #562, opened 2018-05-31
Fix sigchld unit test on macOS

### LINKED ISSUE #561: The sigchld test for SIGCONT being reported is not portable
This loop in *src/cmd/ksh93/tests/sigchld.sh* assumes that the `wait4()` syscall will return when a process is stopped and receives a SIGCONT signal:

```
for stat in EXITED STOPPED CONTINUED EXITED
```

That is true for Linux as of kernel 2.6.0 (from the `waitpid` man page):

```
WCONTINUED (since Linux 2.6.10)
      also return if a stopped child has been resumed by delivery of SIGCONT.
```

It is presumably true for AT&T SysVR4. It is also true for FreeBSD 11 and OpenBSD 6.2. It is not true for macOS. So the test needs to be modified to allow for the possibility that SIGCONT may not be reported by `waitpid()` or `wait4()`.

This is the only test (not counting `signal` since it is currently disabled on all platforms) that still fails on macOS.

krader1961:
This builds on the fix in PR #557 but strictly speaking that dependency isn't needed. I just happened to create this branch from the branch for that pull-request. This is the primary commit message for this PR:

    Don't assume the platform reports SIGCONT via wait

    Some platforms, such as macOS, don't report receipt of SIGCONT via the
    `wait4()` or `waitpid()` functions. So modify the test so it passes if
    all the status changes are reported in the correct order regardless of
    whether or not SIGCONT is reported.

    Fixes #561

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -1683,10 +1683,10 @@ static_fn Namfun_t *clone_svar(Namval_t *np, Namval_t *mp, int flags, Namfun_t *
     Sfdouble_t d;
 
     // cppcheck-suppress pointerSize
-    dp = malloc(fp->dsize + sp->dsize + sizeof(void **));
+    dp = malloc(fp->dsize + sp->dsize + sizeof(void *));
     memcpy(dp, sp, fp->dsize);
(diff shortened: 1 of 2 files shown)


siteshwar:
LGTM

PR #564, opened 2018-05-31
Building api tests should be configurable

siteshwar:
Not everyone would like to build api tests, so building these tests
should be configurable.

Related: #558

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -16,3 +16,7 @@ option('read-timeout', type : 'string', value : '0')
 # the default pathname for recording shell auditable events.
 #
 option('audit-file', type : 'string', value : '/etc/ksh_audit')
+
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -16,3 +16,6 @@ option('read-timeout', type : 'string', value : '0')
 # the default pathname for recording shell auditable events.
 #
 option('audit-file', type : 'string', value : '/etc/ksh_audit')
+
+# Configure if api tests should be built
```
krader1961: We should start including actual examples so that people not familiar with Meson have an easier time. For example:

```
# To disable building and running non-KSH API tests configure the build thusly:
#
#   meson -Dbuild-api-tests=false
```

krader1961:
LGTM

PR #565, opened 2018-06-01
Fix coprocess deadlock

### LINKED ISSUE #563: ksh coprocess tests are flakey
The ksh `coprocess` unit test has always been flakey as of mid 2017 when I started contributing to this project. But since the switch from AST Vmalloc to the system malloc the failure rate has increased dramatically. Specifically, it is now timing out roughly 25% of the time. Whether this is due to a use-after-free bug or a failure exposed only because the layout of malloc()'ed memory is different is TBD (to be determined). Simply increasing that unit test timeout value does not help (see PR #560).

krader1961:
Fixes #563

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -19,6 +19,8 @@
 // prefix groups are listed first. The standard prefix symbols (e.g., `_sys_`,
 // `_hdr_`, etc.) based on tests like `cc.has_function()` or `cc.has_header()`
 // are enumerated later.
+#ifndef _CONFIG_AST_H
+#define _CONFIG_AST_H 1
 
 #define _GNU_SOURCE 1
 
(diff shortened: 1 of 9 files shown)


src/cmd/ksh93/sh/jobs.c
```diff
@@ -48,6 +48,8 @@
 
 #define NJOB_SAVELIST 4
 
+//volatile bool vmbusy_flag = false;
```
siteshwar: Shall this be removed ?

krader1961:
I refactored this so the wrappers are in a new *src/lib/libast/misc/vmbusy.c* module and added more comments and a TODO about refactoring the `job_waitsafe()` function so that the job reaping is never done in interrupt context.

PR #569, opened 2018-06-02
Add info about ksh availability in Mageia Cauldron

kekePower:


siteshwar:
@kekePower Out of curiousity, how frequently do you update this package ? Copr and OBS builds are automatically updated on each push on `master`.

kekePower:
Hi.

I update once or twice a month during the Cauldron stage and when Mageia 7 is released, only major security and major bug fixes will be released.

krader1961:
I am ambivalent about this pull-request. There are a large number of such links we could add to the documentation. While I appreciate that there are projects expressing an interest in keeping ksh relevant I don't think this is something we want to include at this time.

kekePower:
Sure, it's all up to you.

I just want to spread the word about Mageia and since you were already mentioning Fedora/Redhat and OpenSuSE, I thought why not mention Mageia as well.

Cheers.

krader1961:
I'm closing because there are a nearly infinite number of ksh package URLs for different distros that could be documented here. I don't think this information belongs in the primary `README` document (notwithstanding the links already in that doc). Long term I think a wiki based solution as proposed by @siteshwar is the better solution. Assuming that distros including ksh versions newer than ksh93u continue to be rare. My hope is that in a year or two this project has obtained sufficient buy in from mainstream distros that people can just assume they're using the version found here rather than the six year old ksh93u version.

siteshwar:
@kekePower @krader1961 https://github.com/att/ast/wiki/Packages-for-Linux

Issue #570, opened 2018-06-02
Override version numbers in package builds


siteshwar:
OBS and Copr builds set version number string to `0.0.0`. It should be possible to override version numbers while building.

PR #571, opened 2018-06-02
Fallback version number should be configurable

siteshwar:
Related: #570

### DIFF
--- a/.copr/prep.sh
+++ b/.copr/prep.sh
@@ -8,15 +8,20 @@ set -x
 # repo, so ignore if it fails.
 git fetch --unshallow || :
 
+# Fetch tags to determine version number
+git fetch --tags
+
 COMMIT=$(git rev-parse HEAD)
 COMMIT_SHORT=$(git rev-parse --short HEAD)
 COMMIT_NUM=$(git rev-list HEAD --count)
 COMMIT_DATE=$(date --date="@$(git show -s --format=%ct HEAD)" +%Y%m%d)
+VCS_VERSION=$(git describe --always --dirty --tags)
 
 sed "s,#COMMIT#,${COMMIT},;
      s,#SHORTCOMMIT#,${COMMIT_SHORT},;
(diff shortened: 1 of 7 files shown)


krader1961:
LGTM from a correctness standpoint. However, I'm an worried about how this will be used. How will the OBS and Copr builds incorporate the correct version string? That is, how will the `-Dfallback-version-number='x.y.z'` build flag be determined? And why can't we just do the same thing automatically without requiring a config time flag? What is to keep those build systems from specifying a misleading version string?

siteshwar:
@krader1961 See https://github.com/att/ast/pull/571/commits/eada2dce58bf7c27c406d4fbebdb11b1de070801

krader1961:
Since I have almost zero experience with distro packaging (other than as a consumer) I'm going to trust that it's necessary to do what you're doing in those scripts rather than just let Meson run the `git describe`. And for testing purposes being able to set the semantic version string to an arbitrary value might be useful. So, LGTM.

saper:
I'd reverse the logic not to require git at all if not needed: -Dversion-number shall it become and git being invoked only when not provided.

krader1961:
@saper The problem is that doesn't fit well with the Meson DSL. Or at least I don't see an obvious, simple, way to do what you're describing. I don't doubt it could be done but since I'm not a distro maintainer I'm not motivated to figure out how to do it.

Everyone building from a git clone of the project with an available git command just runs `meson` and the correct version string will be extracted. Other situations should be limited to things like distro package builds where they may want more control over the version string anyway (e.g., to reflect locally applied patches). If they want that control and a) can't simply ensure a `git` command isn't found or b) the source tree doesn't contain the *.git* directory then they can add a trivial `git` script that does nothing more than echo the desired version string.

Having said all that if you create a PR that only uses `git` if `-Dversion-string` isn't provided we'd be happy to merge it.

saper:
Ok, I'll be using bin/package for now :)

Issue #375, opened 2018-01-19
`pty` command doesn't handle SIGTSTP correctly


krader1961:
While working on changing the Meson build system to also build the `pty` command so its associated unit tests could be run I noticed something interesting. There is one test for \cZ to suspend the current process. The `ksh` command handles that situation correctly. That is, if you run something like `/bin/sleep 30` and press \cZ the sleep is suspended. The standalone `pty` command results in the blocking syscall of the command being resumed rather than suspending the process.

Here is a `strace` of */bin/sleep* (substituted for the `less` command in that unit test to simplify the problem):

```
1516334684.166476 restart_syscall(<... resuming interrupted nanosleep ...>) = ? ERESTART_RESTARTBLOCK (Interrupted by signal)
1516334685.650671 --- SIGTSTP {si_signo=SIGTSTP, si_code=SI_USER, si_pid=95709, si_uid=1000} ---
1516334685.650699 restart_syscall(<... resuming interrupted restart_syscall ...>) = ? ERESTART_RESTARTBLOCK (Interrupted by signal)
1516334692.311947 --- SIGSTOP {si_signo=SIGSTOP, si_code=SI_USER, si_pid=95727, si_uid=1000} ---
1516334692.312118 --- stopped by SIGSTOP ---
1516334695.540488 --- SIGHUP {si_signo=SIGHUP, si_code=SI_KERNEL} ---
1516334695.541189 +++ killed by SIGHUP +++
```

Notice that the nanosleep is restarted. What should happen is this:

```
1516335967.922413 restart_syscall(<... resuming interrupted nanosleep ...>) = ? ERESTART_RESTARTBLOCK (Interrupted by signal)
1516335970.061081 --- SIGTSTP {si_signo=SIGTSTP, si_code=SI_KERNEL} ---
1516335970.061546 --- stopped by SIGTSTP ---
```

TBD is what `pty` is doing wrong in setting up the environment such that the libc syscall trampoline code is incorrectly deciding to restart the interrupted syscall. Note that this happens on both BSD and GNU Linux.

krader1961:
I double-checked the beta branch and using the legacy build system the same failure occurs:

```
test pty begins at 2018-01-19+12:36:58
        pty.sh[431]: process/terminal group exercise: line 435: expected ":$|:\E|lines", got EOF
test pty failed at 2018-01-19+12:37:13 with exit code 1 [ 20 tests 1 error ]
```

kernigh:
[...]

PR #580, opened 2018-06-07
Enable all pty tests, disable /etc/ksh.kshrc, avoid seq(1)

kernigh:
Disable /etc/ksh.kshrc so it can't change PS1 or other environment
variables; such changes can break the tests.  Switch from `seq 100` to
`{1..100}` because OpenBSD has no seq(1) command.  Remove TODO
comments and enable tests in pty.sh.  Some tests use stty(1); these
should work even if stty(1) is not a builtin.

This enables the test for https://github.com/att/ast/issues/375

### DIFF
--- a/src/cmd/ksh93/tests/pty.sh
+++ b/src/cmd/ksh93/tests/pty.sh
@@ -64,7 +64,9 @@ function tst
     done
 }
 
-export PS1=':test-!: ' PS2='> ' PS4=': ' ENV= EXINIT= HISTFILE= TERM=dumb VISUAL=vi LC_ALL=C
+# ENV begins with /./ to disable /etc/ksh.kshrc.
+export PS1=':test-!: ' PS2='> ' PS4=': ' ENV=/./dev/null EXINIT= HISTFILE=
+export TERM=dumb VISUAL=vi LC_ALL=C
 
 if ! pty $bintrue < /dev/null
 then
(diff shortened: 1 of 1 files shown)


krader1961:
Tentatively LGTM. I'll test it tomorrow on my VMs and make a more careful review of the change.

siteshwar:
LGTM

krader1961:
This appears to make the test robust on OpenBSD. It also doesn't fail on macOS. But it fails intermittently on Ubuntu 16.04:

```
<E> pty[426]: process/terminal group exercise: line 10: expected "Stopped", got EOF
```

It also fails consistently for me on my OpenSuse VM:

```
<E> pty[426]: process/terminal group exercise: line 8: expected "seq-24", got EOF
```

And on FreeBSD 11.1:

```
<E> pty[361]: POSIX sh 137(C): line 17: expected "^hello world\r?\n$", got EOF
<E> pty[426]: process/terminal group exercise: line 10: expected "Stopped", got EOF
```

So I'm inclined to merge it since it looks like a step in the right direction but with the test remaining disabled.

kernigh:
"POSIX sh 137(C)" runs vi(1) in TERM=dumb, and "process/terminal group exercise" runs less(1) in TERM=vt100. Both less(1) and vi(1) are interactive programs that might not behave the same way on every platform.

I now edit this pull request: commit 2571815 keeps the kshrc and seq(1) changes but doesn't enable any disabled tests. I left the old commit (which enables the tests) in a different branch [kernigh-pty-1](https://github.com/kernigh/ast/tree/kernigh-pty-1).

krader1961:
Thanks, @kernigh, but I already merged your change and layered another one on top to disable the last test and one of the tests on FreeBSD only. It's not passing consistently on all platforms we've been testing on.

Issue #567, opened 2018-06-01
Ksh fails to start if current directory is not accessible


siteshwar:
**Description of problem:**
Ksh fails to start if current directory is not accessible

**Ksh version:**
ksh-20180526+git.945.4ed72ac1-0.fc27.x86_64

**How reproducible:**
Always

**Steps to reproduce:**
1. mkdir /tmp/foo
2. chmod o-rwx /tmp/foo
3. useradd test
4. cd /tmp/foo
5. su -c "ksh" test

**Actual results:**
ksh: Can't obtain directory fd. [Permission denied]

and ksh fails to start.

**Expected results:**
ksh should start.

**Additional info:**
Here is the code from `src/cmd/ksh93/sh/init.c` that throws this error:

```
    // This should _never_ happen, guaranteed by design and goat sacrifice.
    if (shp->pwdfd < 0) errormsg(SH_DICT, ERROR_system(1), "Can't obtain directory fd.");
```

We should not need another goat sacrifice to fix it.

krader1961:
Note that this works with ksh93u. It's also not necessary to create a dummy user account. Just do this instead:

1. mkdir /tmp/foo
1. cd /tmp/foo
1. chmod 0 .
1. ksh

PR #583, opened 2018-06-07
Disable code that causes error on start up

siteshwar:
If ksh is started in a directory that is not accessible to it, it will
exit with an error. This code makes invalid assumption and should be
disabled.

Resolves: #567


krader1961:
LGTM. I did a quick review of the uses of that structure member and it looks like they all protect against it being less than zero. Which begs the question of why the error exit you commented out was even added. Another example of why I which the AST team would provide the change history for the code.

Issue #568, opened 2018-06-01
Bash fix for CVE-2014-7169 breaks ksh scripts that try to clear the environment


siteshwar:
**Description of problem:**
Bash fix for CVE-2014-7169 breaks ksh scripts that try to clear the environment

**Ksh version:**
ksh-20180526+git.945.4ed72ac1-0.fc27.x86_64

**How reproducible:**
Always

**Steps to reproduce:**
1. Add these scripts:

`run.sh`:

```
a() {
    echo "BASH_FUNC_a() survives"
}

export -f a
ksh clear.sh
```
 `clear.sh`:

```
#!/bin/ksh
RC=0
for variable in `set | sed 's/=.*//'` ; do
    echo "Unsetting $variable"
    unset $variable || RC=1
done
bash -c "a"
```
2. Execute `bash run.sh`

**Actual results:**
clear.sh[5]: unset: BASH_FUNC_a%%: invalid variable name

**Expected results:**
No error should appear.

**Additional info:**
Bug report has been copied from https://bugzilla.redhat.com/show_bug.cgi?id=1147645
Patch: https://src.fedoraproject.org/rpms/ksh/blob/master/f/ksh-20120801-oldenvinit.patch

This patch skips inheriting invalid variables.

kernigh:
Step 2 should be `bash run.sh`, not `ksh run.sh`.

The bug is that ksh can't unset an environment variable with `%` in its name. Here I reproduce the bug as perl sets `evil%` and ksh can't unset it:

```
$ ksh --version
  version         sh (AT&T Research) 93v-958-g19c078ab
$ perl -e '$ENV{"evil%"}=value; system @ARGV' ksh -c \                         
> 'unset evil%; printenv evil%'
ksh: unset: evil%: invalid variable name
value
```

In fact, bash has the exact same bug as ksh:

```
$ bash --version
GNU bash, version 4.4.19(1)-release (x86_64-unknown-openbsd6.3)
...
$ perl -e '$ENV{"evil%"}=value; system @ARGV' bash -c \                        
> 'unset evil%; printenv evil%'
bash: line 0: unset: `evil%': not a valid identifier
value
```

Other shells are different:

- In dash, `unset evil%` shows an error, but dash always clears `evil%` from the environment, whether or not I use `unset evil%`.
- In yash, `unset evil%` correctly unsets it.

siteshwar:
> Step 2 should be bash run.sh, not ksh run.sh.

Fixed!

PR #584, opened 2018-06-09
Avoid inheriting invalid environment variables

siteshwar:
Variables that have invalid names should not be inherited from
environment. This commit restores env_init() function definition back to
it's definition from ksh-2010-06-21.

Resolves: #568


krader1961:
LGTM although I recommend running `bin/style` to fix the style of the code before merging.

Also, this is yet another example of rolling back a change made after the ksh93u release. I wish we could get our hands on the change log for every change made after that release. Because so much of that work has been shown to introduce serious bugs.

Issue #577, opened 2018-06-05
Test failures on FreeBSD 11.1


krader1961:
We have reduced the consistent unit test failure count on macOS, various Linux distros, and OpenBSD to zero. One platform, FreeBSD 11.1, still has 17 unit tests failing on a consistent basis. Three of those are in the `wchar` test. Here is a sample failure log from that unit test:

```
--- stdout ---
<I> wchar[-1]: TEST_DIR=/tmp/ksh.wchar.XXXXXXX.UlUVLNOF
<I> wchar[33]: LC_ALL=en_US.UTF-8 is supported and will be tested
<I> wchar[35]: LC_ALL=en_US.utf8 not supported
<I> wchar[35]: LC_ALL=en_US.ISO-8859-15 not supported
<I> wchar[35]: LC_ALL=en_US.iso885915 not supported
<I> wchar[33]: LC_ALL=zh_CN.GB18030 is supported and will be tested
<I> wchar[35]: LC_ALL=zh_CN.gb18030 not supported
--- stderr ---
<E> wchar[49]: C.UTF-8 nounicodeliterals FAILED
<E> expect: |0000000 24 27 e2 82 ac 27 0a|
<E> actual: |0000000 24 27 5c 75 5b 32 30 61 63 5d 27 0a|
<E> wchar[55]: C.UTF-8 (nounicodeliterals) FAILED
<E> expect: |0000000 24 27 e2 82 ac 27 0a|
<E> actual: |0000000 24 27 5c 75 5b 32 30 61 63 5d 27 0a|
<E> wchar[49]: zh_CN.GB18030 nounicodeliterals FAILED
<E> expect: |0000000 24 27 e2 82 ac 27 0a|
<E> actual: |0000000 27 5c 75 5b 32 30 61 63 5d 27 0a|
<E> wchar[55]: zh_CN.GB18030 (nounicodeliterals) FAILED
<E> expect: |0000000 24 27 e2 82 ac 27 0a|
<E> actual: |0000000 27 5c 75 5b 32 30 61 63 5d 27 0a|
<W> wchar[-1]: error_count = 4
-------
```

And from the `builtins` unit test:

```
<E> builtins[393]: printf "%T" now = Mon Jun  4 22:00:23 PPT 2018 != Mon Jun  4 22:00:23 PDT 2018
```

Notice the change in TZ .  from `PPT` to `PDT`.

krader1961:
[...]

PR #587, opened 2018-06-11
Fix `wchar` unit test on FreeBSD

krader1961:
Part of resolving #577

### DIFF
--- a/src/cmd/ksh93/tests/wchar.sh
+++ b/src/cmd/ksh93/tests/wchar.sh
@@ -21,7 +21,13 @@
 # Tests for \u[...] and \w[...] input and output
 #
 
-locales="en_US.UTF-8 en_US.utf8 en_US.ISO-8859-15 en_US.iso885915 zh_CN.GB18030 zh_CN.gb18030"
+locales="en_US.UTF-8 en_US.utf8 en_US.ISO-8859-15 en_US.iso885915"
+# On FreeBSD 11.1 the native locale subsystem cannot convert the Euro currency symbol from UTF-8 to
+# GB1803. See https://github.com/att/ast/issues/577#issuecomment-396113061.
+if [[ $OS_NAME != "FreeBSD" ]]
+then
+    locales="$locales zh_CN.GB18030 zh_CN.gb18030"
+fi
 supported="C.UTF-8"
 locale -a > locale.txt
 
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/wchar.sh
```diff
@@ -21,7 +21,13 @@
 # Tests for \u[...] and \w[...] input and output
 #
 
-locales="en_US.UTF-8 en_US.utf8 en_US.ISO-8859-15 en_US.iso885915 zh_CN.GB18030 zh_CN.gb18030"
+locales="en_US.UTF-8 en_US.utf8 en_US.ISO-8859-15 en_US.iso885915"
+# On FreeBSD 11.1 the native locale subsystem cannot convert the Euro currency symbol from UTF-8 to
+# GB1803. See https://github.com/att/ast/issues/577#issuecomment-396113061.
+if [[ $OS_NAME != "FreeBSD" ]]
+then
+    locales="$locales zh_CN.GB18030 zh_CN.gb18030"
```
siteshwar: Are the locales `zh_CN.GB18030` and `zh_CN.gb18030` supported on FreeBSD ? Shall we add a todo note to write separate tests for it ?

siteshwar:
LGTM

src/cmd/ksh93/tests/wchar.sh
krader1961: They are supported but for some reason the U+20AC code point can't be translated. Most likely the FreeBSD GB18030 code table is old and missing that character.

siteshwar: @saper Do you think it could be a bug in FreeBSD ?

saper: Certainly, will have a look!

krader1961: @saper You might also want to take a look into why the `LC_ALL` env var doesn't supercede the other locale env vars. Create a file with a UTF-8 encoded U+20AC (the Euro currency symbol). Then run `env LC_ALL=C LC_CTYPE=en_US.UTF-8 wc -m ~/x`. It should report `3` and does on all the other platforms (macOS, OpenBSD, GNU Linux). On FreeBSD 11.1 it reports `1`.

PR #588, opened 2018-06-11
Make unit test `subshell` more predictable

krader1961:
Some platforms, such as FreeBSD, have a unusually large limit on the number
of open file descriptors. In the case of FreeBSD 11.1 it is 44685. That
causes this test to fail (for unknown reasons). So limit the max file
descriptor number to something more reasonable and comparable to all the
other platforms.

TBD is why such a huge default file descriptor limit causes this test to
fail.

Partial fix for #577


siteshwar:
LGTM

Issue #589, opened 2018-06-11
Wrong return code from a pipe in command substitution


siteshwar:
**Description of problem:**
Return code from command substitution is set as if `pipefail` option is on.

**Ksh version:**
ksh-20180606+git.976.4784bab3-0.fc27.x86_64

**How reproducible:**
Always

**Steps to reproduce:**
1. Run this script:

```
#!/bin/ksh

count=`ls *foo 2> /dev/null | wc -l`
echo $?
```

**Actual results:**
Exit status is `2`

**Expected results:**
Exit status should be `0`

**Additional info:**
Bug report has been copied from https://bugzilla.redhat.com/show_bug.cgi?id=1117316

krader1961:
FWIW, I can understand why they changed the behavior because it is extremely useful and is what most people intuitively expect. It should have been the day one behavior. However, you're absolutely correct that it can't be changed now, at least not without making a major release, because it isn't backward compatible.

PR #590, opened 2018-06-11
Do not set pipefail option in backticks substitution

siteshwar:
This is a backward incompatible change in ksh93v- release.
Backticks substitution should not set pipefail option.    

Resolves: #589


krader1961:
LGTM

krader1961:
This also needs an entry in *CHANGELOG.md* noting that this reverts 93v command substitution behavior to 93u behavior.

PR #591, opened 2018-06-11
Fix timezone name determination

krader1961:
On FreeBSD calling `tzset()` does not guarantee the `tzname` array will
be correctly populated. On most systems that works but on FreeBSD you
have to call `localtime()` or a related function (e.g., `ctime()`).

This change also eliminates a potential, very small, memory leak due to
the `strdup()`'ed tznames not being freed.

Partial fix for #577


krader1961:
merged

PR #592, opened 2018-06-11
Fix couple of crashes while sourcing files

siteshwar:


src/cmd/ksh93/bltins/misc.c
```diff
@@ -298,7 +298,8 @@ int b_dot_cmd(int n, char *argv[], Shbltin_t *context) {
     }
     sh_popcontext(shp, &buff);
     if (buffer) free(buffer);
-    if (!np) free(shp->st.filename);
+    // Following line was commented as a workaround for https://github.com/att/ast/issues/582
+    // if (!np) free(shp->st.filename);
```
krader1961: As best I can tell all but one of the assignments to `shp->st.filename` involve `strdup()` or equivalent. So this introduces a memory leak. The one place I could find which does not make a copy that can be freed is in *src/cmd/ksh93/sh/xec.c*:

```
shp->st.filename = fp->node->nvalue.rp->fname;
```

Does wrapping that in `strdup()` while leaving this `free()` in effect fix it?

src/cmd/ksh93/sh/io.c
```diff
@@ -1226,7 +1226,10 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
                         message = e_file;
                         goto fail;
                     }
-                    sh_iovalidfd(shp, dupfd);
+                    if (!sh_iovalidfd(shp, dupfd)) {
+                        message = e_file;
+                        goto fail;
+                    }
```
krader1961: This looks fine but I'm suspicious of every single place that calls `sh_iovalidfd()` without checking it's return value. It seems to me that at a minimum if a failure should be a "can't happen" situation those sites should be rewritten as follows if we can't handle it as a soft failure like you're doing with this change:

```
if (!sh_iovalidfd(shp, dupfd)) abort();
```

src/cmd/ksh93/bltins/misc.c
siteshwar: Thanks for the hint. It was not this statement, but the one I fixed in https://github.com/att/ast/pull/592/commits/b089e63248c43b1326bedda6f7d454f4497418a5 that fixes the crash.

src/cmd/ksh93/sh/io.c
siteshwar: I have opened #595 to track this.

krader1961:
LGTM

PR #594, opened 2018-06-12
Remove another use of `pathdev()`

krader1961:
The `ast_pathconf()` function is only used in the pathconf.c module.
Remove it and thus a caller of `pathdev()`. Remove the block of code
predicated on preprocessor symbol `_PC_PATH_ATTRIBUTES` as it isn't
defined and appears, based on google searches, to be unique to the UWIN
environment which we no longer support.

This leaves just two callers of `pathdev()`: in fgetcwd.c and intercept.c.

Partial fix for #510


siteshwar:
LGTM

PR #597, opened 2018-06-13
Fix floating point equality tests

krader1961:
You can't compare a computed floating point value to a constant when the
values cannot be expressed exactly in binary floating point representation.
You have to compute the absolute difference and verify it is less than
some epsilon.

Partial fix for #546


siteshwar:
LGTM

PR #599, opened 2018-06-15
Move code to build on docker in different script

siteshwar:
This simplifies the .travis.yml file.


krader1961:
LGTM. Thanks for this refactor. Embedding the script in the Travis config file was fine when it was short and simple which is obviously no longer true.

Issue #595, opened 2018-06-12
Return value of sh_iovalidfd() function should be checked


siteshwar:
`sh_iovalidfd()` function fails on invalid fds and it's return value can be checked to avoid crashes like the one mentioned in #582. But at majority of the places it's return value is not checked. This should be fixed.

PR #600, opened 2018-06-15
Abort if sh_iovalidfd() function fails

siteshwar:
This function fails if an invalid fd is passed to it. Ksh should abort
on such conditions.

Resolves: #595

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -215,7 +215,7 @@ bool sh_inuse(Shell_t *shp, int fd) { return fd < shp->gd->lim.open_max && shp->
 void sh_ioinit(Shell_t *shp) {
     filemapsize = 8;
     filemap = (struct fdsave *)malloc(filemapsize * sizeof(struct fdsave));
-    sh_iovalidfd(shp, 16);
+    if (!sh_iovalidfd(shp, 16)) abort();
     shp->sftable[0] = sfstdin;
     shp->sftable[1] = sfstdout;
     shp->sftable[2] = sfstderr;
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -215,7 +217,7 @@ bool sh_inuse(Shell_t *shp, int fd) { return fd < shp->gd->lim.open_max && shp->
 void sh_ioinit(Shell_t *shp) {
     filemapsize = 8;
     filemap = (struct fdsave *)malloc(filemapsize * sizeof(struct fdsave));
-    sh_iovalidfd(shp, 16);
+    assert(sh_iovalidfd(shp, 16));
```
krader1961: Never use `assert()` if evaluating the condition has side-effects. The assert call can be optimized out depending on build options. Which in this case means the `sh_iovalidfd()` call would not be made. Instead do `if (!sh_iovalidfd(shp, 16)) abort();`.

src/cmd/ksh93/sh/path.c
```diff
@@ -427,7 +427,11 @@ static_fn int path_opentype(Shell_t *shp, const char *name, Pathcomp_t *pp, int
             }
         }
     } while (fd < 0 && pp);
-    sh_iovalidfd(shp, fd);
+
+    if (fd > 0) {
```
krader1961: Should this be `fd >= 0`?

krader1961:
LGTM modulo the one possible incorrect conditional.

src/cmd/ksh93/sh/path.c
siteshwar: Fixed.

PR #602, opened 2018-06-17
Remove no longer needed astlicense.c module.

krader1961:
This module was used by the legacy build system (specifically the
`pprobe` command). Since we no longer use that build system and the
assoiated code has been removed we no longer need this module.


siteshwar:
LGTM

PR #603, opened 2018-06-18
Fix segfault when handling trap

siteshwar:
Resolves: rhbz#1117404

### DIFF
--- a/src/cmd/ksh93/sh/subshell.c
+++ b/src/cmd/ksh93/sh/subshell.c
@@ -423,12 +423,12 @@ void sh_subjobcheck(pid_t pid) {
 Sfio_t *sh_subshell(Shell_t *shp, Shnode_t *t, volatile int flags, int comsub) {
     struct subshell sub_data;
     struct subshell *sp = &sub_data;
-    int jmpval, nsig = 0, duped = 0;
+    int jmpval, isig, nsig = 0, duped = 0;
     long savecurenv = shp->curenv;
     int savejobpgid = job.curpgid;
     int *saveexitval = job.exitval;
     int16_t subshell;
-    char *savsig;
+    char **savsig;
     Sfio_t *iop = 0;
     struct checkpt buff;
     struct sh_scoped savst;
@@ -503,9 +503,14 @@ Sfio_t *sh_subshell(Shell_t *shp, Shnode_t *t, volatile int flags, int comsub) {
         // Save trap table.
         shp->st.otrapcom = 0;
         shp->st.otrap = savst.trap;
-        if ((nsig = shp->st.trapmax * sizeof(char *)) > 0 || shp->st.trapcom[0]) {
-            nsig += sizeof(char *);
-            memcpy(savsig = malloc(nsig), (char *)&shp->st.trapcom[0], nsig);
+        nsig = shp->st.trapmax;
+        if (nsig > 0 || shp->st.trapcom[0]) {
+            ++nsig;
+            savsig = malloc(nsig * sizeof(char *));
+            // Contents of shp->st.st.trapcom may change
+            for (isig = 0; isig < nsig; ++isig) {
+                savsig[isig] = shp->st.trapcom[isig] ? strdup(shp->st.trapcom[isig]) : NULL;
+            }
             // This nonsense needed for $(trap).
             shp->st.otrapcom = (char **)savsig;
         }
@@ -661,7 +666,12 @@ Sfio_t *sh_subshell(Shell_t *shp, Shnode_t *t, volatile int flags, int comsub) {
         shp->curenv = savecurenv;
         shp->st.otrap = 0;
         if (nsig) {
-            memcpy((char *)&shp->st.trapcom[0], savsig, nsig);
+            for (isig = 0; isig < nsig; ++isig) {
+                if (shp->st.trapcom[isig]) {
+                    free(shp->st.trapcom[isig]);
+                }
+            }
+            memcpy(shp->st.trapcom, savsig, nsig * sizeof(char *));
             shp->st.otrapcom = NULL;
             free(savsig);
         }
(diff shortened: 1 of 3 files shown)


siteshwar:
I was not able to reproduce this bug on fedora, but it's easily reproducible on RHEL.

siteshwar:
Patch has been taken from https://src.fedoraproject.org/rpms/ksh/blob/master/f/ksh-20120801-trapcom.patch

src/cmd/ksh93/sh/subshell.c
```diff
@@ -503,9 +503,14 @@ Sfio_t *sh_subshell(Shell_t *shp, Shnode_t *t, volatile int flags, int comsub) {
         // Save trap table.
         shp->st.otrapcom = 0;
         shp->st.otrap = savst.trap;
-        if ((nsig = shp->st.trapmax * sizeof(char *)) > 0 || shp->st.trapcom[0]) {
-            nsig += sizeof(char *);
-            memcpy(savsig = malloc(nsig), (char *)&shp->st.trapcom[0], nsig);
+        nsig = shp->st.trapmax;
+        if (nsig > 0 || shp->st.trapcom[0]) {
```
krader1961: I know you're just copying the original statement but now is the time to think about whether it is correct. Shouldn't it be impossible for `st.trapmax` to be zero yet a trap be defined? Shouldn't this just be `if (nsig)`?

src/cmd/ksh93/sh/xec.c
```diff
@@ -1774,7 +1774,8 @@ int sh_exec(Shell_t *shp, const Shnode_t *t, int flags) {
                     struct checkpt *buffp =
                         (struct checkpt *)stkalloc(shp->stk, sizeof(struct checkpt));
                     shp->st.otrapcom = 0;
-                    if ((nsig = shp->st.trapmax * sizeof(char *)) > 0 || shp->st.trapcom[0]) {
+                    nsig = shp->st.trapmax * sizeof(char *);
```
krader1961: As you did in *subshell.c* the multiplication by `sizeof(char *)` should be deferred to the actual malloc call.

src/cmd/ksh93/sh/xec.c
```diff
@@ -3339,9 +3348,16 @@ int sh_funscope_20120720(Shell_t *shp, int argn, char *argv[], int (*fun)(void *
     shp->topscope = (Shscope_t *)prevscope;
     nv_getval(sh_scoped(shp, IFSNOD));
     shp->end_fn = 0;
-    if (nsig) memcpy((char *)&shp->st.trapcom[0], savstak, nsig);
+    if (nsig) {
```
krader1961: You can omit this since the for loop will execute zero times if nsig is zero.

krader1961: Doh! Never mind. Obviously the other two statements need to be predicated on this.

src/cmd/ksh93/sh/xec.c
```diff
@@ -3339,9 +3348,16 @@ int sh_funscope_20120720(Shell_t *shp, int argn, char *argv[], int (*fun)(void *
...
+        for (isig = 0; isig < nsig; ++isig) {
+            if (shp->st.trapcom[isig] && shp->st.trapcom[isig] != Empty) {
+                free(shp->st.trapcom[isig]);
+            }
+        }
+        memcpy((char *)&shp->st.trapcom[0], savsig, nsig * sizeof(char *));
```
krader1961: The convoluted addressing and typecasting isn't needed; just `memcpy(shp->st.trapcom, ...)`.

src/cmd/ksh93/sh/xec.c
```diff
@@ -3274,9 +3276,16 @@ int sh_funscope_20120720(Shell_t *shp, int argn, char *argv[], int (*fun)(void *
...
-    if ((nsig = shp->st.trapmax * sizeof(char *)) > 0 || shp->st.trapcom[0]) {
-        nsig += sizeof(char *);
-        memcpy(savstak = stkalloc(shp->stk, nsig), (char *)&shp->st.trapcom[0], nsig);
+    nsig = shp->st.trapmax;
+    if (nsig > 0 || shp->st.trapcom[0]) {
+        ++nsig;
+        savsig = malloc(nsig * sizeof(char *));
+        // Contents of shp->st.trapcom may change
+        for (isig = 0; isig < nsig; ++isig) {
+            savsig[isig] = shp->st.trapcom[isig] == Empty
```
krader1961: The magic pointer `Empty` is special-cased here but not in the similar code in subshell.c which seems wrong. I would suggest removing this special-case. Instead change this line in *src/cmd/ksh93/bltins/trap.c*:

```
shp->st.trapcom[sig] = Empty;
```

to

```
shp->st.trapcom[sig] = strdup("");
```

Then all of the places that manipulate `shp->st.trapcom` can just assume every entry was dynamically allocated.

src/cmd/ksh93/sh/subshell.c
siteshwar: I think that's why `shp->st.trapcom[0]` is being checked here.

krader1961: I admit I haven't looked closely at the code to understand how `st.trapmax` is used. If it is the maximum offset into the `st.trapcom` array then it should be set to -1 rather than zero if the array is empty. In which case the check of `st.trapcom[0]` isn't needed. Regardless, it would be better would be to redefine it as `st.trapcount` so that zero means there are no traps.

In any case ignore this rant. I'm just pointing out how stupid and confusing the existing code is about such matters. The existing pattern pretty much guarantees bugs will be present in any code that manipulates shell traps. Which was why I was suggesting expanding the scope of the change to fix the core problem.

krader1961:
LGTM

PR #604, opened 2018-06-19
Suppress `-Wchar-subscripts` warnings

krader1961:
There are quite a few places in the code which use a char to index into
an array. All of them appear to be "safe" in as much as they only do so
when the value involves an ASCII character which, by definition, won't
be negative.


krader1961:
So this doesn't actually "fix" anything. But it does eliminate quite a few compiler warnings which are almost certainly false positives. Which makes it much easier to notice when a change introduces a new warning.

siteshwar:
LGTM

Issue #203, opened 2017-12-08
Reading file through command substitution fails


siteshwar:
Steps to reproduce:
```
~$ ksh -c 'echo .$(</proc/sys/kernel/osrelease).'
.2.6.32-431.5.1.el6.x86_64.
~$ ksh -c 'echo .$(</proc/sys/kernel/osrelease). >&2' >&-
..
~$ ksh -c 'echo .$(</proc/sys/kernel/osrelease).' 2>&-
```

Actual results:
Substituted text is missing between ..

Expected results:
Command substitution should work.

The test case and possible patch to fix it were put in #185. But the patch is causing regressions with other tests.

PR #605, opened 2018-06-19
Fix for reading files through $(<) when stderr is closed

siteshwar:
Resolves: #203

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -1332,6 +1332,7 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
             } else if (!(iof & IOPUT)) {
                 if (flag == SH_SHOWME) goto traceit;
                 fd = sh_chkopen(fname);
+                fd = sh_iomovefd(shp, fd);
             } else if (sh_isoption(shp, SH_RESTRICTED)) {
                 errormsg(SH_DICT, ERROR_exit(1), e_restricted, fname);
             } else {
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -1492,12 +1493,14 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
...
                         if ((fn = fcntl(fd, F_DUPFD, 10)) < 0) goto fail;
                         if (fn >= shp->gd->lim.open_max && !sh_iovalidfd(shp, fn)) goto fail;
-                        if (flag != 2 || shp->subshell)
-                            sh_iosave(shp, fn, indx | 0x10000, tname ? fname : (trunc ? Empty : 0));
                         shp->fdstatus[fn] = shp->fdstatus[fd];
                         sh_close(fd);
                         fd = fn;
                     }
+
+                    if (flag != 2 || shp->subshell)
```
krader1961: Any idea what the magic number `2` is? Is it meant to be `IOWRITE`?

src/cmd/ksh93/sh/io.c
```diff
@@ -1492,12 +1493,14 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
...
+                        sh_iosave(shp, fn, indx | 0x10000, tname ? fname : (trunc ? Empty : 0));
```
krader1961: Same thing here with 0x10000. It looks like it's meant to be `IOPICKFD`.

src/cmd/ksh93/sh/io.c
krader1961: Never mind. I see from the function comment that these are magic constants without associated symbols.

src/cmd/ksh93/tests/io.sh
```diff
@@ -588,3 +588,11 @@ n=$( exec {n}< /dev/null; print -r -- $n)
 
 # According to POSIX <> should redirect stdin when no fd is specified
 read -n 1 -t 0.1 <>/dev/zero || log_error '<> should redirect stdin by default'
+
+TMPF=$(mktemp -p $TEST_DIR)
+echo foo >$TMPF
+export TMPF
+
+[ -n "$($SHELL -c 'echo $(<$TMPF)' <&-)" ] || log_error "Closing stdin causes failure when reading file through \$(<)"
```
krader1961: Any reason you're not using the `[[ ... ]]` syntax?

src/cmd/ksh93/tests/io.sh
```diff
@@ -588,3 +588,11 @@ n=$( exec {n}< /dev/null; print -r -- $n)
 
 # According to POSIX <> should redirect stdin when no fd is specified
 read -n 1 -t 0.1 <>/dev/zero || log_error '<> should redirect stdin by default'
+
+TMPF=$(mktemp -p $TEST_DIR)
```
krader1961: Don't use `mktemp -p` as it isn't available on BSD/macOS. Since the test harness has already created a unique test dir and this test doesn't require a unique file you can just hardcode the file name.

krader1961:
I don't actually expect you to change the magic numbers to equivalent symbols in this change. I was just pointing out that magic numbers of that nature are dangerous and confusing. So I'd love to see a second change to address that.

src/cmd/ksh93/tests/io.sh
siteshwar: Fixed.

krader1961:
LGTM

PR #611, opened 2018-06-21
Disable SHOPT_BASH macro by default

siteshwar:


krader1961:
LGTM

krader1961:
Don't forget to add a comment in *CHANGELOG.md* that this disables a feature that was enabled in ksh93v-.

siteshwar:
> Don't forget to add a comment in CHANGELOG.md that this disables a feature that was enabled in ksh93v-.

Shall this be considered a backward incompatible change ? This change is backward compatible with ksh93u+ but incompatbile with ksh93v- (which was a  beta release).

krader1961:
I guess that depends on how likely we think it is that people have been using the ksh93v- release even though it was never picked up by any distros AFAICT. So far we have one CHANGELOG entry related to a ksh93v- feature we removed. I'm inclined to add a separate section labeled something like `## Changes from ksh93v- beta release` and put these comments there.

krader1961:
@siteshwar Despite my previous comment I am okay with not documenting changes that are backward compatible with 93u+ but not 93v-. You could argue that anyone using a beta release never picked up by any distro should not be surprised if the next non-beta release does not provide some facility of the beta release they were using.

Issue #515, opened 2018-05-08
How to enable CI for non-Linux operating systems ?


siteshwar:
Currently we have setup Travis builds for fedora and opensuse only. Travis has support for macOS but we have not found a way to integrate it in our current setup. I would like to hear ideas about how we can enable CI for non-Linux operating systems like macOS, different BSDs etc.

dannyweldon:
What about something like:

    matrix:
        include:
            - os: linux
            - env: OS_TYPE=fedora
            - env: INSTALL_REQUIREMENTS="dnf repolist; dnf install -y gcc meson sudo langpacks-zh_CN ed ncurses vi"
    
            - os: linux
            - env: OS_TYPE=opensuse
            - env: INSTALL_REQUIREMENTS="zypper refresh; zypper in -y gcc python3 python3-pip ninja sudo glibc-locale glibc-devel ed ncurses-utils vim; pip3 install meson==0.44.0"
    
            - os: linux
            - env: OS_TYPE=dauchida/freebsd
            - env: INSTALL_REQUIREMENTS="pkg install ..."
    
            - os: osx
            - env: OS_TYPE=osx
            - env: INSTALL_REQUIREMENTS="brew update; brew install ..."

    before_install:
        - [[ $TRAVIS_OS_NAME == "linux" ]] && docker pull ${OS_TYPE}
    
    script:
        # Test for osx and run different build or prep commands:
        - [[ $TRAVIS_OS_NAME == "osx" ]] && ...
        - [[ $TRAVIS_OS_NAME == "linux" ]] && ...

I'm not sure if the above will work, but you get the idea.  The script will get a bit messier but it will get the job done.

Adapted from:

https://docs.travis-ci.com/user/multi-os/#Python-example-(unsupported-languages)

krader1961:
I looked into enabling macOS (OS X) for this project long ago but couldn't find a way to do so with a docker based config. What I didn't consider was seeing if OpenSuse and Fedora could be installed using a non-docker based config as in @dannyweldon's example.

I do agree that we should add a mainstream Linux distro (e.g., Ubuntu), macOS, and one other BSD flavor (either FreeBSD or OpenBSD). That should be plenty of CI coverage. Other platforms (e.g., Cygwin, Solaris, Haiku) can be tested adhoc by those with an interest in running ksh on those platforms.

siteshwar:
@dannyweldon Thanks for the hint. Your example may work with our set up.

[...]

PR #612, opened 2018-06-21
Enable unit tests on macOS

krader1961:
I eliminated all unit test failures on macOS some weeks ago. So it is
time to enable them in the Travis macOS environment.

Don't hadcode the value of CORE_COUNT.  Instead query the system and
default to one core if we can't figure out the actual number of CPU cores.

Part of resolving issue #515

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: c
 matrix:
     include:
         - os: linux
+          compiler: gcc
           env:
             - DISTRO_TYPE=fedora
               INSTALL_REQUIREMENTS="dnf repolist; dnf install -y gcc python3 meson sudo langpacks-zh_CN ed ncurses vi findutils which"
@@ -13,6 +14,7 @@ matrix:
         # Note: Meson 0.44.0 is the newest version that works with Python
         # 3.4 available on OpenSuse Leap. See issue #430.
         - os: linux
+          compiler: gcc
           env:
             - DISTRO_TYPE=opensuse
               INSTALL_REQUIREMENTS="zypper refresh; zypper in -y gcc python3 python3-pip ninja sudo glibc-locale glibc-devel ed ncurses-utils vim findutils which; pip3 install meson==0.44.0"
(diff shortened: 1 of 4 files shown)


scripts/build-on-docker.sh
```diff
@@ -1,36 +1,47 @@
...
-# There is at least one open issue asking that this info be provided to the
-# client environment: https://github.com/travis-ci/travis-ci/issues/4696.
-# For now just hardcode the expected value.
-CORE_COUNT=2
-
 set -e
 export LANG=en_US.UTF-8
-export CFLAGS='-fno-strict-aliasing -Wno-unknown-pragmas -Wno-missing-braces -Wno-unused-result -Wno-return-type -Wno-int-to-pointer-cast -Wno-parentheses -Wno-unused -Wno-unused-but-set-variable -Wno-cpp -Wno-char-subscripts'
+export CC=gcc  # see the "compiler: gcc" line in the .travis.yml config file
+export CFLAGS="-fno-strict-aliasing -Wno-unknown-pragmas -Wno-missing-braces"
```
siteshwar: This code is used in macOS build script too. We should move code that is common to both scripts in a different file and source it.

.travis.yml
```diff
@@ -4,6 +4,7 @@ language: c
 matrix:
     include:
         - os: linux
+          compiler: gcc
```
siteshwar: We have to explicitly install `gcc` in docker build. This should not be needed.

krader1961: It's basically a no-op to make it clear to someone reading this config file that we want gcc for the Linux builds. It's to mirror the `compiler: clang` entry that isn't a no-op for the macOS environment. Yes, it's redundant but I think warranted in this case.

scripts/build-on-docker.sh
krader1961: Will do.

krader1961:
Merged after refactoring the common bits into a separate sourced script.

PR #613, opened 2018-06-21
Lint cleanup involving unused var

### LINKED ISSUE #607: Compiler lint warning identifies another bug in nvtree.c
This is another lesson in paying attention to all compiler warnings.

One of the modules implicated in nearly all the use-after-free failures is *src/cmd/ksh93/sh/nvtree.c*. @siteshwar made commit d9a8de212 to fix one of those. However, neither he nor I noticed his change was incomplete and left a bug in place. The compiler was kind enough to alert us to the problem:

```
../src/cmd/ksh93/sh/nvtree.c: In function ‘walk_tree’:
../src/cmd/ksh93/sh/nvtree.c:1054:31: warning: variable ‘odir’ set but not used [-Wunused-but-set-variable]
                 struct nvdir *odir = 0, *dp = (struct nvdir *)dir;
                               ^
```

The problem here is that the assignment to `odir` in that block is supposed to refer to the definition 27 lines earlier outside the `while() ...` loop. The redefinition within that block is a bug.

krader1961:
Fixes #607

### DIFF
--- a/src/cmd/ksh93/sh/nvtree.c
+++ b/src/cmd/ksh93/sh/nvtree.c
@@ -1029,7 +1029,9 @@ static_fn char *walk_tree(Namval_t *np, Namval_t *xp, int flags) {
     while (true) {
         cp = nv_dirnext(dir);
         if (odir) {
-            free(odir);
+            // TODO: Figure out if we should even be freeing the entity pointed to by `odir`.
+            // If we do free it one of the `treemove` tests that involves `typeset -m` fails.
+            // free(odir);
             odir = NULL;
         }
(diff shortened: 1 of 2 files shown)


krader1961:
With this change there are zero compiler warnings on all platforms I'm using: macOS, OpenSuse, Ubuntu, FreeBSD, OpenBSD, and Cygwin. Which means we can finally start paying attention to compiler warnings introduced by a change. In fact, we might want to consider treating warnings as errors that cause the build to fail.

siteshwar:
LGTM

PR #616, opened 2018-06-25
Remove SHOPT_REGRESS

### LINKED ISSUE #614: Eliminate SHOPT_REGRESS and the code predicated on it
While cleaning up the `#include` directives I noticed that we are not building with `SHOPT_REGRESS` defined true and it is not clear that would even work now. It's purpose is to fake things like the effective uid and gid. However, because it doesn't affect the underlying syscalls, and thus doesn't affect external commands, it is not clear to me that this mechanism is worth the cost. I vote we simplify the code by dropping SHOPT_REGRESS and the code predicated on it.

krader1961:
We haven't built with SHOPT_REGRESS defined true since the project was open
sourced. Looking at what this does it is not obvious the value justifies
the complexity. Quite the contrary. Given the availability of virtualized
OS environments it would be better to do the equivalent in a VM using
commands like `sudo` to actually create situations where euid != ruid.

Resolves #614


krader1961:
FYI, I am not opposed to tests that verify correct behavior when conditions like euid != ruid. I am simply opposed to the complexity, and incompleteness, of the unused mechanism this PR removes. If and when someone points out that ksh handles such situations incorrectly (which I predict has a probability close to zero) appropriate tests can be added.

PR #617, opened 2018-06-26
Enable building and running more unit tests

krader1961:
This deliberately removes the tmmap2read.c test because I can't be
bothered to make it work on all the platforms the project supports and
I don't want bin/lint warning about it not having a build rule.

It also adds a workaround for the literal "-NAN" being treated as positive
on some platforms. That's because I can't be bothered to figure out if
the problem is in libast or the system libraries since the distinction
should not matter for 99.9999% of shell scripts.


siteshwar:
LGTM

Issue #598, opened 2018-06-13
Test applications written in ksh through CI


siteshwar:
We should setup CI to test applications that are written in ksh. For e.g. kshdb.

rocky:
While you are waiting to hear back for other ksh programs to test, why not start with kshdb now since that's been a known source for finding ksh breakage. 

Working around what might be another ksh bug, (issue #610), I've just release 0.07 which currently passes all of its regression tests with the current git version of ksh, and SEGV's on Ubuntu's ksh93u+

Around 2011 kshdb worked and starting in 2012 or so it had been broken for the last 6 years. 

On the other hand, if there's no interest in having a debugger for ksh, that too would be useful to know so I won't spend time keeping this up to date. For POSIX shell debugging there still are the zsh and bash debuggers which by far have required less maintenance.

siteshwar:
I have setup a cron job to trigger these tests once in a day on latest copr builds.

rocky:
Many thanks. I greatly appreciate it. 

To reduce the possibility of changes to kshdb messing up ksh testing. I have just added tag  `att-ksh93v-testing` to the kshdb repository.

PR #618, opened 2018-06-26
Add travis configurations to test applications written in ksh

siteshwar:
Related: #598

### DIFF
--- a/app-compat-travis.yml
+++ b/app-compat-travis.yml
@@ -1,0 +1,22 @@
+# Test compatibility with applications written in ksh
+sudo: required
+
+services:
+    - docker
+
+matrix:
+    include:
+        - env:
+            - $PROJECT_NAME=kshdb
+              GIT_URL="https://github.com/rocky/kshdb.git"
+              COMMANDS="ksh ./autogen.sh; make test"
+
+script:
+    - docker pull fedora
+    - docker run fedora bash -c "
+      dnf install -y 'dnf-command(copr)';
+      dnf copr enable -y @ksh/latest;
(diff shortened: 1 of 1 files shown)


app-compat-travis.yml
```diff
@@ -0,0 +1,22 @@
+# Test compatibility with applications written in ksh
+sudo: required
+
+services:
+    - docker
+
+matrix:
+    include:
+        - env:
+            - $PROJECT_NAME=kshdb
```
krader1961: Is the leading dollar-sign supposed to be there?

app-compat-travis.yml
```diff
@@ -0,0 +1,22 @@
...
+              GIT_URL="https://github.com/rocky/kshdb.git"
+              COMMANDS="ksh ./autogen.sh; make test"
+
+script:
+    - docker pull fedora
+    - docker run fedora bash -c "
+      dnf install -y 'dnf-command(copr)';
+      dnf copr enable -y @ksh/latest;
```
krader1961: Just curious, but does this use the latest OBS build or something else? Perhaps a comment somewhere in this file to explain which version of ksh is being used would be a good idea.

krader1961:
LGTM modulo the extraneous dollar-sign.

app-compat-travis.yml
siteshwar: It uses builds from [copr](https://copr.fedorainfracloud.org/coprs/g/ksh/latest/). I added a comment mentioning it.

app-compat-travis.yml
siteshwar: Fixed!

PR #620, opened 2018-06-27
Remove coshell daemon code

### LINKED ISSUE #619: Remove the coshell facility

krader1961:
The coshell service is not secure. It's also not needed since most
companies today will have an equivalent, secure, mechanism. It also
isn't included by any distro AFAICT.

Resolves #619


siteshwar:
LGTM

PR #624, opened 2018-06-27
Elimimnate syscall intercepts

### LINKED ISSUE #622: AST intercept mechanism should be removed
The *ast_intercept.h* header and *src/lib/libast/port/intercept.c* module did not exist in the ksh93u+ release. Their presence causes many problems with respect to the order in which system and project headers are included. While working on resolving IWYU (include what you use) recommendations I regularly see errors about redefinition of functions like `faccessat()`.

Those "intercepts" also hide important implementation details to anyone looking at the code since someone looking at a function call like `chmod()` won't recognize it is really a call to `ast_chmod()`.

Too, those "intercepts" are now mostly (all?) pointless since we removed the 3DFS emulation implemented by `pathdev()` and friends. The only possibly useful thing they still do is to restart syscalls that are interrupted via the `RESTART()` macro. But note that ksh93u+ did not have this mechanism and it seems to be sufficiently robust with respect to signals.

Note that `RESTART()` invokes the `astserial()` function which appears to be broken, or at least incompletely implemented, because neither `AST_SERIAL_ENVIRON` or `AST_SERIAL_LOCALE` is used. And `AST_SERIAL_RESTART` is only used in *src/cmd/ksh93/sh/fault.c* which seems to be pointless on initial inspection. That's because if a signal results in `sh_fault()` being called and the syscall was interrupted without being restarted then there doesn't seem to be any need to inspect `&GLOBAL(restart)` (from the `astserial()` function). The return value of the syscall and `errno` will make it clear the syscall was interrupted.

I am no doubt missing the reasons why this mechanism was added. But it is opaque and the source of problems when trying to do simple tasks such as normalizing included headers.

krader1961:
Fixes #622


src/cmd/ksh93/sh/main.c
```diff
@@ -135,11 +134,6 @@ int sh_main(int ac, char *av[], Shinit_f userinit) {
     bool rshflag; /* set for restricted shell */
     char *command;
 
-#ifdef AST_SERIAL_RESTART
-    // Restart all ast_*() intercepted syscalls on EINTR.
-    astserial(AST_SERIAL_RESTART, AST_SERIAL_always);
-#endif
-    free(malloc(64 * 1024));
```
siteshwar: Strange! Not sure what this was intended for.

siteshwar:
LGTM

siteshwar:
Nitpick: Fix spelling of `eliminate` in commit message.

src/cmd/ksh93/sh/main.c
krader1961: I'm hoping it was something like an attempt to "prime the pump" so that subsequent Vmalloc calls would be a trifle faster. Regardless, it's obviously silly.

PR #625, opened 2018-06-27
A couple of uninitialized vars found by cppcheck

krader1961:


src/cmd/ksh93/sh/macro.c
```diff
@@ -2256,7 +2256,8 @@ static_fn void endfield(Mac_t *mp, int split) {
 //
 static_fn int substring(const char *string, size_t len, const char *pat, int match[], int flag) {
     const char *sp = string;
-    int size, nmatch, n;
+    int size, n;
```
siteshwar: While you are touching this code, why not initialize these variables too ?

krader1961: Because they aren't used before being set.

PR #628, opened 2018-06-28
Remove standalone commands

### LINKED ISSUE #615: What to do with the code in src/cmd/builtin
While working on the `#include` statements to make it clear which includes are system versus project specific, and fix IWYU warnings, I noticed that several modules are not being built. In fact, they will fail to build due to missing headers. For example, *src/cmd/builtin/uuencode.c* contains `#include <uu.h>`. That header used to be located at *src/lib/libuu/uu.h* but no longer exists since we eliminated that directory.

The only command in that directory that we are building is `pty`. What do we want to do about the other commands and files in that directory?

krader1961:
Fixes #615


src/cmd/builtin/pty.c
```diff
@@ -1,977 +0,0 @@
-/***********************************************************************
```
siteshwar: What do you plan to do with `pty` ? Currently it's being built as a standalone command. If we want to retain it, it does not make sense to remove this code.

krader1961: I had a second commit turning it into a builtin that I forgot to push. With a couple of small changes to block SIGCHLD so that the normal ksh job termination code didn't reap pty's child process it seemed to be working on macOS. Then this morning I tested it on OpenSuse and it fails in `_sffilbuf()` (from the `sfgetr()` call in *pty.c*). Which suggests there may be bugs in the code (quell surprise). So for the moment I'm going to reinstate it as a standalone command.

Perhaps we should reopen the question of whether we want to keep `pty` as the only other external command we build.

dannyweldon:
> So for the moment I'm going to reinstate it as a standalone command.

Indeed, I had wondered if it would work as a builtin and presumed that
was why it was only built as an external command.  Either way is fine
by me.

On 29 June 2018 at 09:01, Kurtis Rader <notifications@github.com> wrote:
> @krader1961 commented on this pull request.
>
> ________________________________
>
> In src/cmd/builtin/pty.c:
>
>> @@ -1,977 +0,0 @@
> -/***********************************************************************
>
> I had a second commit turning it into a builtin that I forgot to push. With
> a couple of small changes to block SIGCHLD so that the normal ksh job
> termination code didn't reap pty's child process it seemed to be working on
> macOS. Then this morning I tested it on OpenSuse and it fails in _sffilbuf()
> (from the sfgetr() call in pty.c). Which suggests there may be bugs in the
> code (quell surprise). So for the moment I'm going to reinstate it as a
> standalone command.
>
> Perhaps we should reopen the question of whether we want to keep pty as the
> only other external command we build.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 
Regards

Danny

src/cmd/builtin/pty.c
siteshwar: +1 for reopening the discussion to keep `pty` command.

src/cmd/builtin/RELEASE
```diff
@@ -1,156 +0,0 @@
-16-01-06 added dummy code in builtin.c to keep the linker from optimizing it out
```
siteshwar: We lost a bit of history here. This would be useful if we choose to keep `pty` command, but can look in legacy branches as an alternative.

krader1961: That was my thought. The history of this command isn't particularly useful so in the interest of eliminating useless cruft I just tossed the file. I'm hoping to convince people to drop the `pty` command so this really is focused on `ksh` and nothing else. Googling for mentions of this project's `pty` command I couldn't find any unambiguous mentions. The closest was these [course notes](https://www.ict.griffith.edu.au/anthony/info/co-processing/general_hints.txt) and it is far from obvious it is referring to the AST implementation.

krader1961:
Merged

PR #629, opened 2018-06-28
Remove remaining uses of `NoN` macro

krader1961:
It's not clear why this macro was ever needed but it is certainly no
longer needed. So remove its definition and the places that use it.


siteshwar:
LGTM

PR #632, opened 2018-06-29
Remove the unsed `strnacmp()` function

krader1961:
This function was present in the ksh93u+ branch but not used there
either. The top Google results are to this project. Nuke it.


siteshwar:
LGTM

PR #635, opened 2018-06-30
Rewrite /bin/style In ksh

etscrivner:
Currently, the /bin/style script is written to be executed in Fish shell. This
change converts it to using ksh instead, since it just seems right that shell
scripts for the ksh project should be runnable in ksh itself.

### DIFF
--- a/bin/style
+++ b/bin/style
@@ -1,6 +1,4 @@
-#!/usr/bin/env fish
-#
-# TODO: Rewrite this as a ksh script.
+#!/usr/bin/env ksh
 #
 # Usage: bin/style [--all] [source_filename]...
 #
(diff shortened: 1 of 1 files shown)


etscrivner:
I have tested all possible branching combinations of this script locally and confirmed that it works for each of them.

krader1961:
Thank you very much, @etscrivner. How well do you know Fish shell? I ask because the first statement in your version that caught my eye was this one:

```
for file in ${c_files[*]}
```

That will fail to produce the correct enumeration if any of the paths contains whitespace. The equivalent in fish doesn't need special quoting because it has saner variable expansion rules than POSIX 1003 and does not split on whitespace after the variable is expanded.

Try this:

```
f=('abc/def' 'ghi' 'j k l')
for file in ${f[*]}
do
    echo $file
done
```

You should use `for file in "${c_files[@]}"` to mimic the fish behavior.

krader1961:
Alternatively, do `IFS=$'\n'` at the top of the script.

etscrivner:
Thank you very much @krader1961 for the quick and thoughtful review. I am not very familiar with the semantics of fish shell or ksh (as you may have guessed), and appreciated the information and example you shared with respect to whitespace splitting.

I've updated the branch and tested each of the branching variations of the program against directories and files with one or more spaces in their names - some of them with multiple spaces - and have fixed the issues I found. I opted for the `IFS=$'\n'` method after initially trying to use `"${var[@]}"` during looping but finding it did not prevent errors in other places.

krader1961:
Looks pretty good but a couple of nits. I'm not a big fan of implicit tests for whether a var is set and a non-empty string like `[[ ${1} ]]`. This feature is a bogosity that POSIX adopted from the legacy Thompson/Bourne shell that results in bugs. Please explicitly test if it is set via `[[ -n ${1} ]]` or `[[ ${1} != "" ]]`. I prefer the latter but either is acceptable. The other nit is there are a few places you use the legacy single bracket test syntax: `if [ something ]`. Please use the double-bracket form.

etscrivner:
@krader1961 Updated and replaced all implicit tests with the preferred `[[ ${...} != "" ]]`. I've also replaced all single bracket test syntax with the double-bracket form.

bin/style
```diff
@@ -10,104 +8,136 @@
...
-            set files $files {$f}**.c {$f}**.h
-        end
-    end
+        exit
+    fi
+
+    files=( $(find src -name "**.h" -o -name "**.c") )
+elif [[ ${1} != "" ]]
+then
+    for next_file in $*
```
siteshwar: Nitpick: This can be shortened to `for next_file` but I think it's better to explicitly specifiy `$*`.

bin/style
```diff
@@ -10,104 +8,136 @@
...
+                c_files+=( "${file}" )
+            fi
+            ;;
+    esac
+done
 
-# Run the C reformatter if we have any C files.
-if not set -q c_files[1]
+# Run the C reformatter if we have any C files
+if [[ "${c_files}" == "" ]]
```
siteshwar: `if [[ "${#c_files[@]}" -eq 0 ]];`

siteshwar:
@etscrivner I would suggest you to scan all your shell scripts through [shellcheck](https://www.shellcheck.net/). It has given some warnings for this script that should be fixed.

```
Line 42:
    files=( $(git status --porcelain --short --untracked-files=all | sed -e 's/^ *[^ ]* *//') )
            ^-- SC2207: Prefer read -A or while read to split command output (or quote to avoid splitting).
 
Line 52:
    files=( $(find src -name "**.h" -o -name "**.c") )
            ^-- SC2207: Prefer read -A or while read to split command output (or quote to avoid splitting).
 
Line 55:
    for next_file in $*
                     ^-- SC2048: Use "$@" (with quotes) to prevent whitespace problems.
 
Line 59:
            files+=( ${next_file} )
                     ^-- SC2206: Quote to prevent word splitting, or split robustly with read -A or while read.
 
Line 62:
            files+=( $(find "${next_file}" -name "**.h" -o -name "**.c") )
                     ^-- SC2207: Prefer read -A or while read to split command output (or quote to avoid splitting).
 
Line 70:
    files=$(git diff-index --cached HEAD --name-only)
    ^-- SC2178: Variable was used as an array but is now assigned a string.
 
Line 71:
    files+=( $(git ls-files --exclude-standard --others --modified) )
             ^-- SC2207: Prefer read -A or while read to split command output (or quote to avoid splitting).
 
Line 78:
        files=$(git diff-tree --no-commit-id --name-only -r HEAD)
        ^-- SC2178: Variable was used as an array but is now assigned a string.
 
Line 96:
if [[ "${c_files}" == "" ]]
       ^-- SC2128: Expanding an array without an index only gives the first element.
 
Line 112:
        git add "${c_files}"
                 ^-- SC2128: Expanding an array without an index only gives the first element.
```

siteshwar:
In fact it will be a good idea to add shellcheck in our CI. I will open an issue to follow up on it.

etscrivner:
@siteshwar @krader1961 Thank you again for all the feedback. I've gone ahead and made the necessary modifications based on both of your comments and the output of `shellcheck`. The script now fully passes `shellcheck`.

The biggest change you will notice is based on the `SC2207` error given by shellcheck. I followed the recommended corrections on [the shellcheck page for the error](https://github.com/koalaman/shellcheck/wiki/SC2207).  This involved changing many commands from using an `array=( "$(command)" )` style to a `command | while IFS="" read -r line; do array+=("$line"); done` style.

I have also replaced all `${var} == ""` instances with `"${#var}" -eq 0` as recommended.

I have retested the script and think these changes have made it much more robust.

YvesCrespin:
*I have also replaced all ${var} == "" instances with "${#var}" -eq 0 as
recommended.*

Rather

 [[ "${#var}" -eq 0 ]]

Why not

 (( ${#var} == 0 ))

 ${#var} returns an integer.

Regards,

Yves

Le lun. 2 juil. 2018 20:15, Eric Scrivner <notifications@github.com> a
écrit :

krader1961:
FWIW, The "SC2207: Prefer read -A or while read to split command output" warning from shellcheck is better addressed by putting `IFS=$'\n'` at the top of the script. Something I consider a best practice. It allows `var=( $(some_command) )` to be used safely assuming `some_command` puts each value on a separate line. And that is far simpler and clearer than the read loop alternative.

etscrivner:
@krader1961 I agree that the previous style was simpler. I've updated the file so so that shellcheck warning SC2207 is disabled, leaving a comment below the line where this happens to indicate why it has been disabled. I've then reverted all of the `read` lines back to the previous, simpler `var=( $(some command) )` style.

krader1961:
Thanks for all the hard work, @etscrivner. There are still some minor problems with the rewrite but I've merged it anyway with a second commit to fix those issues. For example, the double asterisks in `find src -name "**.h" -o -name "**.c"` aren't needed. In fish they cause a a recursive glob. Merged.

etscrivner:
@krader1961 Thank you!

Issue #633, opened 2018-06-29
Investigate signal test failures caused by removal of vmalloc


siteshwar:
`signal` test fails consistently after removal of vmalloc and has been kept disabled. We should investigate why it fails and reenable this test.

siteshwar:
`read` builtin in `empty_fifos` function gets blocked because there is a minor difference between time when `read` should timeout and the time `sigalrm` gets invoked.

This is related code in `src/cmd/ksh93/sh/timers.c`:

```
107                 if (tp->action) {
108                     if (tp->wakeup <= now) {
109                         if (!tpold || tpold->wakeup > tp->wakeup) tpold = tp;
110                     } else {
```

From my gdb debugging session:

```
(gdb) p tp->wakeup
$1 = 1530278491.5283189

(gdb) p now
$2 = 1530278491.5183508

(gdb) p tp->wakeup <= now
$3 = 0
```

And `sigalrm` never gets called again.

krader1961:
In my PR #631 I noticed that the handling of SIGALRM is suspect. I didn't explicitly fix it but it is possible, even likely, that it fixes this specific failure mode. However, I enabled the `signal` test with that PR applied and it fails in new ways that are clearly suggestive of a use-after-free bug:

```
/tmp/ksh.signal.ypnl9dw/signal.sh: line 430: 37983: User signal 2
/tmp/ksh.signal.ypnl9dw/signal.sh: line 512: Èf: not found
/tmp/ksh.signal.ypnl9dw/signal.sh: line 512: ¢±^?: not found
```

siteshwar:
@krader1961 I suspect you are not hitting this race because you are on a slower system. I am able to reproduce it consistently even with changes from #631.

krader1961:
@siteshwar The speed of my system is definitely not the issue. My primary workstation is a lightly loaded latest generation Mac Pro tower with 12GB of memory and a quad core 2.7 GHz Xeon CPU 😈. I have absolutely no difficulty reproducing the problem on my VMware hosted Linux distros. I can't reproduce it on macOS. Which is a very strong indication this is due to the behavior of the system malloc implementation. Furthermore, on Linux I see messages like this logged immediately after the first time function `rttrap` is run in response to receiving one of the signals from a `kill -q`:

```
/tmp/ksh.signal.sFbLdMs/signal.sh: line 515: èF^Fh^N^?: not found
```

[...]

PR #636, opened 2018-07-01
Simplify create_svar()

krader1961:
While looking into issue #633 I realized the create_svar() function is
far more complicated than necessary. The var names do not, and can not,
contain any of the chars looked for by the loop that figures out how
long the var name is. The `goto` is also completely superfluous.

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -1608,7 +1608,7 @@ static const char *shdiscnames[] = {"tilde", 0};
 static_fn Namval_t *create_sig(Namval_t *np, const char *name, int flag, Namfun_t *fp) { return (0); }
 #endif
 
-struct Svars {
+typedef struct Svars {
     Namfun_t hdr;
     Shell_t *sh;
     Namval_t *parent;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/init.c
```diff
@@ -1631,30 +1630,22 @@ static_fn Namval_t *next_svar(Namval_t *np, Dt_t *root, Namfun_t *fp) {
...
-found:
-    if (nq) {
-        fp->last = (char *)&name[n];
-        shp->last_table = sp->parent;
-    } else {
-        errormsg(SH_DICT, ERROR_exit(1), e_notelem, n, name, nv_name(np));
+    assert(name);
+    for (int i = 0; i < sp->numnodes; i++) {
+        Namval_t *nq = nv_namptr(sp->nodes, i);
+        if (strcmp(name, nq->nvname) == 0) {
```
siteshwar: What if name contains an assignment ? I have seen pieces of code where assignments like `foo=bar` are passed as a single name parameter.

krader1961: That isn't an issue for this function. The sole use of this function is to assign values to the `.sh.sig.*` attributes. It's not used in any context where your scenario can happen. I'll add a doc-string comment at the top of the function.

PR #640, opened 2018-07-01
Remove SHOPT_COSHELL

### LINKED ISSUE #619: Remove the coshell facility

krader1961:
This removes libcoshell and the ksh code that depends on it. Since we
already removed the `coshell` server there is no need for ksh to be able
to talk to it.

Resolves #619


src/cmd/ksh93/meson.build
```diff
@@ -26,7 +25,6 @@ shared_c_args += [
     '-D_BLD_shell',
     '-DERROR_CONTEXT_T=Error_context_t',
     '-DKSHELL',
-    '-DSHOPT_COSHELL',
```
siteshwar: This option and related code was enabled in the last stable release. I suspect there are going to be side effects for removing this code.

krader1961: Obviously there shouldn't be any side-effects of removing, or simply disabling, this feature if is truly is optional as implied by the build time `SHOPT_COSHELL` symbol. Other than making it impossible to talk to a coshell server. And honestly a) I would be surprised if this feature still works in light of all the other changes, and b) I would be even more surprised if there is anyone using this feature.

I'm willing to change this to just build without the feature enabled by default plus the fix to `job_walk()` I mentioned in my previous comment and a deprecation warning in *CHANGELOG.md*. That way if a distro like Red Hat believes there are users of this facility they can still enable it.

siteshwar: I would prefer to just disable this feature instead of removing associated code.

krader1961:
Closing in favor of the more conservative #649.

PR #642, opened 2018-07-02
Remove SHOPT_ENV and the code predicated on it

### LINKED ISSUE #504: `SHOPT_ENV` should be removed by either fixing its code and making it the default or removing that code
The `SHOPT_ENV` flag was present in the old master branch (ksh93u+) but not enabled. Enabling `SHOPT_ENV` results in multiple, significant, compilation errors. Such as

```
../src/cmd/ksh93/sh/name.c:1417:84: warning: incompatible pointer types passing 'Env_t *' (aka 'void **') to parameter of type 'Shell_t *' (aka 'struct Shell_s *') [-Wincompatible-pointer-types]
            if (sp && ((flags & NV_EXPORT) || nv_isattr(np, NV_EXPORT))) sh_envput(shp->env, np);
                                                                                   ^~~~~~~~

../src/cmd/ksh93/sh/name.c:1862:25: error: expected ';' at end of declaration
    struct adata Shell_t *shp = (struct adata *)data->sh;
                        ^
                        ;

../src/cmd/ksh93/sh/name.c:1877:37: error: too few arguments to function call, expected 3, have 2
    sfputr(shp->stk, nv_name(np) - 1);
    ~~~~~~                          ^

../src/cmd/ksh93/../../lib/libast/include/sfio.h:253:1: note: 'sfputr' declared here
extern ssize_t sfputr _ARG_((Sfio_t *, const char *, int));
^
../src/cmd/ksh93/sh/name.c:1938:17: error: no member named 'stkme_envmarker' in 'struct Shell_s'
    sfputr(shp->stkme_envmarker, -1);
           ~~~  ^
```

The non-SHOPT_ENV code appears to be substantially the same from ksh93u+ to now. The SHOPT_ENV code is obviously a broken work in progress. Do we try to fix the broken code, even though we have no idea what the author's intent was, or just jettison it?

krader1961:
Note that this also removes code predicated on `#ifdef _ENV_H` and similar
constructs because those depend on `SHOPT_ENV`. That code is broken and not
enabled in either the ksh93u+ release or any subsequent version. It appears
to be either an experiment or a work in progress that was abandoned.

Resolves #504


etscrivner:
ACK for unit-tests and build locally.

I noticed that there is one more file reference and defining `_ENV_H` namely, `src/cmd/ksh93/include/env.h` which sets `_ENV_H` to 1 as part of its include guard.

The only place where this file is included appears to be `src/cmd/ksh93/sh/env.c`, which implements its functions. None of the `env_*` methods define here are used anywhere in the code base except `env_delete` which is also defined as a macro in `src/cmd/ksh93/include/defs.h`.

Following this investigation it seems safe to delete both `src/cmd/ksh93/include/env.h` and `src/cmd/ksh93/sh/env.c` - any reason for not doing so as part of this cleanup?

krader1961:
Thanks for taking a look, @etscrivner. The only reason I didn't remove those two files is I was tired and didn't notice they are now superfluous. As you can see from the open PRs I was in a bit of a monomaniacal mood trying to get as many cleanups done as possible over the weekend. And you're correct. Those two files can be removed. Done.

etscrivner:
@krader1961 Ah great, thought I might have missed something. 

I've confirmed the latest commit builds and passes unit-tests locally. Ran it a passed a few scripts through ksh and it seems to work just fine. Looks like a pretty straightforward removal of dead code.

👍

PR #644, opened 2018-07-03
Rewrite bin/lint In ksh

etscrivner:
Rewrite `bin/lint` from fish shell to ksh. Rewriting the `cppcheck` template
required approximating the fish shell colors using `tput` which may not map
well to other terminals, depending on how they are configured.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -1,8 +1,6 @@
-#!/usr/bin/env fish
+#!/usr/bin/env ksh
 #
-# TODO: Rewrite this as a ksh script.
-#
-# Usage: bin/lint [--all] [directory_or_filename]...
+# Usage: bin/lint [all | directory_or_filename...]
 #
 # Run the source through various lint detection tools. If invoked with `-all` then all the
 # src/cmd/ksh93 source files will be linted. If invoked with one or more path names they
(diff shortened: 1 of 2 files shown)


krader1961:
@etscrivner Thanks, again, for contributing. But FYI you should open an issue and indicate you plan on addressing the issue. I was going to apply the work you did on *bin/style* to *bin/lint* in the next day or two. So I'm glad to see you beat me to it before I did any actual work on rewriting that script. :smile: I'll review this tomorrow morning (i.e., in roughly 12 hours).

etscrivner:
@krader1961 Ahh, thank you for the info! Wasn't too sure on process and definitely happy to open up issues to signal future work - my apologies there.

Appreciate the reviews. Having a bit of fun with these so thought I'd knock this one out too. Thanks again 🙏

krader1961:
> Wasn't too sure on process...

Not everyone will agree with me but my advice is as follows.

Unless you are making a trivial change such as running the code through *bin/style* it is generally a good idea to open an issue that talks about the problem and your thoughts on how to fix it, if any. Then wait at least a few hours, if not a couple of days, for feedback before creating a pull request (PR). That way you don't risk wasting your time doing work that won't be accepted or duplicates work being done by others.

And, obviously, I was about to violate my own guidelines by converting *bin/lint* to a ksh script without opening an issue :smile: I should have expected you to follow up with a similar PR for *bin/lint*.

etscrivner:
@krader1961 Sounds perfectly reasonable to me as a way to coordinate. Will definitely do this in the future 😀

siteshwar:
LGTM

bin/lint
```diff
@@ -96,79 +114,93 @@ end
...
+        Darwin | FreeBSD | OpenBSD )
+            mapping_file="../etc/iwyu.bsd.map"
+            ;;
+        Linux | CYGWIN* )
+            mapping_file="../etc/iwyu.linux.map"
+            ;;
+    esac
+    for c_file in "${c_files[@]}"
+    do
+        if [[ "${#c_file}" -ne 0 ]]
```
krader1961: This should be `"$mapping_file" -ne ""`. I'll fixup when merging.

bin/lint
```diff
@@ -96,79 +114,93 @@ end
...
+        then
             include-what-you-use -Xiwyu --transitive_includes_only \
-                -Xiwyu --mapping_file=$mapping_file \
+                -Xiwyu --mapping_file="${mapping_file}" \
                 --std=c99 -Wno-expansion-to-defined -Wno-nullability-completeness \
-                $lint_args (../scripts/extract_flags $project_file $c_file) $c_file 2>&1
-        else  # hope for the best
+                "${lint_args[@]}" "$(../scripts/extract_flags "${project_file}" "${c_file}")" "${c_file}" 2>&1
```
krader1961: No quotes around the `../scripts/extract_flags` subcommand. That's because in this instance we need the values to be split into individual tokens. Otherwise IWYU gets a single arg as if you had typed `"-Idir1 -Idir2"` rather than two flags if the quotes weren't present. In other words, while it is often a good idea, or even required, to quote var expansion and subcommands sometimes you actually want the tokenizing after the expansion. I'll fix when merging.

krader1961:
merged

PR #645, opened 2018-07-03
Reenable signal tests on Linux

siteshwar:
- Disable a test that started failing after removing vmalloc.
- Increase timeout for read from 0.01 to 0.1 seconds to workaround a
  timing issue in sigalrm handler.

Related: #633

### DIFF
--- a/src/cmd/ksh93/tests/signal.sh
+++ b/src/cmd/ksh93/tests/signal.sh
@@ -1,7 +1,10 @@
-log_info "TODO: Enable this test when the bug documented in issue #396 is fixed."
-cd /tmp
-rm -rf $TEST_DIR
-exit 0
+if [[ "$(uname -s)" != "Linux" ]]
+then
+    log_info "TODO: Enable this test when the bug documented in issue #633 is fixed."
+    cd /tmp
+    rm -rf $TEST_DIR
+    exit 0
+fi
 
 ########################################################################
 #                                                                      #
@@ -151,7 +154,7 @@ actual=$(echo $(
 ))
 [[ $actual == $expect ]] ||
     log_error "subshell ignoring signal does not send signal to parent" "$expect" "$actual"
-read -u9 -t0.01 x || log_error 'parent unexpectedly consumed fifo go signal'
+read -u9 -t0.1 x || log_error 'parent unexpectedly consumed fifo go signal'
 
 expect='done SIGUSR1'
 actual=$(echo $(
@@ -162,7 +165,7 @@ actual=$(echo $(
 ))
 [[ $actual == $expect ]] ||
     log_error "subshell catching signal does not send signal to parent" "$expect" "$actual"
-read -u9 -t0.01 x || log_error 'parent unexpectedly consumed fifo go signal'
+read -u9 -t0.1 x || log_error 'parent unexpectedly consumed fifo go signal'
 
 # ====================
 # Verify exit due to signal can be mapped to the correct signal name.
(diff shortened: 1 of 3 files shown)


src/cmd/ksh93/tests/signal.sh
```diff
@@ -151,7 +154,7 @@ actual=$(echo $(
 ))
 [[ $actual == $expect ]] ||
     log_error "subshell ignoring signal does not send signal to parent" "$expect" "$actual"
-read -u9 -t0.01 x || log_error 'parent unexpectedly consumed fifo go signal'
+read -u9 -t0.1 x || log_error 'parent unexpectedly consumed fifo go signal'
```
krader1961: I have no idea why I worded that error message like that. If this `read` fails, due to timing out, it means the subcommand unexpectedly executed its `read x < fifo9` which isn't supposed to happen. Please replace `parent` with `subcommand`.

src/cmd/ksh93/tests/signal.sh
```diff
@@ -162,7 +165,7 @@ actual=$(echo $(
 ))
 [[ $actual == $expect ]] ||
     log_error "subshell catching signal does not send signal to parent" "$expect" "$actual"
-read -u9 -t0.01 x || log_error 'parent unexpectedly consumed fifo go signal'
+read -u9 -t0.1 x || log_error 'parent unexpectedly consumed fifo go signal'
```
krader1961: Here too s/parent/subcommand/.

src/cmd/ksh93/tests/signal.sh
krader1961: If increasing the timeout fixes the problem it means the alarm signal was processed and interrupted the read syscall and thus the read command. It means your system was slow enough that the kernel delivered the signal before the read was started. This contradicts the comment you added to `sigalrm()`. Did you really determine `sigalrm()` isn't getting invoked or was that just your best guess as to what was going wrong?

In fact, I can induce a hang at this point in the test by simply adding `sh_delay(0.01);` to `sh_readline()` after the `sh_timeradd()` call. So I'm okay with increasing the timeout to make the test more robust on heavily loaded systems. But I think the comment you added to *src/cmd/ksh93/sh/timers.c* should be removed.

krader1961: Got it. I misread the source comment. And I've confirmed via debug printf()'s that the problem is essentially as you say. What's interesting is that when the failure occurs `sigalrm()` is being run three orders of magnitude sooner than it should be run. On my system approximately 10 usec after `sh_timeradd()` returns rather than the expected 10,000 usec. Even more interesting sometimes a second SIGALRM is delivered while the read syscall is in progress.

The right thing to do is rewrite this to use `select()` or `poll()` rather than `alarm()` and `siglongjmp()`.

krader1961:
LGTM as a workaround. But as I wrote above the real solution isn't to fix whatever is wrong with `sh_timeradd()` and friends (although that should be done). The entire mechanism is inherently racy and should be replaced with `select()` or `poll()`.

krader1961:
@siteshwar I just applied this change to the master branch and it's passing all tests, including on my two macOS systems and on Travis. So I don't see any reason not to merge it.

siteshwar:
@krader1961 This test is consistently failing on macOS in Travis:

```
[1;31m207/242 sigchld/shcomp                          FAIL    66.71 s (exit status 1)[0m
```

krader1961:
I don't see that in your travis test history: https://travis-ci.org/siteshwar/ast/builds. You've got another, related, commit that is failing. This commit is fine. Note too that I've fixed the detection that `sigqueue` isn't available on macOS with PR#652.

krader1961:
@siteshwar I've gone ahead and merged this and my SI_USER fix (PR #652). As I wrote earlier I think you got confused by another branch you've been working on where sigchld is failing. I also added another commit that moves your `if [[ "$(uname -s)" != "Linux" ]]` to just keep the SIGINT test from being run on non-Linux systems since that is the only *signal.sh* test not passing on all platforms.

siteshwar:
@krader1961 If you look at the PRs in travis [here](https://travis-ci.org/att/ast/pull_requests), this pull request consistently failed in macOS builds, but that issue probably got fixed later by some other patch, so it's fine that you merged it.

PR #649, opened 2018-07-05
Build with coshell support disabled by default

### LINKED ISSUE #619: Remove the coshell facility

krader1961:
Since the coshell server was removed there should be no need for coshell
support in ksh. So build with it disabled by default but make it easy to
turn on if needed.

Resolve #619

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -76,6 +76,10 @@ shared_c_args = [
     '-D_AST_no_spawnveg=1',
 ]
 
+if get_option('build-api-tests') == true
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -76,6 +76,10 @@ shared_c_args = [
     '-D_AST_no_spawnveg=1',
 ]
 
+if get_option('build-api-tests') == true
```
siteshwar: Shall it be `coshell` instead of `build-api-tests` ?

krader1961: Yes, fixed

PR #650, opened 2018-07-05
Move detecting max signum to the build config phase

### LINKED ISSUE #663: What is preprocessor symbol `_NEXT_SOURCE`?
While looking into the purpose of the modules *astintercept.c*, *getenv.c*, and *setenviron.c* I noticed the `_NEXT_SOURCE` preprocessor symbol. It's not defined anywhere. Including in the legacy beta and master branches. Here are the sum total of the mentions:

```
src/cmd/ksh93/bltins/sleep.c:#ifdef _NEXT_SOURCE
src/cmd/ksh93/bltins/sleep.c:#endif  // _NEXT_SOURCE
src/cmd/ksh93/sh/main.c:#if !defined(_NEXT_SOURCE)
src/cmd/ksh93/sh/main.c:#else  // _NEXT_SOURCE
src/cmd/ksh93/sh/main.c:#endif  // _NEXT_SOURCE
src/cmd/ksh93/sh/main.c:#if !defined(_NEXT_SOURCE)
src/cmd/ksh93/sh/main.c:#endif  // _NEXT_SOURCE
src/cmd/ksh93/sh/name.c:#ifndef _NEXT_SOURCE
src/cmd/ksh93/sh/name.c:#endif  // _NEXT_SOURCE
```

Looks like cruft. Remove it.

krader1961:
The ksh code has been using a very opaque (i.e., hard to understand)
block of code to detect the max signal number. Replace that with a much
easier to understand build time feature detection.

AFAICT the run time calculation was correct. At least empirical testing
on a bunch of platforms (BSD and Linux) yield results identical to the
new mechanism. But I'm making this change anyway since a) there is no
point figuring out the answer every time ksh starts and b) it's much
clearer to do it using signal.h header symbols.

Partially resolves #663

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -74,6 +74,7 @@
 #mesondefine _stk_down
 #mesondefine isnanl
 #mesondefine const_const_fts_open
+#mesondefine MAX_SIGNUM
 
 // What follows are all the symbols that should be based on Meson feature
 // tests like `cc.has_header()` and `cc.find_library()`.
--- a/features/max_signum.c
+++ b/features/max_signum.c
@@ -1,0 +1,22 @@
+// This is used during the Meson config step to detect the largest signal number.
+#include <signal.h>
+#include <stdio.h>
+
+int main() {
+#ifdef _NSIG
+    printf("%d\n", _NSIG - 1);
+    return 0;
+#else  // _NSIG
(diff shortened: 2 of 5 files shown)


siteshwar:
> Partially resolves #663

Should be #633

PR #651, opened 2018-07-05
Add test case for ${parameter=word} expansion

siteshwar:


src/cmd/ksh93/tests/attributes.sh
```diff
@@ -269,8 +269,16 @@ else
     log_error 'typeset -usi cannot be used for unsigned short'
 fi
 
+# Check ${parameter:-word} style of parameter expansion. ':' is optional and is used to check for null (empty string).
+[[ $($SHELL -c 'unset foo;typeset -Z2 foo; print ${foo-3}' 2> /dev/null) == 3 ]]  || log_error  '${foo-3} not 3 when typeset -Z2 field undefined'
```
krader1961: I know you're just copying the style of the existing tests but the stderr redirection isn't needed. And if an error is emitted it would be nice to see it. What I don't understand is why the person who wrote the original tests felt the need to redirect stderr.

krader1961:
LGTM modulo the nit about the unnecessary stderr redirection.

krader1961:
LGTM modulo the nit about the unnecessary stderr redirection.

siteshwar:
@krader1961 Sorry for late response. I disappeared due to holidays. Merging it now with requested fix.

PR #656, opened 2018-07-07
Solve the definition of LIBICONV_PLUG

catull:
The way LIBICONV_PLUG is defined must go through feature detection, thus it must first land in config_ast.h.
This way, <iconv.h> can be included normally, provided that the source file first includes "config_ast.h".

This is already the case in all C source files.

This solves issue #280 in a more robust manner.

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -41,6 +41,7 @@
 #mesondefine ALIGN_BOUND
 #mesondefine ALIGN_BOUND1
 #mesondefine ALIGN_BOUND2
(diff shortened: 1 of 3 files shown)


krader1961:
LGTM, merged, thx. Now all I need to do is figure out how to motivate people to fix all the other chicken/egg include ordering. For example, there are quite a few headers that define private macros, such as `_NV_PRIVATE`, which require a specific include order rather than just doing a (mostly) alphabetical ordering of the headers.

PR #660, opened 2018-07-09
Fix bug in hist_exceptf()

### LINKED ISSUE #659: Bug in hist_exceptf()
Coverity Scan was nice enough to point out a bug:

```
*** CID 301310:  API usage errors  (USE_AFTER_FREE)
/src/cmd/ksh93/edit/history.c: 978 in hist_exceptf()
972             }
973             if (sh_fcntl(newfd, F_DUPFD_CLOEXEC, oldfd) != oldfd) {
974                 close(newfd);
975                 return -1;
976             }
977     
>>>     CID 301310:  API usage errors  (USE_AFTER_FREE)
>>>     Passing closed handle "oldfd" as an argument to "fcntl".
978             fcntl(oldfd, F_SETFD, FD_CLOEXEC);
979             close(newfd);
980             if (lseek(oldfd, (off_t)0, SEEK_END) < hp->histcnt) {
981                 int index = hp->histind;
982                 lseek(oldfd, (off_t)2, SEEK_SET);
983                 hp->histcnt = 2;
```

The warning is misleading. The use of `oldfd` in this case is fine because the preceding

```
if (sh_fcntl(newfd, F_DUPFD_CLOEXEC, oldfd) != oldfd)
```

guarantees that it will be an alias for `newfd` if it succeeds. It does, however, point us to a potential bug. If you assume no gaps in the range of open file descriptors then just prior to that attempt to duplicate `newfd` it would equal `oldfd`. Thus making it impossible for the fcntl call to return oldfd; it would have to return a fd greater than `oldfd`. Thus making the test in the above statement true causing the function to fail. Now in practice it is likely to work but only because the the history fd will normally be 11 and 10 is open on the CWD. There will normally be unused fds between 3 and 9 inclusive. But if the user had done `exec 3>file_a`, `exec 4>file_b`, etc. up thru fd 9 then this function would fail.

krader1961:
Make sure `newfd` does not equal `oldfd` since it would cause the subsquent
`if (fcntl(new_fd, F_DUPFD, oldfd) != oldfd)` to be true if it did.

Fixes #659

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -960,11 +960,10 @@ static int hist_exceptf(Sfio_t *fp, int type, Sfdisc_t *handle)
     if (type == SF_WRITE) {
         if (errno == ENOSPC || hp->histwfail++ >= 10) return 0;
         // Write failure could be NFS problem, try to re-open.
-        int oldfd = sffileno(fp);
-        sh_close(oldfd);
-
         int newfd = open(hp->histname, O_BINARY | O_APPEND | O_CREAT | O_RDWR | O_CLOEXEC,
                          S_IRUSR | S_IWUSR);
+        int oldfd = sffileno(fp);
+        sh_close(oldfd);
         if (newfd == -1) {
             errormsg(SH_DICT, 2, "History file write error-%d %s: file unrecoverable", errno,
                      hp->histname);


siteshwar:
Is [this line](https://github.com/krader1961/ast/blob/8ad6279ad1bb44477b67d9a6ff5fa55cd9cab958/src/cmd/ksh93/edit/history.c#L977) still needed ?

krader1961:
Yes if the system doesn't support `F_DUPFD_CLOEXEC`. See src/cmd/ksh93/include/defs.h*. That `fcntl` is going to be redundant on most platforms. This pattern exists elsewhere so it's something that should be cleaned up in a methodical manner. For now I'd prefer to leave that statement in the code since it isn't really related to the purpose of this change.

siteshwar:
LGTM

PR #661, opened 2018-07-09
Simplify code and address Coverity Scan warnings

krader1961:
Coverity Scan was warning about `sh_open()` potentially accessing bytes
past the end of a pathname. Many, possibly all, of those warnings were
false positives. Nonetheless, the way the code is structured it could
theoretically access a byte past the end of the pathname string. Perhaps
more importantly the code is hard to read. It's not at all obvious it's
checking for "/dev/fd/" for example.

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -606,8 +606,7 @@ int sh_open(const char *path, int flags, ...) {
     mode = (flags & O_CREAT) ? va_arg(ap, int) : 0;
     va_end(ap);
 
-    errno = 0;
-    if (path == 0) {
+    if (path == NULL) {
         errno = EFAULT;
         return -1;
     }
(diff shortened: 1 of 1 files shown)


dannyweldon:
That code is certainly much cleaner and easier to understand.  Do you think there will be much of a performance hit?  Are modern compilers able to optimise this at all now?  Or do you think the old code was needlessly hand-optimised anyway?

What about?

    if (path[0] == '/' && strncmp(path, "/dev/", sizeof("/dev/") - 1) == 0)

It's no big deal but I thought I would ask.

src/cmd/ksh93/sh/io.c
```diff
@@ -606,44 +606,28 @@ int sh_open(const char *path, int flags, ...) {
...
-                            if (path[9] == 'u' && path[10] == 't' && !path[11]) fd = 1;
-                            break;
-                        }
-                    }
-                }
-            }
+
+    if (strncmp(path, "/dev/", sizeof("/dev/") - 1) == 0) {
+        if (strncmp(path, "/dev/fd/", sizeof("/dev/fd/") - 1) == 0) {
+            if (flags == O_NONBLOCK) return (1);
```
siteshwar: `return (1)` -> `return 1`

src/cmd/ksh93/sh/io.c
```diff
@@ -606,44 +606,28 @@ int sh_open(const char *path, int flags, ...) {
...
+            fd = (int)strtol(path + sizeof("/dev/fd/") - 1, &e, 10);
+            if (*e) fd = -1;
+        } else if (strcmp(path, "/dev/stdin") == 0) {
```
siteshwar: Do we have a test case for this code ? If not, we should try to write tests for each code that we refactor.

krader1961:
@dannyweldon I think the original version was a needless optimization. But, since you bring it up, we can avoid this block of code on most platforms ksh runs on since they already support those device names. This would only be needed on platforms which do not have them. So I'll add a build time feature test and predicate this on the system not having those device nodes.

Too, if you look at the rest of the function you'll see it does attempt to open the */dev* path and uses the returned fd if successful. It only uses the value assigned to `fd` by this block of code if that open fails. I would have done an even more aggressive refactoring to do these tests after the `open()` failed but I wanted to keep the change low risk.

@siteshwar I agree in general about ensuring we have unit tests that exercise the code being changed. But note that per my previous paragraph we can't actually test this because all the platforms we're currently testing on support those magic */dev* paths natively. So this emulation code doesn't have any effect.

krader1961:
I'm going to close this because adding feature tests and refactoring the code for readability and add unit tests is going to take a few days. I also want to rebase this on other changes already committed or that I just opened a PR for such as #670.

PR #662, opened 2018-07-09
Mark statements not reached

krader1961:
Some lint warnings are due to statements like

errormsg(SH_DICT, ERROR_exit(1), e_nospace);

Those never return because they perform a longjmp() to return control
to the execution engine. But only if the value passed to ERROR_exit()
is not zero. Adding __builtin_unreachable() after those statements tells
the compiler and lint tools subsequent statements aren't reachable.
Thus improving their ability to determine, for example, whether or not
a null pointer might be dereferenced.

Long term I'd prefer to modify the error logging API to separate the
calls that can return from those that cannot.


siteshwar:
LGTM

PR #669, opened 2018-07-09
Add test cases for parameter expansion modifiers

siteshwar:
- Add tests for ${:=}, ${:-}, ${:?} and ${:+} style of parameter expansions.

- Move some tests to parameter expansion modifiers test file.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -95,7 +95,7 @@ all_tests = [
     ['basic', 90], ['bracket'], ['builtins'], ['case'], ['comvar'],
     ['comvario'], ['coprocess', 50], ['cubetype'], ['directoryfd'], ['enum'],
     ['exit'], ['expand'], ['functions'], ['glob'], ['grep'], ['heredoc'],
-    ['io'], ['locale'], ['math', 50], ['nameref'], ['namespace'],
+    ['io'], ['locale'], ['math', 50], ['nameref'], ['namespace'], ['modifiers'],
     ['options'], ['path'], ['pointtype'], ['pty'], ['quoting'], ['quoting2'],
     ['readcsv'], ['recttype'], ['restricted'], ['return'], ['select'],
     ['sh_match'], ['sigchld', 100], ['signal'], ['statics'], ['subshell', 100],
--- a/src/cmd/ksh93/tests/modifiers.sh
+++ b/src/cmd/ksh93/tests/modifiers.sh
@@ -1,0 +1,46 @@
+# These are the tests for various style of parameter expansion modifiers supported by ksh93
+
+# Check ${parameter:-word} style of parameter expansion. ':' is optional and is used to check for null (empty string).
+unset foo
+[[ ${foo:-bar} == bar ]]  || log_error  '${foo:-bar} not bar when foo is not set'
+
+unset foo
+[[ ${foo-bar} == bar ]]  || log_error  '${foo-bar} not bar when foo is not set'
+
+foo=""
+[[ ${foo-bar} == "" ]]  || log_error  '${foo-bar} not "" when foo is null'
+
+# Check ${parameter:=word} style of parameter expansion. ':' is optional and is used to check for null (empty string).
+unset foo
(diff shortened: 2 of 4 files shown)


src/cmd/ksh93/meson.build
```diff
@@ -95,7 +95,7 @@ all_tests = [
     ['basic', 90], ['bracket'], ['builtins'], ['case'], ['comvar'],
     ['comvario'], ['coprocess', 50], ['cubetype'], ['directoryfd'], ['enum'],
     ['exit'], ['expand'], ['functions'], ['glob'], ['grep'], ['heredoc'],
-    ['io'], ['locale'], ['math', 50], ['nameref'], ['namespace'],
+    ['io'], ['locale'], ['math', 50], ['nameref'], ['namespace'], [ 'modifiers' ],
```
etscrivner: Small style nit: Spacing between the square brackets when everywhere else there is no spacing.

src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -0,0 +1,42 @@
...
+[[ ${foo:=bar} == bar ]]  || log_error '${foo:=bar} not bar when foo is not set'
+
+unset foo
+[[ ${foo=bar} == bar ]]  || log_error '${foo=bar} not bar when foo is not set'
+
+foo=""
+[[ ${foo=bar} == "" ]]  || log_error  '${foo=bar} not "" when foo is null'
+
+# Check ${parameter:?word} style of parameter expansion. ':' is optional and is used to check for null (empty string).
+[[ $( $SHELL -c 'unset foo; print ${foo:?bar not set}' 2>&1 ) =~ "bar not set" ]]  || log_error '${foo:?bar} does not display error if foo is unset'
```
krader1961: FYI, for complex situations like this I prefer this pattern so that the test log provides more useful data:

```
expect="bar not set"
actual=$( $SHELL -c 'unset foo; print ${foo:?bar not set}' 2>&1 )
if [[ $actual =~ $expect ]] || log_error 'message' "$expect" "$actual"
```

krader1961:
LGTM modulo the two minor nits.

src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -0,0 +1,42 @@
...
+
+[[ $( $SHELL -c 'unset foo; print ${foo?bar not set}' 2>&1 ) =~ "bar not set" ]]  || log_error '${foo?bar} does not display error if foo is unset'
+
+[[ $( $SHELL -c 'foo=""; print ${foo?bar}' 2>&1 ) == "" ]]  || log_error  '${foo=bar} not "" when foo is null'
+
+# When nothing is specified after :?, a default error message is printed
+unset foo
+[[ $( (print ${foo:?}) 2>&1) =~ "parameter not set" ]] || log_error 'Incorrect error message with ${foo?}'
```
etscrivner: Error message should say `...with ${foo:?}` instead of `with ${foo?}`.

src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Check ${parameter:+word} style of parameter expansion. ':' is optional and is used to check for null (empty string).
+unset foo
+[[ ${foo:+bar} == "" ]]  || log_error '${foo:+bar} not null when foo is not set'
+
+unset foo
+[[ ${foo+bar} == "" ]]  || log_error '${foo+bar} not null when foo is not set'
+
+foo="non-null value"
+[[ ${foo+bar} == "bar" ]]  || log_error  '${foo+bar} not bar when foo is not null'
```
etscrivner: Is it worth testing the `:+` variation when foo is not null as well?

etscrivner:
Is it worth adding the remaining variations `${#}`, `${##}`, `${%}`, `${%%}`, `${x||pattern1||pattern2}`?

etscrivner:
LGTM other than nits and questions.

src/cmd/ksh93/tests/modifiers.sh
siteshwar: I added a test case for it.

etscrivner:
Updates LGTM

PR #670, opened 2018-07-10
Remove symbol O_SERVICE

krader1961:
It's sole purpose was to allow the `mkservice` command in the
src/cmd/ksh93/bltins/mkservice.c module to open a network endpoint in
listening mode. Since we removed that command the O_SERVICE symbol is no
longer needed.

I noticed this while working on PR #661 to eliminate Coverity Scan warnings
about reading past the end of the path name string. I decided to fix this
in a separate PR since PR #661 is already larger than I would like.

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -80,10 +80,6 @@
 #endif  // !O_NONBLOCK
 #endif  // FNDELAY
 
-#ifndef O_SERVICE
-#define O_SERVICE O_NOCTTY
-#endif
-
 #ifndef ERROR_PIPE
 #ifdef ECONNRESET
 #define ERROR_PIPE(e) ((e) == EPIPE || (e) == ECONNRESET || (e) == EIO)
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -628,14 +616,12 @@ int sh_open(const char *path, int flags, ...) {
             fd = 2;
         }
 
-#ifdef O_SERVICE
         if (fd < 0) {
             if ((fd = inetopen(path + 5, flags, onintr, shp)) < 0 && errno != ENOTDIR) return -1;
```
siteshwar: This looks like specific to `mkservice` builtin.

krader1961: Yes it does look like that and it's wrong :smile: The `inetopen()` function is what implements support for pathnames as in this example: `echo hello >/dev/tcp/localhost/1234`. Personally I would just remove it as I don't think ksh is the right tool for network communication. But that would break backward compatibility with ksh93u+. The only part of this change that is specific to `mkservice` is the `int server` var and the code predicated on it being true. That I did remove since `server` can never be true now that we've removed `mkservice`.

krader1961:
Merged

PR #672, opened 2018-07-11
Add tests for special /dev/... path names

krader1961:
I intend to refactor `sh_open()` to a) fix lint warnings about accessing
bytes past the end of the pathe string and b) improve readability. Before
doing so it would be nice if there were tests of the existing behavior.

These tests are necessarily incomplete because we don't (yet) have a way to
force on or off the emulation code or pretend that paths like /dev/stdin
are not actually supported by the platform. This is a small step towards
creating a comprehensive set of tests for the magic /dev/... path names.

### DIFF
--- a/src/cmd/ksh93/tests/special-dev-paths.sh
+++ b/src/cmd/ksh93/tests/special-dev-paths.sh
@@ -1,0 +1,171 @@
+#
+# Verify that the special /dev paths such as /dev/stdin, /dev/fd/0, and /dev/tcp/host/port work when
+# used in a test and actually opened. This is primarily meant to exercise the `sh_open()` functions
+# handling of special paths.
+#
+readonly host=localhost
+readonly port=65123
+
+fd_check=no
+
+# Some tests will want to verify whether or not a fd was allocated. We could use the `lsof` command
+# but don't want to add that as a dependency. Instead just do the verification on platforms which
+# support /proc/self/fd.
+[[ -d /proc/self/fd ]] && fd_check=yes
+[[ $fd_check == yes ]] && fd_count=$( ls -1 /proc/$$/fd | wc -l )
+
+# Check if the `nc` (netcat) command is available.
+whence -q nc && readonly nc_available=yes || readonly nc_available=no
+
+# Check if the system supports the /dev/{stdin,stdout,stderr} paths.
+[[ -e /dev/stdin ]] && readonly dev_stdin=yes || readonly dev_stdin=no
+
+# ==========
+# Verify that testing variations of the special dev paths work as expected.
+# Note that these should not result in a fd being opened.
+[[ -e /dev/stdin ]] || log_error '-e /dev/stdin failed'
+[[ -e /dev/stdout ]] || log_error '-e /dev/stdout failed'
+[[ -e /dev/stderr ]] || log_error '-e /dev/stderr failed'
+[[ -e /dev/stdinX ]] && log_error '-e /dev/stdinX succeeded but should have failed'
+[[ -e /dev/stdoutX ]] && log_error '-e /dev/stdoutX succeeded but should have failed'
+[[ -e /dev/stderrX ]] && log_error '-e /dev/stderrX succeeded but should have failed'
+
+[[ -e /dev/fd/0 ]] || log_error '-e /dev/fd/0 failed'
+# TODO: This test currently fails. It thinks the bogus path is valid.
+# [[ -e /dev/fd/xxx ]] && log_error '-e /dev/fd/xxx succeeded but should have failed'
+
+[[ -e /dev/tcp/localhost/666 ]] || log_error '-e /dev/tcp/localhost/666 failed'
+[[ -e /dev/TCP/localhost/666 ]] && \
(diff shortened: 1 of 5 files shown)


krader1961:
Note that I tried installing `netcat`, `netcat-gnu`, and `netcat-openbsd` on Fedora. Happy to add the correct package if you can tell me it's name, @siteshwar.

siteshwar:
@krader1961 That's `nmap-ncat`.

src/cmd/ksh93/tests/special-dev-paths.sh
```diff
@@ -0,0 +1,157 @@
...
+[[ -e /dev/stderr ]] || log_error '-e /dev/stderr failed'
+[[ -e /dev/stdinX ]] && log_error '-e /dev/stdinX succeeded but should have failed'
+[[ -e /dev/stdoutX ]] && log_error '-e /dev/stdoutX succeeded but should have failed'
+[[ -e /dev/stderrX ]] && log_error '-e /dev/stderrX succeeded but should have failed'
+
+[[ -e /dev/fd/0 ]] || log_error '-e /dev/fd/0 failed'
+# TODO: This test currently fails. It thinks the bogus path is valid.
+# [[ -e /dev/fd/xxx ]] && log_error '-e /dev/fd/xxx succeeded but should have failed'
+
+[[ -e /dev/tcp/localhost/666 ]] || log_error '-e /dev/tcp/localhost/666 failed'
```
siteshwar: This may be expected behavior, but why should this test succeed ?

krader1961: The original test in *brackets.sh* simply verified that `/dev/tcp/` is valid. Which could be valid or invalid depending on how you look at it. You can see from the other tests I've included but commented out with a "TODO" that I consider it undesirable because you can't actually do `echo hello >/dev/tcp` or `echo bye >/dev/tcp/localhost`. On the other hand if you treat this as if the */dev/tcp* directory actually existed and hostnames formed another layer of directories then `[[ -e /dev/tcp ]]` arguably should be true. Too, we don't want to go to the expense of validating the hostname is valid and we can validate the port number only to the extent it is in the range 0 to 65535. But we should be verifying that `/dev/tcp/localhost/1234/abc` is invalid.

krader1961: > This may be expected behavior, but why should this test succeed ?

I see I did not directly answer this question with my previous reply. That path is theoretically valid and thus the `[[ -e ... ]]` test should be true. The test cannot verify the path is actually valid since that can only be done by establishing a network connection. Something we don't want to do because the side-effects are undesirable. This is behavior that is insufficiently documented. What does it even mean to test if a */dev/tcp/...* path exists? Should it validate the path is valid if actually used to open a connection? Or should it just validate the path has the proper form?

siteshwar: Ok, it would be good to mention these assumptions in comments with this test case.

src/cmd/ksh93/tests/special-dev-paths.sh
```diff
@@ -0,0 +1,159 @@
...
+# used in a test and actually opened. This is primarily meant to exercise the `sh_open()` functions
+# handling of special paths.
+#
+readonly host=localhost
+readonly port=65123
+
+# Some tests will want to verify whether or not a fd was allocated. We could use the `lsof` command
+# but don't want to add that as a dependency. Instead just do the verification on platforms which
+# support /proc/self/fd.
+[[ -d /proc/self/fd ]] && readonly fd_check=yes || readonly fd_check=no
```
siteshwar: Probably it would be bettter to just declare variables like `fd_check` with default value of `no` at top and then change it to `yes` if this test succeeds.

src/cmd/ksh93/tests/special-dev-paths.sh
```diff
@@ -0,0 +1,159 @@
...
+fi
+
+# ==========
+# Verify that opening the special /dev/tcp/host/service path works.
+# This requires the netcat program.
+if [[ $nc_available == no ]]
+then
+    log_info 'Skipping the /dev/tcp/... I/O tests because nc is not available'
+else
+    nc -l $host $port >nc.out &
```
siteshwar: This test can not be executed in parallel. It is failing on Travis with this error:

```
Ncat: bind to 127.0.0.1:65123: Address already in use. QUITTING.
```

src/cmd/ksh93/tests/special-dev-paths.sh
krader1961: Will do.

src/cmd/ksh93/tests/special-dev-paths.sh
krader1961: Agreed. I was just trying a different idiom to see how it looked/worked.

src/cmd/ksh93/tests/special-dev-paths.sh
krader1961: Doh! Of course. I'll blacklist this one to run serially.

Issue #673, opened 2018-07-11
Enable ASAN when running unit tests


krader1961:
I was reviewing open issues and noticed #403 which talks about testing for memory leaks. A comment in it mentions that if we get ASAN working we shouldn't need an explicit memory leak unit test. Since we are now using the native, rather than AST, malloc implementation I decided to try enabling ASAN when running unit tests. It was more work than I expected and currently only works with gcc, not clang. Building with clang on macOS results in undefined ASAN symbols which I haven't figured out how to deal with.

On OpenSuse 42.3 there are 45 test failures with ASAN enabled and zero with it disabled. There are also 10 timeouts. This is with API tests disabled because, for some unknown reason, most of them time out with ASAN enabled. Even with a huge timeout multiplier. Also, some tests fail without any ASAN errors when ASAN is enabled; e.g., `namespace/C` and `sh_match/C`. The ASAN failure categories break down thusly:

```
     15 global-buffer-overflow
      9 heap-buffer-overflow
     15 heap-use-after-free
```

It is highly probably that if we fix the causes of these ASAN failures many other failures, including intermittent failures of the `types` unit test, will disappear. Note that ASAN reports a `heap-buffer-overflow` for the `types` test:

```
==90338== ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602c0000ff73 at pc 0x5683f0 bp 0x7ffcb3c097d0 sp 0x7ffcb3c097c8
READ of size 1 at 0x602c0000ff73 thread T0
    #0 0x5683ef in create_type /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/nvtype.c:510
    #1 0x5574c7 in create_tree /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/nvtree.c:104
    #2 0x5375d6 in nv_create /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/name.c:1035
    #3 0x539aa2 in nv_open /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/name.c:1254
    #4 0x55cf96 in outval /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/nvtree.c:707
    #5 0x55fb1e in genvalue /home/krader/projects/3rd-party/ast/build/../src/cmd/ksh93/sh/nvtree.c:899
[...]

PR #674, opened 2018-07-11
Enable support for ASAN

krader1961:
This is a proof of concept change. It has problems such as not supporting
clang and presumes the gcc version is less than 4.9.3.  Nonetheless,
it provides crucial insights into the many use-after-free and related
bugs in the code.

See issue #673


krader1961:
Obviously this PR does not enable ASAN in the Travis builds because it introduces a huge number of failures. The intent is simply to make it possible for individuals to leverage ASAN when testing on their local machine(s).

siteshwar:
LGTM

PR #675, opened 2018-07-11
Fix reference past end of malloced buffer

krader1961:
ASAN brought this to our attention.

If var `name` points to a string like "schedule.job" then r == 8 (the
length of the first token). If the string pointed to by `nq->nvname` is
shorter than that then `nq->nvname[r]` will read a location past the end
of the buffer. This is yet another case where the person writing the
original version made a dangerous optimization.


krader1961:
This may fix the sporadic flakey failure of the `types` unit test on BSD/macOS. If the buffer we're indexing past the end of ends at or near a page boundary and there is no contiguous page the invalid access fixed by this PR would result in a SIGSEGV.

siteshwar:
LGTM

etscrivner:
LGTM

Issue #676, opened 2018-07-12
nv_putval() is calculating the wrong size of the buffer/string to move


krader1961:
When I started working on this project a year ago one of the first things I did was try to build and run it on macOS which is largely based on BSD. One of the first hard failures I saw was the `memcpy()` in `nv_putval()` failing because the buffers overlapped. Not knowing at the time if the buffers could legitimately overlap I made the simple fix and replaced `memcpy()` with `memmove()`.

Fast forward to today when we now have ASAN to help us find problems. It reported that the `memmove()` sometimes accesses memory beyond the end of the buffer pointed to by `sp`. That happens because `dot` at that junction is from `nv_size(np)` which returns 12 and the `sp` buffer is typically 9 bytes long.

Note that even though the `sp` buffer is 9 bytes long it contains a single character; i.e., a string of length one. There is a magic `strlen() + 8` involved in the allocation of the buffer. What the magic `8` means is anyone's guess. It wouldn't surprise me if it was added solely to deal with the `memcpy()` size argument being too large.

I am now 98% certain that all the places where we replaced `memcpy()` with `memmove()` do not intentionally involve overlapping buffers. Rather, the size argument is too large. If the two buffers are from `malloc()` and happen to be adjacent then it can appear they overlap due to the too large size argument.

PR #677, opened 2018-07-12
Hack around another read past the end of heap buf

krader1961:
ASAN told us the buffer pointed to `sp` in the `memmove()` in `nv_putval()`
was reading past the end of the malloced buffer pointed to by `sp`. This
doesn't actually fix the bug. It merely papers over the problem documented
in issue #676. This is in the hope that by eliminating one problem other
problems may go away or be easier to debug.

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -1394,7 +1394,7 @@ void nv_putval(Namval_t *np, const void *vp, int flags) {
     char *cp;
     union Value *up;
     int size = 0;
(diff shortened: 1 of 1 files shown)


krader1961:
FWIW, I would obviously prefer a proper fix rather than this hack. But there are so many problems identified by ASAN, and timeouts when running with ASAN enabled, that I'm desperate to simply eliminate any of those failures. Even if that means "papering over" (i.e., masking) the real bug.

siteshwar:
LGTM

Issue #475, opened 2018-04-17
ksh misparses parentheses in ${var-word}, ${var+word}, ${var:-word}, ${var:+word}


McDutchie:
$ ksh -c 'echo ${var-(word)}'
    ksh: syntax error at line 1: `(' unexpected
    $ ksh -c 'echo ${var-w(or)d}'
    w(or)d
    $ ksh -c 'echo ${var-(wor)d}'
    ksh: syntax error at line 1: `(' unexpected
    $ ksh -c 'echo ${var-w(ord)}'
    w(ord)
    $ ksh -c 'echo ${var-w(ord}'
    ksh: syntax error at line 1: `}' unexpected
    $ ksh -c 'echo ${var-wor)d}'
    ksh: syntax error at line 1: `)' unexpected

Same behaviour with `${var+word`}, `${var:-word}`, and `${var:+word}`.

These syntax errors should not happen; parentheses should not be parsed in this type of parameter substitution, as the `word` is a string literal and not a pattern (unlike `${var#word}`, etc.). [POSIX says](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) that the `word` is (only) subject to expansion.

This seems to be a regression introduced in early ksh93; "Version M 1993-12-28 s+" has the bug, but on ksh88 (on Solaris) it works as expected. It also works fine on every other current POSIX shell.

McDutchie:
Also see #431 which seems like it could be related.

siteshwar:
For the record, the error is coming from [here](https://github.com/att/ast/blob/448bf23d5621d58c45bdc69a331015627854c120/src/cmd/ksh93/sh/lex.c#L900-L904). But I will avoid touching parser code for now.

McDutchie:
This fix seems incomplete:
```
$ build/src/cmd/ksh93/ksh -c 'echo ${var-w(ord}'
w(ord
$ build/src/cmd/ksh93/ksh -c 'echo ${var:-w(ord}'
build/src/cmd/ksh93/ksh: syntax error at line 1: `}' unexpected
$ build/src/cmd/ksh93/ksh -c 'echo ${var+w(ord}'
build/src/cmd/ksh93/ksh: syntax error at line 1: `}' unexpected
$ build/src/cmd/ksh93/ksh -c 'echo ${var:+w(ord}'
build/src/cmd/ksh93/ksh: syntax error at line 1: `}' unexpected
```
and even stranger (note the third output):
```
[...]

PR #678, opened 2018-07-12
Fix Invalid Lexing Of Parameter Expansions With Parens

etscrivner:
This addresses #475 where ksh was found to incorrectly parse parameter
expansions with parentheses when the right-hand side of the expansion is a
string literal.

After a significant amount of analysis it was found that the lexer was mostly
functioning correctly, and that in this particular state it would actually do
the correct thing so long as it was placed in a mode where it would begin
treating the values following the parameter expansion characters as quoted.

### DIFF
--- a/src/cmd/ksh93/tests/modifiers.sh
+++ b/src/cmd/ksh93/tests/modifiers.sh
@@ -44,3 +44,41 @@ unset foo
 foo="non-null value"
 [[ ${foo:+bar} == "bar" ]]  || log_error  '${foo:+bar} not bar when foo is not null'
 [[ ${foo+bar} == "bar" ]]  || log_error  '${foo+bar} not bar when foo is not null'
+
+# Check for regressions on issue #475 where parens after `-', `+', and `=' were causing syntax errors
+unset foo
+[[ ${foo:-(bar)} == "(bar)" ]]  || log_error  '${foo:-(bar)} not (bar) when foo is not set'
+
+unset foo
+[[ ${foo-(bar)} == "(bar)" ]] || log_error '${foo-(bar)} not (bar) when foo is not set'
+
+unset foo
+[[ ${foo-(wor)d} == "(wor)d" ]] || log_error '${foo-(wor)d} not (wor)d when foo is not set'
+
+unset foo
+[[ ${foo-w(or)d} == "w(or)d" ]] || log_error '${foo-w(or)d} not w(or)d when foo is not set'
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -44,3 +44,41 @@ unset foo
...
+
+foo="non-null value"
+[[ ${foo:+(bar)} == "(bar)" ]]  || log_error  '${foo:+(bar)} not (bar) when foo is not null'
+[[ ${foo+(bar)} == "(bar)" ]]  || log_error  '${foo+(bar)} not (bar) when foo is not null'
+
+unset foo
+[[ ${foo=(word)} == "(word)" ]] || log_error '${foo=(word)} not (word) when foo is not set'
+
+unset foo
+[[ ${foo=w(or)d} == "w(or)d" ]] || log_error '${foo=w(or)d} not w(or)d when foo is not set'
```
siteshwar: You can convert one of the test cases for `$(parameter=value}` to `:=` style.

src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -44,3 +44,41 @@ unset foo
...
+unset foo
+[[ ${foo-(wor)d} == "(wor)d" ]] || log_error '${foo-(wor)d} not (wor)d when foo is not set'
+
+unset foo
+[[ ${foo-w(or)d} == "w(or)d" ]] || log_error '${foo-w(or)d} not w(or)d when foo is not set'
+
+unset foo
+[[ ${foo-w(ord} == "w(ord" ]] || log_error '${foo-w(ord} not w(ord when foo is not set'
+
+ unset foo
```
siteshwar: Nitpick: Empty space at beginning of line.

siteshwar:
Thanks for including test cases. Other than couple of suggestions LGTM.

src/cmd/ksh93/tests/modifiers.sh
etscrivner: Converted the last test to this format.

src/cmd/ksh93/tests/modifiers.sh
etscrivner: Good catch! Fixed.

krader1961:
Merged. Thx, @etscrivner. You're a brave man to tackle something like this.

PR #679, opened 2018-07-12
Enable ASAN testing with clang on macOS

krader1961:
This makes it possible to build with ASAN enabled using clang on macOS
and FreeBSD. However, the resulting ksh binary fails to run on FreeBSD
due to the `getenv()` intercept which interferes with ASAN
initialization. TBD is what to do about that.


siteshwar:
LGTM

Issue #403, opened 2018-02-11
The `leaks` ksh test is broken


krader1961:
While working on issue #396 I noticed that the `leaks` test fails on all my platforms other than the Linux based ones. The failure occurs whether AST Vmalloc or the system malloc is used. That's because the test is fundamentally broken. You cannot assume the virtual size of the process won't change when executing a block of ksh script. At a minimum you have to execute the block of code that might cause a leak a few times (in an attempt) to stabilize the memory allocator. Then establish the baseline size of the process, run the block of code some more times, then test whether the size has changed. Even that might result in false positives due to factors such as whether or not malloc debug features are enabled or tools like Valgrind or Asan are used.

krader1961:
This test is also outrageously expensive in terms of run time. So I'm going to disable it until these problems with the test are fixed.

siteshwar:
I think we should not need a separate `leaks` tests. If we manage to switch to standard malloc functions, it should be sufficient to run current tests through asan.

DavidMorano:
@siteshwar 
> If we manage to switch to standard malloc functions, it should be sufficient to run current tests through asan.

Hoping for the day AST VMALLOC can be swapped out for system MALLOC, and we can get everything run w/ ASAN! Everything should have been originally tested using the regular system MALLOC (and debugging versions of MALLOC subsystem) from day one, but at least now we know where we should be going in the future w/ memory allocation sanitation.

PR #681, opened 2018-07-13
Reenable leaks tests

siteshwar:
Also, move one test for command substitution to subshell tests.
Resolves: #403


krader1961:
LGTM

PR #686, opened 2018-07-14
Fix ASAN bug in lex.c

### LINKED ISSUE #685: ASAN reports a bug in lex.c when running the "expand" unittest
```
==20797==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000108e7e4bf at pc 0x000108afe918 bp 0x7fff5725dbc0 sp 0x7fff5725dbb8
READ of size 1 at 0x000108e7e4bf thread T0
    #0 0x108afe917 in sh_lex lex.c:1159
    #1 0x108bc53ae in simple parse.c:1430
    #2 0x108bbd0db in item parse.c:1278
    #3 0x108bb9e15 in term parse.c:560
...
0x000108e7e4bf is located 1 bytes to the left of global variable 'sh_lexstate2' defined in '../src/cmd/ksh93/data/lexstates.c:92:19' (0x108e7e4c0) of size 256
0x000108e7e4bf is located 63 bytes to the right of global variable 'sh_lexstate1' defined in '../src/cmd/ksh93/data/lexstates.c:61:19' (0x108e7e380) of size 256
```

The problem is this block of code in *src/cmd/ksh93/sh/lex.c*:

```
                if (sh_isoption(lp->sh, SH_BRACEEXPAND) && c == LBRACE && !assignment &&
                    state[n] != S_BREAK && !lp->lex.incase && !lp->lex.intest &&
                    !lp->lex.skipword) {
                    wordflags |= ARG_EXP;
                }
```

The error is that `n` is -1 and thus `state[n]` is indexing the entry prior to the array pointed to by `state`. FWIW, `state` is an alias for array `sh_lexstate2` at that juncture.

krader1961:
The `sh_lex()` function incorrectly dereferences `state[n]` when `n` is -1.
If `fcgetc()` returns -1 it means we've reached EOF. So any attempt to
use that unavailable character is meaningless. Just break from the loop
regardless of whether `lp->lex.reservok` is true.

Fixes #685

### DIFF
--- a/src/cmd/ksh93/sh/lex.c
+++ b/src/cmd/ksh93/sh/lex.c
@@ -1142,13 +1142,10 @@ int sh_lex(Lex_t *lp) {
                 }
                 isfirst = (lp->lexd.first && fcseek(0) == lp->lexd.first + 1);
                 n = fcgetc();
+                if (n < 0) break;
                 // Check for {}.
                 if (c == LBRACE && n == RBRACE) break;
-                if (n > 0) {
-                    fcseek(-LEN);
-                } else if (lp->lex.reservok) {
-                    break;
-                }
+                fcseek(-LEN);
                 // Check for reserved word { or }.
                 if (lp->lex.reservok && state[n] == S_BREAK && isfirst) break;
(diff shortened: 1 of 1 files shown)


etscrivner:
LGTM

PR #688, opened 2018-07-15
Eliminate getenv() intercept

### LINKED ISSUE #687: `getenv()` and `setenviron()` intercepts cause problems for ASAN
The modules *src/lib/libast/misc/astintercept.c*, *src/lib/libast/misc/setenviron.c*, and *src/lib/libast/misc/getenv.c* cause problems for ASAN testing on macOS, FreeBSD and other platforms. The mechanism implemented by those modules was, arguably, justified when libast and ksh were distinct projects. Today that mechanism is no longer needed. And since it interferes with tools like ASAN and makes it hard for a human to reason about the code it should be removed.

Note that `setenviron()` isn't even a standard interface found on any UNIX like OS I have used or can find via Google. And intercepting it while not also intercepting `putenv()` and `setenv()` seems rather pointless and silly. I'm guessing this intercept is to support legacy libast APIs which are no longer relevant.

krader1961:
The libast intercepts of `getenv()` and `setenviron()` cause problems
for ASAN testing. Since we no longer support building and using libast
independent of ksh remove the indirection. This makes it possible to
enable ASAN on platforms like FreeBSD. Yes, this tightly couples libast
with ksh. But that is no longer a negative since the libast in this
project is not used outside of the `ksh` command.

Note that I could have, at least in theory, resolved the interaction
problems with ASAN without removing the intercept redirection which allows
libast to be used independent of the `ksh` program. But doing so seems
pointless since we no longer support building libast for any reason other
than building the `ksh` command.

I also make no claim that this is the minimal or optimal change to achieve
compatibility with tools like ASAN. But this does improve the usefulness of
ASAN on macOS and FreeBSD without introducing problems on Linux platforms.

Fixes #687


krader1961:
Note that I recognize this change couples libast with ksh more tightly. That is not a problem since this project stopped providing a standalone libast library a year ago. The intent or this change is to eliminate unneeded complications due to the legacy loose couplings which allowed ksh to use a libast it was not built with. This is important since the legacy code causes problems for modern tools like ASAN which would otherwise help us identify problems with the code.

etscrivner:
LGTM, a nice bit of cleanup. 👍

PR #692, opened 2018-07-16
Compiler warnings

krader1961:
There is no longer any need to suppress a lot of warnings as the code
has been cleaned up to avoid triggering those warnings. Furthermore, the
one compiler flag that does make sense, `-fno-strict-aliasing`, is only
being enforced for Travis builds, not local builds. Move that flag into
the Meson config.

### DIFF
--- a/scripts/build-on-docker.sh
+++ b/scripts/build-on-docker.sh
@@ -15,7 +15,7 @@ mkdir build
 cd build
 
 echo ==== Configuring the build
-if ! meson
+if ! meson -Dwarnings-are-errors=true
 then
     cat meson-logs/meson-log.txt
     exit 1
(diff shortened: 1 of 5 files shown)


krader1961:
FYI, Note that the `CFLAGS` env var has no effect on the `ninja` compiler invocations. It only affects the compiler invocations during the `meson` build config step. That env var was never doing anything useulf. If you add `-Werror` to that env var it breaks the config step because meson does a lot of source compiles that are expected to have compiler warnings. Hence the need to do this via a Meson config option.

scripts/build-on-docker.sh
```diff
@@ -15,7 +15,7 @@ mkdir build
 cd build
 
 echo ==== Configuring the build
-if ! meson
+if ! meson -Dwarnings-are-errors=true
```
siteshwar: I am fine with merging it now. Let's see how it goes in practice.

siteshwar:
LGTM

etscrivner:
LGTM, makes sense to me to tighten down a bit on compiler warnings.

PR #693, opened 2018-07-17
Small refactoring of sh_lex

etscrivner:
The `sh_lex` function is currently a 1000+ line behemoth that severely needs to
be broken apart in order to reclaim any degree of intellectual understanding
and control.

This is a first, modest attempt at such a refactoring. Namely, the following
things are done to help clarify portions of the function:

* Add `S_NOP` lex state to indicate states that are not processed.
* Add `read_to_next_state_transition` function to encapsulate lexer behavior.
* Break apart opening assignments so they are one per line.
* Rename `state` to `state_table`.
* Add `data` variable to take on overloaded `state_table` usage as string data holder.

### DIFF
--- a/src/cmd/ksh93/include/lexstates.h
+++ b/src/cmd/ksh93/include/lexstates.h
@@ -23,6 +23,7 @@
 #include <wchar.h>
 #include <wctype.h>
 
+#define S_NOP 0    // absence of a state change, do nothing.
 #define S_BREAK 1  // end of token
 #define S_EOF 2    // end of buffer
 #define S_NL 3     // new-line when not a token
@@ -92,7 +93,7 @@
 #define LEN _Fcin.fclen
 #define isaname(c) ((c) > 0x7f ? isalpha(c) : sh_lexstates[ST_NAME][(c)] == 0)
 #define isaletter(c) ((c) > 0x7f ? isalpha(c) : sh_lexstates[ST_DOL][(c)] == S_ALP && (c) != '.')
-#define STATE(s, c) (s[mbwide() ? ((c = fcmbget(&LEN)), LEN > 1 ? 'a' : c) : (c = fcget())])
+#define STATE(s, c) ((s)[mbwide() ? (((c) = fcmbget(&LEN)), LEN > 1 ? 'a' : (c)) : ((c) = fcget())])
 #define isadigit(c) (sh_lexstates[ST_DOL][c] == S_DIG)
 #define isastchar(c) ((c) == '@' || (c) == '*')
 #define isexp(c) (sh_lexstates[ST_MACRO][c] == S_PAT || (c) == '$' || (c) == '`')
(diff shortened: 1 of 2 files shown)


etscrivner:
This PR is largely intended as a conversation starter around the following things:
- Do we think refactorings like this are a useful way to help clean up the system?
- How would we like to document new methods, constants, etc?

src/cmd/ksh93/sh/lex.c
```diff
@@ -258,7 +258,7 @@ Lex_t *sh_lexopen(Lex_t *lp, Shell_t *shp, int mode) {
 #ifdef DBUG
 extern int lextoken(Lex_t *);
 int sh_lex(Lex_t *lp) {
-    Shell_t *shp = lp->sh;
+    Shell_t *shell_interpreter = lp->sh;
```
krader1961: As much as I dislike a lot of the existing idioms (not to mention too short var names) this is a case where you should follow the existing idiom of naming this var `shp`. I want to be able to `grep shp` and find all the places this `Shell_t*` is used. I have no objections to establishing a different, clearer, idiom but that should be done in a separate change.

etscrivner: 👍 Got it, will make sure to isolate these kinds of changes in the future.

etscrivner:
@krader1961 Thank you for the thoughtful commentary. 

I've gone ahead and split this change into 3 separate commits to hopefully make it easier to evaluate each change individually. I've also removed the TODO comment in the newly created method and opted to avoid the check altogether as per your recommendation.

src/cmd/ksh93/include/lexstates.h
```diff
@@ -23,6 +23,7 @@
 #include <wchar.h>
 #include <wctype.h>
 
+#define S_NOP 0    // absence of a state change, do nothing.
```
krader1961: I like the introduction of this symbol rather than using the magic `0` below but it should be used in the state tables like `sh_lexstate0` as well rather than magic zero in those tables.

src/cmd/ksh93/sh/lex.c
```diff
@@ -340,11 +354,7 @@ int sh_lex(Lex_t *lp) {
     lp->typed = 0;
 
     while (1) {
-        // Skip over characters in the current state.
-        state = sh_lexstates[mode];
-        while ((n = STATE(state, c)) == 0) {
-            ;  // empty loop
-        }
+        read_to_next_state_transition(mode, &state, &n, &c);
```
krader1961: I don't think the `state = sh_lexstates[mode];` statement belongs in this new function as that assignment is orthogonal to the purpose of the while loop you moved into that function. And `n` should be passed in by value and returned by the function.

Note that there are only two places that use the `STATE` macro. It is both badly named and the wrong abstraction. Mostly because the author chose to implement it as a macro rather than a function. The assignment to `c` should be done separate from the state table lookup, for example. And it should probably be named something like `NEXT_TOKEN_STATE`. Which assumes we kept it as a macro which we probably shouldn't.

etscrivner:
@krader1961 Thanks for the review and no worries at all. I was mostly hoping to get a sense of how you all are thinking about changes with this PR and I think it definitely succeeded in helping me gain some understanding in that regard.

I'll work on taking some smaller steps here and close this.

PR #695, opened 2018-07-18
Better regression tests for parameter expansions

### LINKED ISSUE #475: ksh misparses parentheses in ${var-word}, ${var+word}, ${var:-word}, ${var:+word}

etscrivner:
This change replaces the previous set of regression tests for #475 with a more
complete suite of tests as recommended in the ticket.

Closes #475

### DIFF
--- a/src/cmd/ksh93/tests/modifiers.sh
+++ b/src/cmd/ksh93/tests/modifiers.sh
@@ -45,70 +45,52 @@ foo="non-null value"
 [[ ${foo:+bar} == "bar" ]]  || log_error  '${foo:+bar} not bar when foo is not null'
 [[ ${foo+bar} == "bar" ]]  || log_error  '${foo+bar} not bar when foo is not null'
 
-# Check for regressions on issue #475 where parens after `-', `+', and `=' were causing syntax errors
-unset foo
-[[ ${foo:-(bar)} == "(bar)" ]]  || log_error  '${foo:-(bar)} not (bar) when foo is not set'
-
-unset foo
-[[ ${foo-(bar)} == "(bar)" ]] || log_error '${foo-(bar)} not (bar) when foo is not set'
-
-unset foo
-[[ ${foo-(wor)d} == "(wor)d" ]] || log_error '${foo-(wor)d} not (wor)d when foo is not set'
-
-unset foo
-[[ ${foo-w(or)d} == "w(or)d" ]] || log_error '${foo-w(or)d} not w(or)d when foo is not set'
-
-unset foo
-[[ ${foo-w(ord} == "w(ord" ]] || log_error '${foo-w(ord} not w(ord when foo is not set'
-
-unset foo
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -45,70 +45,52 @@ foo="non-null value"
...
+# Check for regressions on issue #475 where parens after `-', `+', and `=' were causing syntax
+# errors. We check both the unset variable case and the set variable case for each set of symbols.
+unset -v foo
+for op in - :- = :=
+do
+    for word in '(word)' 'w(or)d' '(wor)d' 'w(ord)' 'w(ord' 'wor)d'
+    do
+        if [[ $(eval "echo \${foo${op}${word}}") != "${word}" ]]
+        then
+            log_error "\${foo${op}${word}} not ${word} when foo is not set"
```
krader1961: Don't bother changing this because the tests pass and we don't expect them to ever fail. But FWIW the reason I rewrote the original `error` function and renamed it `log_error` was that I wanted it to be easy to compare the values when the expected and actual values are different. So I would have written this along these lines:

```
for expect in '(word)' ...
do
    actual=$(eval "echo ...")
    [[ "$expect" == "$actual" ]] || log_error "mesg" "$expect" "$actual"
done
```

That will vertically align the expected and actual values which is often quite helpful in comparing the values (albeit perhaps not so much in this situation).

krader1961:
LGTM Merged. Doing this in what amounts to a table driven fashion is much better 👍

src/cmd/ksh93/tests/modifiers.sh
etscrivner: 👍 Will remember this in future patches.

PR #696, opened 2018-07-18
Convert getop() from a macro to a function

krader1961:
The getop macro is hard to read, understand, and edit. Convert it to a
function. I tested both implementations. The non-ASCII implementation
works fine on ASCII systems it is just less efficient than the ASCII
version.


etscrivner:
LGTM. Significantly easier to comprehend and maintain in this form.

PR #697, opened 2018-07-19
Integrate csbuild in travis builds

siteshwar:


krader1961:
I merged the PR and added a second commit to introduce an unused function. Csbuild took 9 minutes and produced so much output I didn't know where to begin. It did seem to detect the new warning. I then removed the second commit that introduced the unused function and added a commit to not enable `-Werror`. I force pushed the branch and the build failed with this error:

```
csbuild: error: failed to resolve the range of git revisions: 431f751510d7...9153041651e0
```

I then did `git checkout -b csbuild2` and pushed that to Github. It failed with this error:

```
csbuild: error: argument -g/--git-commit-range: expected one argument
```

That's because `TRAVIS_COMMIT_RANGE` is defined thusly:

> The range of commits that were included in the push or pull request. (Note that this is empty for builds triggered by the initial commit of a new branch.)

I'll grant you it is somewhat odd to push a fresh branch but it probably should not result in a build failure.

So I added a dummy change and pushed that to Github. Ten minutes later and with 8,362 lines in the Travis log we learn `/usr/bin/csbuild: status: no new defects found!`. Yay!

Okay, that's better. I then introduced two warnings in a single module. The build took 9 minutes and [8788 log file lines](https://travis-ci.org/krader1961/ast) but it did fail the build due to the two new warnings. Something we get much cheaper with `-Werror`. And if we want checks like `-Wmissing-field-initializers` we should just enable them even if just for local, non Travis, builds. It seems highly unlikely anyone is going to go through the Travis log to find warnings to be fixed.

In addition to the output being not particularly helpful there are lots of errors like this one:

```
/source/src/lib/libast/regex/regcomp.c: internal warning: child 6415 timed out after 5s <--[clang]
cswrap: error: child 6415 (/usr/bin/clang) terminated by signal 15 (timed out)
```

siteshwar:
> `csbuild: error: failed to resolve the range of git revisions: 431f751510d7...9153041651e0`

If you do a force push, `TRAVIS_COMMIT_RANGE` variable contains reference to previous commit which is not valid anymore. So that's why you see this error.

> I'll grant you it is somewhat odd to push a fresh branch but it probably should not result in a build failure.

We can skip `csbuild` if `TRAVIS_COMMIT_RANGE` is not set.

>  It seems highly unlikely anyone is going to go through the Travis log to find warnings to be fixed.

New warnings are shown at the end of the build. Due to a bug in `csbuild`, line `ADDED DEFECTS` appears after the warnings that were introduced:

```
Error: COMPILER_WARNING:
src/cmd/ksh93/sh/lex.c:2249:15: warning: unused parameter â€˜xâ€™ [-Wunused-parameter]
# void *WTF(int x) {
#           ~~~~^
# 2247|   }
# 2248|   
# 2249|-> void *WTF(int x) {
# 2250|       return 2;
# 2251|   }

Error: COMPILER_WARNING:
src/cmd/ksh93/sh/lex.c: scope_hint: In function â€˜WTFâ€™
src/cmd/ksh93/sh/lex.c:2250:12: warning: returning â€˜intâ€™ from a function with return type â€˜void *â€™ makes pointer from integer without a cast [-Wint-conversion]
#     return 2;
#            ^
# 2248|   
# 2249|   void *WTF(int x) {
# 2250|->     return 2;
# 2251|   }

ADDED DEFECTS
```

```
/source/src/lib/libast/regex/regcomp.c: internal warning: child 6415 timed out after 5s <--[clang]
cswrap: error: child 6415 (/usr/bin/clang) terminated by signal 15 (timed out)
```

Since `clang-analyzer` is slow and we have limited resources on the docker instance,  `csbuild` job was consistently timing out on Travis. Unfortunately this is a limitation of Travis. I set up time out to 5 seconds for each module.

I am currently checking if we can set up some infra within our company to run `csbuild`. The downside is that results will not be available publicly, something I wanted to avoid.

krader1961:
> If you do a force push...

Okay but that is something I do sometimes as a precursor to opening a PR to fix a stupid mistake. Kinda annoying to have the build fail just because csbuild didn't like the value of `TRAVIS_COMMIT_RANGE`.

> New warnings are shown at the end of the build.

Yes, I saw that. What's annoying is having to scroll through 8000+ lines just to see the new warnings. Too, the new warnings csbuild tells us about the compiler has already told us about, and will have halted the build due to `-Werror`, unless they are in one of the classes of warnings that csbuild enables which we do not normally enable. And my answer to that is we should enable those classes of warnings for normal, non-csbuild, builds if we care about those warnings.

Note that `bin/lint` will by default run IWYU, cppcheck, and oclint (with clang static analysis enabled) for the modified files in your work in progress or the most recent commit. Which makes it pretty fast most of the time. And if you want to focus on just a subtree you can do that too; e.g., `bin/lint src/cmd/ksh93/sh`.

The only thing csbuild does for us is keep new instances of warnings we currently ignore such as `-Wmissing-field-initializers` from being introduced. Rather than pay the huge cost of running csbuild with every commit I'd prefer to fix the existing instances of those warnings and then just enable them for all builds and let the compiler convert them to errors.

krader1961:
I decided to address the `-Wunused-parameter` warnings because it's a straightforward janitor cleanup involving nothing more than adding `UNUSED(param_name);` (see commit 15f8de6). The csbuild log is now 6000+ lines rather than 8000+ lines long. Any future commits which introduce unused parameters will now cause the build to fail without needing csbuild. That took me less than two hours.

krader1961:
I also downloaded the raw log file from my latest change and summarized the warnings from csbuild:

```
$ sed -ne 's/^.* \[\(-W[a-zA-Z-]*\)\].*$/\1/p' log.txt | sort | uniq -c
  90 -Wcast-function-type
  18 -Wempty-body
 283 -Wmissing-field-initializers
 291 -Wsign-compare
   2 -Wunused-but-set-parameter
```

Seems like it should be straightforward to fix/suppress those warnings. Then add `-Wall` (which is what `csbuild` is doing when invoked with `-w1`) and have the benefits of csbuild without the cost. If we really don't want code to be added which causes those warnings then we should just fix the existing instances and add `-Wall` to our config. We can do so incrementally as I did with commit 15f8de6.

kdudka:
@krader1961 Take it easy.  Nobody is forcing you to use csbuild.

I can easily hook it on my own fork of ksh and schedule a cron job to synchronize it with upstream hourly.  This has been working successfully for other upstream projects, too -- see the `csbuild` branches in my forks of  `curl` and `zsh` for example.

On top of that, @siteshwar is working on integrating upstream ksh with our internal static analysis service.  We can create pull requests and issues for the bugs it finds but cannot give you direct access to that service.

krader1961:
> Nobody is forcing you to use csbuild.

If this PR is merged I am most definitely being forced to use it because every time I push a commit to Github it will run it through the Travis csbuild environment. Which would be fine if it was adding more value than cost. I just don't see the the value proposition for this particular project. As I've said before I am sure there are situations where using this tool adds more value than cost. Your examples, @kdudka, of dealing with upstream projects like `curl` and `zsh` which need to be integrated into Red Hat distros are good examples for how csbuild can add value. But that is not the situation for anyone contributing changes to this project.

With respect to Red Hat's proprietary code analysis tools I fail to see the relevance to integrating `csbuild` into this projects CI environment. I've already made it drop dead simple to run the files touched by a commit, or the entire code base, through IWYU, cppcheck, and oclint. It's not obvious that predicating merging a change on cppcheck not reporting new problems is something we want to enforce.

krader1961:
@siteshwar I have added the `-Wextra` flag to the build. Which is what `csbuild` apparently uses despite my earlier misstatement that it uses `-Wall`. I have disabled two of the warnings `-Wextra` implicitly enables because they result in far too many false positives for this project. Especially with respect to GCC8's `-Wcast-function-type`. Although I did fix several of the easier ones to resolve. I have also fixed all the other existing code which caused warnings under `-Wextra`. Coupled with our use of `-Werror` this PR is no longer relevant so I'm closing it.

PR #698, opened 2018-07-20
Replace Magic 0 In Lexer Tables

etscrivner:
Replace the magic `0` value indicating that an entry does not produce a state
change with the new macro constant `S_NOP`.

In an attempt to keep this small no implicit or explicit checks that rely on
the value `0` are yet replaced outside of lexstates.h. This is reserved for future changes.

### DIFF
--- a/src/cmd/ksh93/data/lexstates.c
+++ b/src/cmd/ksh93/data/lexstates.c
@@ -25,11 +25,11 @@
 // This is the initial state for tokens.
 //
 static const char sh_lexstate0[256] = {
-    S_EOF,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   0,      S_NLTOK,
+    S_EOF,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_NOP,  S_NLTOK,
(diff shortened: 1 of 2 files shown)


krader1961:
LGTM. Merged with much thanks. This type of grunt work is not glamorous or fun. But it has a huge downstream impact.

A project I worked on at Google had, like nearly all Google projects, good code quality and extensive unit tests. But one day we found a bug not caught by the unit tests. That was because the unit tests all used zero as the starting value for critical values. And several tests were confusing one objects value for another; e.g., a XX object ID for a YY object ID. Because both object IDs started at zero and were incremented in lockstep the mistake wasn't detected. So I introduced randomization of the initial values used for each such object ID (possible because in this case they were stored in a MySQL DB). That not only found several bugs in the unit tests it found a bug in the code being tested. This is why I really dislike seeing every single @%#$(!$ enum and similar sequences start at zero. It makes it far too easy for mistaken identity to occur. And it's why I want to see symbols like `S_NOP` associated with some value other than zero. Preferably a semi-random value like -1, -2, 789, INT_MAX, etc. Semi-random in the sense that we don't use -1 or +1, for example, everywhere as a replacement for zero. I'd bet $100 USD that such a change will identify at least one bug. And probably more than one bug.

etscrivner:
@krader1961 Of course, happy to do this kind of thing. I'm hoping an accretion of small changes like this eventually adds up to a decent size impact.

Definitely agree with you about using a semi-random value in place of `S_NOP`. It's my goal to get to the place where this change can be made and these errors surfaced. I don't doubt there's >= 1 bug hiding in here somewhere related to this.

PR #699, opened 2018-07-20
Tidy Up Lexer Tables

etscrivner:
Cleanly divide all of the tables into equal length rows. Add index ranges in
comments on the right-hand side to make looking up particular entries easier.

### DIFF
--- a/src/cmd/ksh93/data/lexstates.c
+++ b/src/cmd/ksh93/data/lexstates.c
@@ -25,252 +25,313 @@
 // This is the initial state for tokens.
 //
 static const char sh_lexstate0[256] = {
-    S_EOF,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_NOP,  S_NLTOK,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,
+    S_EOF,  S_REG,  S_REG,   S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  //   0 -   7
+    S_REG,  S_NOP,  S_NLTOK, S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  //   8 -  15
+    S_REG,  S_REG,  S_REG,   S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  //  16 -  23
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/data/lexstates.c
```diff
@@ -25,252 +25,313 @@
...
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  S_REG,
-    S_REG,  S_REG,  S_REG,  S_REG,  S_REG,  S_REG,   S_REG,
+    S_EOF,  S_REG,  S_REG,   S_REG,   S_REG,  S_REG,   S_REG,   S_REG,  /*   0 -   7 */
```
siteshwar: We use `//` for comments.

etscrivner: Ah right, will fix.

krader1961:
FWIW, the legacy C style `/* ... */` comments should be used only when they have to be embedded the middle of a statement. Which should be extremely rare and most of those cases should probably be modified to hoist the comment out of the statement so the more modern `// ...` form can be used.

etscrivner:
@krader1961 That makes perfect sense and I've gone ahead and fixed the commenting style in this PR.  This is also well called out in the "C Style Guide" in the contributor guidelines https://github.com/att/ast/blob/master/CONTRIBUTING.md#c-style-guide

krader1961:
Merged. As before, @etscrivner, much thanks for doing unglamorous scut work. This type of cleanup both makes it easier to understand the code and forms crucial baby steps for more substantive refactoring.

PR #706, opened 2018-07-22
Remove __OBSOLETE__

### LINKED ISSUE #702: Symbol __OBSOLETE__ needs to be removed
While investigating issue #701 I noticed symbol `__OBSOLETE__`. This is a mechanism to automate the removal of code from the build artifacts. Philosophically I am opposed to this approach. The code that is compiled should not change just because I compiled the code today rather than yesterday assuming no commits between the two builds. Adding `TODO:` comments and open issues as reminders that blocks of code have a sunset date and which require human intervention to either retain or remove the code is safer.

The master and beta branches defines it thusly:

```
__OBSOLETE__ == $("6 months ago":@F=%(%Y0101)T)
```

The value in the resulting build files is `-D__OBSOLETE__=20120101` for the master branch and `-D__OBSOLETE__=20140101` for the beta branch. We are not defining the symbol which means it is implicitly zero. Because it is used in tests like `#if __OBSOLETE__ < 20140101` we are effectively building code that the AST team did not intend to be used.

krader1961:
Resolves #702


etscrivner:
👍 LGTM, compilation and tests verified locally. Another nice bit of cleanup.

PR #708, opened 2018-07-23
Remove another silly use of pointerof()

krader1961:
In this case the use appears to be solely to keep `makelist()` from
erroring out due to not getting both a left and right expression pointer.
Note that if the line number that was coerced to a pointer was actually
used it would result in a SIGSEGV on most platforms since it points into
page zero which is normally not accessible precisely for this reason.


siteshwar:
LGTM

Issue #634, opened 2018-06-29
Stop building `pty` and drop the source?


krader1961:
A conscious decision was made a year ago to limit the focus of this project to the `ksh` command. Libast would be maintained only to the extent required by ksh so that features like Vmalloc could be dropped. And we finally removed all but one of the commands in *src/cmd/builtin*. In part because some of the commands were of historical interest only (e.g., `asa`). But also because every distro includes implementations of the others, generally from BSD or GNU, that they are not going to replace with the versions in this project.

At present we only build two commands: `ksh` and `pty`. I propose dropping `pty`. It has at least one known bug (see issue #375). And it doesn't provide any functionality not provided by `expect`; a much more widely known and used tool.

siteshwar:
@bitstreamout It is shipped with suse. I would appreciate your comments on this issue.

siteshwar:
I will open a pull request to drop it.

PR #710, opened 2018-07-24
Remove 'pty' command and related tests

siteshwar:
Resolves: #634


krader1961:
LGTM. Verified it builds and passes on FreeBSD and OpenBSD in addition to the Travis environments.

PR #711, opened 2018-07-24
Fix integer truncation in put_enum

etscrivner:
Coverity identified an issue with integer truncation in `put_enum`.  The
function was truncating the return values of `strcasecmp` and `strcmp` from an
`int` to an `unsigned short` when assigning them to the local variable `n`.
Since either of these methods can return a value that is not in the set
`{0, 1, -1}` the later check if `n == 0` could spuriously evaluate to true. For
example, in the case where either function returned `-65536`.

The fix is simply to change `n` from an `unsigned short` to an `int` to avoid
the possibility of truncation. Since the only purpose of `n` is the store the
return values of these checks, this does not have any side effects.


krader1961:
Thx. Merged. Your analysis is correct and in fact some systems do return the result of subtracting the first two codepoints in each string which are not identical.

PR #712, opened 2018-07-24
Remove `kill -Q n` flag

### LINKED ISSUE #709: Remove `kill -Q n` flag
There are now just four uses of the `pointerof()` macro. One of them involves sending a queued signal via `kill -Q n`. Note that this feature does not exist in the last stable release, ksh93u+. From the new man page:

```
On  systems  that support sigqueue(2) the −Q or −q option can be
used to send a queued signal with message number n.  With  −Q  n
is  sent  in  the pointer portion of the signal value and it can
hold an unsigned long integer.  With −q n is sent in the integer
portion of the signal value, it can hold a signed integer.
```

That paragraph will make little sense to anyone who isn't an experienced user of the UNIX API. While support for queued signals can be justified I see zero justification for being able to send an unsigned rather than signed int as the auxiliary datum. Removing that distinction will simplify the signal sending code and remove a highly questionable use of `pointerof()`.

krader1961:
It makes zero sense for ksh to support including a value that looks like
a pointer when sending a queued signal via the `kill` command. This
is another case of the original team implementing a capability solely
because the underlying POSIX API supports it. Not because there was a
use case or it made sense in the context of a shell.

Resolves #709


etscrivner:
LGTM. Confirmed tests locally 👍

PR #715, opened 2018-07-24
Allow switching between different ksh implementations

siteshwar:
through alternatives in fedora.


krader1961:
LGTM in as much as I don't see any obvious mistakes but since I don't know anything about RH packaging I'll just trust this makes sense.

PR #718, opened 2018-07-25
Remove another use of `pointerof()`

krader1961:
In addition to removing another questionable use of `pointerof()` this also
improves the clarity of the code by replacing magic numbers with symbols
that make the intent clear; i.e., whether to disable or delete the builtin.

### DIFF
--- a/src/cmd/ksh93/bltins/typeset.c
+++ b/src/cmd/ksh93/bltins/typeset.c
@@ -867,8 +867,8 @@ typedef struct Libcomp_s {
     unsigned int attr;
 } Libcomp_t;
 
-static Libcomp_t *liblist;
-static int nlib;
+static Libcomp_t *liblist = NULL;
+static int nlib = 0;
 
 #if SHOPT_DYNAMIC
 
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/bltins/typeset.c
```diff
@@ -1070,18 +1070,18 @@ int b_builtin(int argc, char *argv[], Shbltin_t *context) {
         sfputr(stkp, name, 0);
         errmsg = 0;
         addr = 0;
-        if (dlete || liblist) {
-            for (n = (nlib ? nlib : dlete); --n >= 0;) {
+        if (delete || nlib) {
+            for (n = (nlib ? nlib : delete ? 1 : 0); --n >= 0;) {
```
siteshwar: Before this patch, could this line have been called with `dlete=2` ?

krader1961: Yes, and that's a bug. Because if nlib is zero or one you shouldn't execute the body of this loop with `n == 2`.

etscrivner:
LGTM

PR #720, opened 2018-07-25
Refactor nv_mapchar()

krader1961:
There is exactly one place that calls `nv_mapchar()` with a null `np`
argument. That is the `typeset -M` use case. All that does is test
whether `wctrans(name)` is true. So replace that one caller with a
direct call to `wctrans()` and simplify `nv_mapchar()`.


etscrivner:
LGTM

PR #722, opened 2018-07-26
Remove fmtls()

krader1961:
The fmtls() function was only used by the src/cmd/tw/find.c module which
we dropped since we no longer support an external `find` command.
Removing fmtls() also means we can remove _iblocks().


krader1961:
Thanks, @etscrivner. Removing code that isn't used is important. It opens up opportunities for removing more dead code. Removing dead code decreases cognitive load, build time, lint warnings that can be ignored, false positive matches when "grep"ing for some string in the code, etcetera.

etscrivner:
@krader1961 Makes perfect sense, I'm definitely a big fan of removing dead/unused code for all of those reasons. Smaller programs haver few bugs and like you said lower cognitive load.

PR #724, opened 2018-07-26
Remove unused parameter from path_pwd

etscrivner:
The `flags` parameter formerly passed to path_pwd was completely unused.
Therefore, it was a prime target for removal.

This change removes the `flags` parameter and updates all invocations of
`path_pwd` to no longer pass this value in. In addition, the documentation
above the function is updated to more accurately describe its current behavior.


krader1961:
Merged. Thx. FWIW, this may be a result of removing the emulation of the 3D Filesystem (3DFS) from ksh a few months ago. Regardless, good catch and thanks for the simplification.

PR #725, opened 2018-07-26
Remove dead code in print.c

etscrivner:
Coverity issue 253596 identified a section of dead code. The variable `neg` in
`extend` was assigned a constant `0` value and then unused and unassigned until
it was checked for a positive value in the removed `if` statement. Since the
constant value `0` would never evaluate to true, the section of code guarded by
the `if` would never execute.

### DIFF
--- a/src/cmd/ksh93/bltins/print.c
+++ b/src/cmd/ksh93/bltins/print.c
@@ -674,7 +674,6 @@ static_fn const char *mapformat(Sffmt_t *fe) {
 static_fn int extend(Sfio_t *sp, void *v, Sffmt_t *fe) {
     UNUSED(sp);
     char *lastchar = "";
-    int neg = 0;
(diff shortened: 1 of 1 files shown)


krader1961:
Merged. This type of issue makes me wonder. Did the var used to have some use and an earlier change that made it irrelevant simply failed to entirely remove it? Or is it supposed to serve a purpose and there is a bug present because the var serves no purpose? I checked the 93v- and 93u+ branches and this issue exists there as well.

etscrivner:
@krader1961 It seems it certainly had a purpose, but it is unclear what that was. In addition, it appears this entire function is actually unused so I'm putting together a PR to remove it in its entirety right now.

etscrivner:
Actually, I spoke too soon. It is passed around as a function pointer and used that way.

PR #726, opened 2018-07-26
Remove dead code in nvtree

etscrivner:
Coverity issue 253760 identified more dead code due to constant assignment to
a variable. The variable `dsize` traversed the following path to its current
state:

* In 93u `dsize` was assigned `fp->dsize` if `fp` was not `NULL`, otherwise it was `0`.
* In 93v this was strangely changed to always be zero in a constant `#if 1` branch, 93u version was the other side of this branch.
* In master the `#if 1` branch was just removed and so this became a constant assignment.

So, this has been a part of the codebase since 93v and does not appear to be
the source of any known issue.

### DIFF
--- a/src/cmd/ksh93/sh/nvtree.c
+++ b/src/cmd/ksh93/sh/nvtree.c
@@ -1134,7 +1134,7 @@ Namfun_t *nv_isvtree(Namval_t *np) {
 // Get discipline for compound initializations.
 //
 char *nv_getvtree(Namval_t *np, Namfun_t *fp) {
-    int flags = 0, dsize = 0;
+    int flags = 0;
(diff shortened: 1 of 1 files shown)


krader1961:
FWIW, There are a huge number of questionable changes after the 93u+ release. Many of which introduced blatant problems such as use-after-free bugs not present in the 93u+ release. Hindsight being 20/20 it would have been safer to base our work on the 93u+ code and abandon the 93v- changes. It appears that changes to the post 93u+ code had inadequate review within the AST team by anyone competent to review those changes.

krader1961:
Merged. Thanks for dealing with issues reported by tools like oclint and Coverity Scan.

PR #728, opened 2018-07-26
Remove sigunblock()

krader1961:
There is only one use of sigunblock() and that is in the coshell code we
no longer enable by default and intend to remove in the future. So move
the definition of that function into that module and remove the globally
visible implementation from libast.


siteshwar:
LGTM

PR #730, opened 2018-07-26
Remove recfmt(), reclen(), recstr(), fmtrec()

krader1961:
More functions that are not used and thus just adding cost without benefit.


siteshwar:
LGTM

PR #731, opened 2018-07-26
Remove unused optesc(), optjoin()

krader1961:


siteshwar:
For some reason this test is occasionally failing on macOS in CI :

```
/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/ksh.subshell.XXXXXXX.iq92nTTT/subshell.sh[494]: wait: 52620: Killed

<E> subshell[445]: 'x=$( builtin -d cat /bin/cat; cat 128ki.lin | /bin/cat ) && print ${#x}' failed -- expected '131072', got 'hung'
```

But it's failure does not seem to be related to this pull request. I will open an issue to follow up on it.

PR #732, opened 2018-07-26
Remove debug.c and debug.h

krader1961:
This debug code might be useful but we haven't enabled it once in the
year since work on this project resumed. Seems safe to say those of us
working on the code will continue to use adhoc printf()'s and the like.


siteshwar:
LGTM

PR #736, opened 2018-07-28
Fix copy-paste error in ed_pcomplete

etscrivner:
In the `ed_pcomplete` method, the value used to determine the suffix length
`slen` was actually taking the length of the prefix. This seemed obviously
incorrect and was flagged in coverity in #253617 as a possible copy-paste
error.


krader1961:
LGTM. Merged.

PR #737, opened 2018-07-28
Fix expansion of multibyte IFS characters

### LINKED ISSUE #13: ksh93: "$*" joins positional parameters on the first byte of $IFS instead of first character
```
$ ksh -c 'IFS=é; set : :; echo "$*"' | hd
00000000  3a c3 3a 0a                                       |:.:.|
00000004
$ echo é | hd
00000000  c3 a9 0a                                          |...|
00000003
$ locale charmap
UTF-8
```

Expected `:é:` (3a c3 a9 3a 0a)

POSIX says it must be the first character not byte http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_02

bash and zsh use the first character.

etscrivner:
Closes #13. Previously, the `varsub` method used for the macro expansion of
`$param`, `${param}`, and `${param op word}` would incorrectly expand the
internal field separator (IFS) if it was a multibyte character. This was due to
truncation based on the incorrect assumption that the IFS would never be larger
than a single byte.

This change fixes this issue by carefully tracking the number of bytes that
should be persisted in the IFS case and ensuring that all bytes are written
during expansion and substitution.

### DIFF
--- a/src/cmd/ksh93/sh/macro.c
+++ b/src/cmd/ksh93/sh/macro.c
@@ -1792,10 +1792,21 @@ retry2:
                 mp->atmode = mode == '@';
                 mp->pattern = oldpat;
             } else if (d) {
-                if (mp->sp) {
-                    sfputc(mp->sp, d);
+                Sfio_t *sfio_ptr = (mp->sp) ? mp->sp : stkp;
+
+                // We know from above that if we are not performing @-expansion
+                // then we assigned `d` the value of `mp->ifs`, here we check
+                // whether or not we have a valid string of IFS characters to
+                // write as it is possible for `d` to be set to `mp->ifs` and
+                // yet `mp->ifsp` to be NULL.
+                if (mode != '@' && mp->ifsp) {
+                    // Handle multi-byte characters being used for the internal
+                    // field separator (IFS).
+                    for (int i = 0; i < mbsize(mp->ifsp); i++) {
+                        sfputc(sfio_ptr, mp->ifsp[i]);
+                    }
                 } else {
-                    sfputc(stkp, d);
+                    sfputc(sfio_ptr, d);
                 }
             }
         }
--- a/src/cmd/ksh93/tests/ifs.sh
+++ b/src/cmd/ksh93/tests/ifs.sh
@@ -1,0 +1,30 @@
+# These are the tests for the internal field separator (IFS).
+
+IFS=e
+set : :
+[[ "$*" == ":e:" ]] || log_error "IFS failed" ":e:" "$*"
+
+IFS='|' read -r first second third <<< 'one|two|three'
+[[ "${first}" == "one" ]] || log_error "IFS failed" "one" "${first}"
+[[ "${second}" == "two" ]] || log_error "IFS failed" "two" "${second}"
+[[ "${third}" == "three" ]] || log_error "IFS failed" "three" "${third}"
+
+# Multi-byte character checks will only work if UTF-8 inputs are enabled
+if [ "${LC_ALL}" = "en_US.UTF-8" ]
+then
+    # 2 byte latin accented e character
+    IFS=é
+    set : :
+    [[ "$*" == ":é:" ]] || log_error "IFS failed with multibyte character" ":é:" "$*"
+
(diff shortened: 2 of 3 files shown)


src/cmd/ksh93/tests/ifs.sh
```diff
@@ -0,0 +1,30 @@
...
+set : :
+[[ "$*" == ":e:" ]] || log_error "IFS failed" ":e:" "$*"
+
+IFS='|' read -r first second third <<< 'one|two|three'
+[[ "${first}" == "one" ]] || log_error "IFS failed" "one" "${first}"
+[[ "${second}" == "two" ]] || log_error "IFS failed" "two" "${second}"
+[[ "${third}" == "three" ]] || log_error "IFS failed" "three" "${third}"
+
+# Multi-byte character checks will only work if UTF-8 inputs are enabled
+if [ "${LANG}" = "C.UTF-8" ]
```
krader1961: Please use `en_US.UTF-8` rather than `C.UTF-8`. I'm working on killing the AST locale subsystem which means `C.UTF-8` will no longer be supported unless the platform supports it.

src/cmd/ksh93/sh/macro.c
```diff
@@ -1792,10 +1798,16 @@ static_fn bool varsub(Mac_t *mp) {
                 mp->atmode = mode == '@';
                 mp->pattern = oldpat;
             } else if (d) {
-                if (mp->sp) {
-                    sfputc(mp->sp, d);
+                Sfio_t *sfio_ptr = (mp->sp) ? mp->sp : stkp;
+
+                // Handle multi-byte characters being used for the internal
+                // field separator (IFS).
+                if (ifs_bytes) {
```
krader1961: Why introduce `ifs_bytes` rather than just test `mode == '@'` here? The existing code is already schizophrenic which I don't expect you to fix. For example, this definition of a member of `Mac_t`:

```
unsigned char ifs;        // first char of IFS
```

First, it isn't a non-ASCII char and second, even within the context of ASCII why is this var an `unsigned char` where the underlying var is a simple `char` which may or may not be signed.

krader1961: Too, I think it is more important to make this test very clear that it is really testing whether a "@" expansion or not it being handled rather than doing so indirectly via the variable you introduced. Long term this hideous function needs to be refactored. And the entire code base should consistently use wide chars or use UTF-8 internally. But I expect to be dead before that happens.

src/cmd/ksh93/tests/ifs.sh
```diff
@@ -0,0 +1,30 @@
...
+then
+    # 2 byte latin accented e character
+    IFS=é
```
krader1961: This statement raises an interesting question. All (nearly all?) of the unit tests presume to be running in an ASCII locale unless they explicitly set the relevant locale env vars to a non-ASCII locale. What does it mean to run a ksh script with a UTF-8 encoded character in a non-UTF-8 locale? Yes, your test a few lines above should keep this statement from being executed in that case. But the script will have already been parsed.j

etscrivner: I actually ran this test locally and in such a case the script parses fine - although the 2-byte character naturally becomes two separate characters. It strikes me that the internal lexer and parser themselves are highly reliable on the ASCII/UTF-8 encoding to function - so I suspect deeply deviant encodings would have other problems here.

src/cmd/ksh93/sh/macro.c
etscrivner: I suspect I was being overly conservative here - I'm a little afraid to reuse the defined variables in the method since it's not always easy to tell what unexpected side-effects an additional usage might introduce.

This way is definitely simpler and nixes the extra variable, so I've gone ahead and migrated to using this approach.

krader1961:
Merged. That's much cleaner. Although I have to say after looking at how the `ifs` and `ifsp` structure members and the `d` var are used I think the entire mechanism needs an overhaul. But that pretty much goes for the entire *macro.c* module.

PR #740, opened 2018-07-30
Fix coverity issues for unused values

siteshwar:
Resolves: CID#305912, CID#253749, CID#253822, CID#253739, CID#253552

### DIFF
--- a/src/cmd/ksh93/edit/edit.c
+++ b/src/cmd/ksh93/edit/edit.c
@@ -1309,7 +1309,7 @@ int ed_histgen(Edit_t *ep, const char *pattern) {
     if (ac > 0) {
         l = ac;
         argv = av = (char **)stkalloc(ep->sh->stk, (ac + 1) * sizeof(char *));
-        for (mplast = 0; l >= 0 && (*av = (char *)mp); mplast = mp, mp = mp->next, av++) {
+        for (; l >= 0 && (*av = (char *)mp); mp = mp->next, av++) {
             l--;
         }
         *av = 0;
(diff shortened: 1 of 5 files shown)


src/cmd/ksh93/edit/edit.c
```diff
@@ -1309,7 +1309,7 @@ int ed_histgen(Edit_t *ep, const char *pattern) {
     if (ac > 0) {
         l = ac;
         argv = av = (char **)stkalloc(ep->sh->stk, (ac + 1) * sizeof(char *));
-        for (mplast = 0; l >= 0 && (*av = (char *)mp); mplast = mp, mp = mp->next, av++) {
+        for (; l >= 0 && (*av = (char *)mp); mp = mp->next, av++) {
```
krader1961: Isn't the `mplast = mp;` at line 1300 also pointless?

src/lib/libcmd/wclib.c
```diff
@@ -390,7 +390,6 @@ int wc_count(Wc_t *wp, Sfio_t *fd, const char *file) {
                         }
                         while (mbc(c) && ((c & WC_ERR) || (c & 7) == 0)) c = type[*cp++];
                         if (eol(c) && (cp > endbuff)) {
-                            c = WC_MB | WC_ERR;
```
krader1961: Note that I recently fixed a related bug on line 391 which had an always true condition: `((c | WC_ERR) || (c & 7) == 0)`. I now have very low confidence this code does the right thing.

krader1961:
LGTM

src/cmd/ksh93/edit/edit.c
siteshwar: Probably not. It's value can be used in the next iteration of enclosing `for` loop.

PR #741, opened 2018-07-30
Remove a condition which is always true

siteshwar:
'a' is always 97 in decimal, so remove the test to check for it's value.

Resolves: cid#253601

### DIFF
--- a/src/cmd/ksh93/sh/string.c
+++ b/src/cmd/ksh93/sh/string.c
@@ -335,7 +335,8 @@ char *sh_fmtstr(const char *string, int quote) {
         while (op = cp, c = mbchar(cp)) {
             state = 1;
             switch (c) {
-                case ('a'==97?'\033':39): {
+                // Escape character
+                case ('\033'): {
                         c = 'E';
                         break;
                     }


src/cmd/ksh93/sh/string.c
```diff
@@ -335,7 +335,8 @@ char *sh_fmtstr(const char *string, int quote) {
         while (op = cp, c = mbchar(cp)) {
             state = 1;
             switch (c) {
-                case ('a'==97?'\033':39): {
```
krader1961: This is testing whether we're on a machine that uses ASCII or EBCDIC. If the latter `a` has value 129 and escape is 39. Note, however, that Linux on IBM zSeries machines uses ASCII so this compatibility code isn't needed in that environment. It's only needed if using ksh on something like z/OS UNIX. Personally I'd like to remove the EBCDIC compatibility code but it's probably something we should discuss in an issue dedicated to the question.

etscrivner: Concur on removing EBCDIC support, would be interested to hear what others think.

etscrivner:
LGTM

krader1961:
@siteshwar If you want to proceed with this before we've reached a decision on issue #742 please convert that runtime test into a build time test:

```
#if 'a' == 97
    case '\033':
#else
    case 39:
#endif
    {
```

DavidMorano:
@krader1961 
> If you want to proceed with this before we've reached a decision on issue #742 please convert that runtime test into a build time test:

I know that the more in depth clean-up does not come until later, but ...
I almost feel badly (almost) about butting in on such a small matter, but can this build-time code please be changed into something like:

/* near top of file where we are deciding how to compile */
#if 'a' == 97
#define  CH_WHICH '\033'
#else
 #define CH_WHICH  39
#endif

/* down in the depths of the code where ever this is normally */
switch (whatever) {
case CH_WHICH:
   /* stuff */
  break ;
/* other cases */
} /* end switch */

The idea is to get as many stupid conditional compiles *out* of the more normal code path as much as possible!

Thanks for any consideration.

DavidMorano:
@krader1961 
> What you're proposing is more or less what I proposed in issue #742 ...

Yes, you are correct. There you stated:
> Probably coupled with formal abstractions and preprocessor symbols.

I originally was going to try to stay away from this whole EBCDIC question -- trying in vain to be a "nice" guy -- but your insistence on wanting to get rid of EBCDIC entirely has inspired me to speak out just a little bit on this.

Yes, I agree with you completely!  I can not agree with you more. Has not that EBCDIC stuff been around long enough now? I mean really! How long should the world be required to ~~bow down to~~ put up with that ~~crap~~ stuff, especially given that the world has abandoned tons of other stuff much younger than EBCDIC?

IBM deserves respect for many of the things they have done in computing. No question there. But they (IBM) should have transitioned the world away from EBCDIC decades ago now! IBM should have taken the lead in getting rid of EBCDIC from the world. They did not entirely succeed. Now it is our turn. Yes, get rid of support for EBCDIC!

krader1961:
LGTM. It is pretty clear the AST/ksh in its current form is not going to work correctly on an EBCDIC system. And no one outside of IBM has any interest in making it work on such a system.

PR #743, opened 2018-07-30
Disable building api tests in coverity scan

siteshwar:
It's not on priority to fix defects found in these tests, so skip
building them.


krader1961:
LGTM

PR #746, opened 2018-07-30
Remove SHOPT_P_SUID

krader1961:
It isn't enabled by default, is essentially undocumented, and does not
appear to be needed. With this change we have just two legacy config time
symbols: SHOPT_COSHELL and SHOPT_BASH.


etscrivner:
LGTM

PR #750, opened 2018-07-31
Fix a case of double sh_close()

siteshwar:
File descriptor 'f' is closed by funload() function. There is no need to
call sh_close() on it separately.

Resolves: cid#287609


krader1961:
LGTM

PR #751, opened 2018-07-31
Check if sh_fcntl() failed in sh_redirect() function

siteshwar:
Resolves: cid#279530


krader1961:
LGTM but note there is at least one other place in this module that fails to deal with `sh_fcntl()` returning an error:

```
fdnew = sh_iomovefd(shp, sh_fcntl(fdold, F_DUPFD_CLOEXEC, 10));
```

siteshwar:
`sh_iomovefd` function checks it's second parameter by calling `sh_iovalidfd` and aborts if it's invalid.

PR #752, opened 2018-07-31
Remove a useless call

siteshwar:
This call has no effect.

Resolves: cid#253871


src/cmd/ksh93/sh/nvtype.c
```diff
@@ -668,7 +668,6 @@ static_fn int typeinfo(Opt_t *op, Sfio_t *out, const char *str, Optdisc_t *od) {
         if (strchr(nq->nvname, '.')) continue;
         tp = nv_type(nq);
         if (tp) {
-            nv_hasdisc(nq, &type_disc);
```
krader1961: Note that this was originally `Namfun_t *pp = nv_hasdisc(nq,&type_disc);` but the var was removed because it was unused. The function call was presumably left over concerns it might have a side-effect.

krader1961:
LGTM

PR #754, opened 2018-08-01
Disclaim support for EBDCIC

### LINKED ISSUE #742: Should we still support EBCDIC?
Coverity pointed out a condition that is always true on ASCII using systems. Which lead @siteshwar to open PR #741. That transformation, however, is only valid on ASCII using systems. There are other places in the code which attempt to support both ASCII and EBCDIC. See, for example, the definition of `getop()` in *src/cmd/ksh93/sh/streval.c*. But given all the places in the code which have hardcoded chars like `\033` (escape) my confidence is low that the current code will function correctly in an EBCDIC environment.

If we continue to support EBCDIC it should be done via build time tests such as `#if ('a' == 97)`. Probably coupled with formal abstractions and preprocessor symbols. Personally I would just as soon drop whatever support there is for EBCDIC and formally state that ksh only supports ASCII and encodings like ISO-8859 and UNICODE which are compatible with ASCII. Which basically means everything but EBCDIC.

krader1961:
This removes all the attempts to accomodate EBCDIC I could find other
than those in src/lib/libast/include/ast_ccode.h. That's because I expect
those to be removed in the future when we eliminate the AST locale code
in favor of the platform implementation.

Resolves #742


siteshwar:
LGTM

etscrivner:
LGTM

PR #756, opened 2018-08-01
Remove couple of functions related to EBCDIC

siteshwar:
Related: #742


krader1961:
LGTM. What that code implies is that at one time they were making a real attempt to support EBCDIC. But given the presence of statements that would only work on an ASCII system they weren't consistent about enforcing code set neutrality.

Issue #757, opened 2018-08-01
erroneous behaviour (confusing directories with executables/binaries)


jghub:
seen today after doing a de-novo git clone (from an OSX system): installation succeeded without issues. ksh ( Version A 93v-1373-gffd2462b) is running but there is a strange problem.

minimal example:
PATH=". :$PATH"   #i.e. put your working dir in front of search path
mkdir man    #generate a directory named like some command (in this example: man)
man ksh   #fails with 'ksh: man: cannot execute [Is a directory]'

so, empirically, ksh seemingly looks for the first file/directory with the specified name ('man' in this example) on the search path with the 'x' flag set and tries to execute it. it looks like it does not check whether it is dealing with a directory or an ordinary file. it should of course skip over/ignore directories here and locate and use the 'man' command.

siteshwar:
It works as expected in last stable release. I will look into it.

siteshwar:
It's a regression introduced by commit 17f2fe2caa809e7aa7f27746048fc8f231d7111a

siteshwar:
@jghub Thanks for the report. #758 should fix it.

PR #758, opened 2018-08-01
Do not show error if ksh fails to execute a directory

siteshwar:
Ksh shows an error if it fails to execute a command that has same name
as a directory that appears under PATH. This is a regression introduced
by commit 17f2fe2caa809e7aa7f27746048fc8f231d7111a and worked as
expected in ksh93u+ release.

Resolves: #757


krader1961:
LGTM but it really needs a unit test to ensure we don't regress again.

krader1961:
LGTM

PR #759, opened 2018-08-02
Tiny step to replacing AST locale subsystem

krader1961:
This is a very tiny step to replacing all the complexity of the AST
locale subsystem with the platform implementation. This effectively makes
`LC_OPTIONS=native` the default behavior.

This removes the wchar.sh unit test because it explicitly depends on the
`LC_OPTIONS` var which, while still technically valid with this change,
is not compatible with the goal of removing the AST locale subsystem.
Furthermore, the undocumented `printf "%(nounicodeliterals)q\n"` capability
is not present in the ksh93u+ release and I can see zero reason for
supporting it. The removal of that feature from the implementation will
be done in a subsequent commit.


krader1961:
Most of the unit test changes are due to one of three things:

a) They relied on a non-standard AST feature (e.g., `LC_ALL=debug`).

b) They relied on something like codepoint U+5929 being classified as an alphanum which should be true but some platforms got wrong. So I switched the codepoint to a non-ASCII value that was an alphanum.

c) They relied on locale C.UTF-8 which is only available on my Ubuntu machine and even there had some quirks. So switch to en_US.UTF-8.

Subsequent changes will incrementally remove code that is no longer needed. Then I'll shift back to removing the remaining AST'isms such as indirections like `ast.mb_towc`.

etscrivner:
LGTM, nice step in this direction.

krader1961:
I've merged the changes in this PR. More such changes soon to come.

PR #760, opened 2018-08-03
System locale

krader1961:


krader1961:
As I mentioned elsewhere this cleanup is going to require a large number of incremental changes unlike the removal of the AST Vmalloc implementation.

krader1961:
Merged because I want to keep my momentum up on ripping out the AST locale code.

PR #761, opened 2018-08-03
Fix error handling in hist_exceptf function

siteshwar:
Resolves: cid#294201

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -959,6 +959,7 @@ static int hist_exceptf(Sfio_t *fp, int type, Sfdisc_t *handle)
 {
     UNUSED(data);
     History_t *hp = (History_t *)handle;
+
     if (type == SF_WRITE) {
         if (errno == ENOSPC || hp->histwfail++ >= 10) return 0;
         // Write failure could be NFS problem, try to re-open.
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/edit/history.c
```diff
@@ -959,24 +959,27 @@ static int hist_exceptf(Sfio_t *fp, int type, Sfdisc_t *handle)
...
             close(newfd);
-            return -1;
+            goto fail;
         }
 
-        fcntl(oldfd, F_SETFD, FD_CLOEXEC);
+        do {
+            ret = fcntl(oldfd, F_SETFD, FD_CLOEXEC);
+            if (ret < 0 && errno != EINTR) goto fail;
+        } while (ret < 0 && errno == EINTR);
```
krader1961: It's not possible for this specific `fcntl()` call to fail with EINTR or to fail at all. Which is why the original author didn't check the return value. It should be sufficient to just do `(void)fcntl(...);`.

krader1961:
LGTM although the `sh_fcntl(newfd, F_DUPFD_CLOEXEC, oldfd) != oldfd` is also technically a "can't happen" situation that means something is seriously wrong (e.g., corrupted memory) that probably warrants an `abort()`. However, leaving it as a soft error is fine since it should never occur.

PR #762, opened 2018-08-04
Refactor error handling in sh_histinit function

siteshwar:
Resolves: cid#294200

### DIFF
--- a/src/cmd/ksh93/edit/history.c
+++ b/src/cmd/ksh93/edit/history.c
@@ -218,13 +218,14 @@ retry:
         if (shgd->userid)
 #endif  // KSHELL
         {
-            if (!(fname = pathtmp(NULL, 0, 0, NULL))) return 0;
+            if (!(fname = pathtmp(NULL, 0, 0, NULL))) goto fail;
             fd = open(fname, O_BINARY | O_APPEND | O_CREAT | O_RDWR, S_IRUSR | S_IWUSR | O_CLOEXEC);
         }
     }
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/edit/history.c
```diff
@@ -218,13 +218,14 @@ int sh_histinit(void *sh_context) {
         if (shgd->userid)
 #endif  // KSHELL
         {
-            if (!(fname = pathtmp(NULL, 0, 0, NULL))) return 0;
+            if (!(fname = pathtmp(NULL, 0, 0, NULL))) goto fail;
```
krader1961: There is no good reason to introduce a bunch or `goto`'s. That label does nothing but return. We want to remove goto's not add them; at least not without good reason. Other than that adding `(void)` to silence the unused `fcntl()` retval error is fine. Feel free to merge that single line change without the goto's.

siteshwar: Isn't it better to do error handling at just one place ?

siteshwar: Anyways, I will open a separate pull request to fix all these warnings.

krader1961: > Isn't it better to do error handling at just one place ?

If there was some actual error handling, sure. But in this case there is none. Long ago I worked with some programmers who believed a function should have at most to `return` statements: one for success and one for failure. That resulted in some awful code. There is nothing wrong with multiple return points within a function.

PR #763, opened 2018-08-04
Fix some coverity warnings for unchecked return values

siteshwar:
These calls should never fail, so we can ignore their return value.


krader1961:
LGTM. These are all innocuous, safe, changes to silence bogus lint warnings and I didn't see any obvious typos.

PR #767, opened 2018-08-06
Fix dir in $PATH on OpenBSD

### LINKED ISSUE #766: The recently introduced executable dir in $PATH test fails on OpenBSD 6.2
The test introduced by commit 497d3ed3 fails on OpenBSD 6.2.  This is due to a different failure path being taken in `path_spawn()` than on other systems. It's a text book example for why adding a unit test for newly reported bugs is so essential. Not just to help detect regressions but also to detect the same bug on other platforms.

krader1961:
If a directory in `$PATH` matches the name of the command and has its
execute bit (i.e., search bit) set on OpenBSD it can still be incorrectly
treated as a match for the command. It should be ignored since it is
a directory.

Fixes #766
Related #757

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -1094,15 +1094,20 @@ pid_t path_spawn(Shell_t *shp, const char *opath, char **argv, char **envp, Path
     }
     if (pid > 0) return pid;
 retry:
-    switch (shp->path_err = errno) {
+    shp->path_err = errno;
+    switch (shp->path_err) {
+        case EISDIR: {
+            return -1;
+        }
         case ENOEXEC:
         case EPERM: {
             // Some systems return EPERM if setuid bit is on.
             errno = ENOEXEC;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/path.c
```diff
@@ -1094,15 +1094,20 @@ pid_t path_spawn(Shell_t *shp, const char *opath, char **argv, char **envp, Path
...
+        }
         case ENOEXEC:
         case EPERM: {
             // Some systems return EPERM if setuid bit is on.
             errno = ENOEXEC;
             if (spawn) {
-                if (shp->subshell) return (-1);
+                if (shp->subshell) return -1;
                 do {
-                    if ((pid = fork()) > 0) return (pid);
```
krader1961: It is a shame that these anti-patterns are so common in this project. I don't think the authors were deliberately trying to make the code hard to read but that was the net effect.

siteshwar:
LGTM

Issue #579, opened 2018-06-05
Replace the AST locale subsystem with the system version


krader1961:
We recently replaced the AST Vmalloc subsystem with the system malloc subsystem. In the process numerous bugs were found that were being masked by Vmalloc. It also allowed us to shed over 10000 LOC that no longer needs to be maintained. I'm proposing we replace the AST locale subsystem with the system provided one. Because that may be difficult we might want to start by removing or disabling AST features that are not found in GNU or BSD locale subsystems. For example, env vars like `LC_OPTIONS=nounicodeliterals`. This would at least keep us from being locked into the AST implementation forever.

krader1961:
Something I didn't realize until just now is that the AST locale support can selectively use the native implementation. For example, if the native *wchar.h* header defines `iswprint` as a macro it is used rather than the AST implementation. See https://github.com/att/ast/issues/577#issuecomment-394929360. That caused me to Picard face-palm.

krader1961:
This is going to take many more commits than the removal of AST Vmalloc needed but it is definitely doable. The first round of changes has removed ~2200 LOC. At this point `ksh` is using the platform's locale subsystem. The bits of the AST locale subsystem still present are solely because I want to do this in as many bite size changes as possible to minimize the chance of a stupid mistake. And if such a mistake is made it should be easier to bisect. Too, there are a couple of bits of logic which still need to be implemented on top of the platform's locale subsystem.

krader1961:
It is really annoying how many places directly access what should be a private locale subsystem var such as `locales[]`. Either directly or indirectly via the `LCINFO()` macro. Annoying because it both obscures the purpose of the access and makes it harder to replace the AST locale subsystem with the platform implementation. Many, if not all, of these accesses appear to be done in the name of efficiency. Which is rarely a legitimate reason and definitely not so in any of the cases I've examined.

Despite my metaphorical teeth grinding as evidenced by the previous paragraph I think I'm close to being able to eliminate almost all of the AST locale implementation. Which mostly just wraps the native implementation other than the `C` and `C.UTF-8` locales which are special-cased and implemented directly by AST. Most of the references to the private implementation details can be replaced by calls to `setlocale(..., NULL)` or `nl_langinfo()`. I'm down to just two modules which have somewhat more complex interactions by virtue of modifying the AST locale subsystem state.

siteshwar:
Although the tests are passing in CI. I am getting these failures on my fedora 27 machine:

```
[...]

PR #768, opened 2018-08-06
Remove feature symbol `_locale_utf8_str`

krader1961:
It turns out we can't rely on the value assigned to `_locale_utf8_str`
during the build config phase to match the value returned by
`nl_langinfo(CODESET)` at run time. So drop that feature test and
broaden the set of values that mean the current locale uses UTF-8.

Related #579


krader1961:
@siteshwar Please confirm this fixes the failure on your Fedora 27 system.

PR #769, opened 2018-08-08
Fix write after free error in history.c

etscrivner:
This fixes Coverity #253710 in which we attempt to access the variable `hp`
after it has already been closed. This happens in the case where the predicate
`sfsync(hp->hisfp) < 0` evaluates to true. This results in a call to
`hist_close` which also frees `hp`.

Instead, we now avoid accessing `hp` altogether in the case where the predicate
is true, avoiding this issue.


krader1961:
LGTM. Merged. Thanks for taking care of these types of issues as they can theoretically lead to data corruption if not an outright SIGSEGV. And fixing these types of issues eliminates ASAN failures. Which today is the project's highest priority as far as I am concerned.

PR #771, opened 2018-08-08
More removal of EBCDIC system support

krader1961:
This only eliminates the uses of CC_NATIVE which clearly are due to
determining if we're not building on an EBCDIC system. Removing that,
and related, symbols and header files will occur in a future change
because I need to explore if the AST iconv implementation can be
eliminated.

Related #742


src/cmd/ksh93/sh/tdump.c
```diff
@@ -58,9 +58,7 @@ int sh_tdump(Sfio_t *out, const Shnode_t *t) {
 static_fn int outstring(Sfio_t *out, const char *string, int n) {
     int r;
     char *cp = (char *)string;
-    ccmaps(cp, n, CC_NATIVE, CC_ASCII);
```
siteshwar: What does this line do ?

PR #772, opened 2018-08-08
Remove AST iconv subsystem

### LINKED ISSUE #770: Remove AST iconv subsystem
As with issue #579 I see no good reason to retain the AST iconv subsystem when we're now targeting only (mostly) POSIX compliant systems which should all provide an iconv API. Its presence is making removal of the EBCDIC vs ASCII symbols in *ast_ccode.h* more difficult. And, of course, using a private implementation increases the likelihood that ksh will behave differently than other programs on the user's system.

krader1961:
The only platforms we now target are expected to provide a working iconv
subsystem. So remove the AST implementation. This is related to #579
which removed the AST local subsystem.

Resolves #770


siteshwar:
LGTM

Issue #774, opened 2018-08-08
`read' does not recognize newline as end-of-input (or another delimiter set via `-d delim')


jghub:
seen with Version A 93v-1410-gededd114 under OSX:

(hopefully) minimal example:
```
print -n "gimme input > "
read input
print got "'$input'"
```

instead of expected result (printing out the (newline terminated) input in quotes), the `read` reads beyond the newline indefinitely: it seemingly does not recognise newline as end-of-input as it should. same is happening when using `read -d delim`: the first character of the specified `delim` text is not terminating but just read as part of the input. in this case it is easier to directly see it on the screen that it is happening:
```
print -n "gimme input > "
read -dz input
print got "'$input'"
```
(just enter some text and than a final letter "z": it is simply echoed as part of the provided input instead of ignoring it and returning the rest in variable `input`).

overall it *seems* as if the delimiter to `read` is not honoured at all. that definitely must be some sort of regression (I noted it because one of my interactive scripts stopped working after switching to the above ksh version).

thanks for taking care of ksh!

joerg

kusalananda:
Testing the same version of the sources on OpenBSD 6.3, the `read -dz input` work as expected (`z` does not become part of the value stored in `input`).  However, `read input` locks the shell (it starts using 100% CPU) and the shell session has to be killed with the `KILL` signal.

jghub:
On 08.08.18 12:53 , Andreas Kusalananda Kähäri wrote:
> Testing the same version of the sources on OpenBSD 6.3, the `read -dz input` work as expected (`z` does not become part of the value stored in `input`).  However, `read input` locks the shell (it starts using 100% CPU) and the shell session has to be killed with the `KILL` signal.
> 

I may add that on OSX I see (sporadic, not quite reproducible) instances, where the never-returning 
`read input' cannot be aborted via SIGINT (ctrl-C) so that indeed it has to be killed, but most of 
the time SIGINT suffices.

siteshwar:
This issue is an example why we need to be extra careful while trying to fix warnings produced by static analysis tools. The commit in 1dc56d2151541c8f723240f20faf368a9fee02d9 fixed lint warnings, but introduced a side effect which is very hard to spot during reviews.

@jghub If you are looking for a quick fix, please pull my changes from #775.

jghub:
On 08.08.18 16:00 , Siteshwar Vashisht wrote:
[...]

PR #775, opened 2018-08-08
read builtin should not wait indefinitely

siteshwar:
read builtin waits indefinitely on executing commands like "read foo".
This is a regression introduced by
1dc56d2151541c8f723240f20faf368a9fee02d9 and before this commit
condition in for loop had a side effect even when it's false. This
commit reverts change to this line.

Resolves: #774

### DIFF
--- a/src/cmd/ksh93/edit/edit.c
+++ b/src/cmd/ksh93/edit/edit.c
@@ -488,8 +488,7 @@ void ed_setup(Edit_t *ep, int fd, int reedit) {
                     int skip = 0;
                     ep->e_crlf = 0;
                     if (pp < ppmax) *pp++ = c;
-                    for (n = 1; *last; n++) {
-                        c = *last++;
+                    for (n = 1; c = *last++; n++) {
                         if (pp < ppmax) *pp++ = c;
                         if (c == '\a' || c == ESC || c == '\r') break;
                         if (skip || (c >= '0' && c <= '9')) {


siteshwar:
Executing `read` is a very basic use case and this regression should have been caught by our test cases. 

@krader1961 I am thinking about starting to write tests with `expect`. Do you have any issues against it or suggestions for different tool ?

src/cmd/ksh93/edit/edit.c
```diff
@@ -488,8 +488,7 @@ void ed_setup(Edit_t *ep, int fd, int reedit) {
                     int skip = 0;
                     ep->e_crlf = 0;
                     if (pp < ppmax) *pp++ = c;
-                    for (n = 1; *last; n++) {
-                        c = *last++;
+                    for (n = 1; c = *last++; n++) {
```
krader1961: The problem is the structure of this loop which necessitates the `if (c == 0 || c == ESC || c == '\r') last--;` that follows it. Notice that the only reason it has to decrement `last` when the end of the string is reached is because of how the code is structured. And the other two conditions in that `if` should be handled in the block containing the `break`.  Better would be to defer the increment of `last` till the end of the loop just before the next iteration.

krader1961:
Yes, we should definitely write tests using `expect` to exercise behavior that requires interacting with the shell via a tty. The fish shell has quite a few such tests. It can take some work to make them robust but it's worth it.

krader1961:
See my counter proposal for fixing this in #774.

src/cmd/ksh93/edit/edit.c
krader1961: Also, if we decide to merge this simple fix rather than my more complex fix in PR #774, we should at least enclose the assignment in parens to eliminate the lint warning.

krader1961:
Closing in favor of my more complicated but clearer and more idiomatic solution from PR #776 that I've merged.

PR #776, opened 2018-08-09
Fix regression in the `read` builtin

### LINKED ISSUE #774: `read' does not recognize newline as end-of-input (or another delimiter set via `-d delim')

krader1961:
Please note the weird conditional increment of `last` at line 505. This
is equivalent to the logic prior to the change by commit 1dc56d2. Your
guess is as good as mine as to what the purpose of this is.

Resolves #774

### DIFF
--- a/src/cmd/ksh93/edit/edit.c
+++ b/src/cmd/ksh93/edit/edit.c
@@ -433,7 +433,6 @@ void ed_setup(Edit_t *ep, int fd, int reedit) {
     char *last, *prev;
     char *ppmax;
     int myquote = 0;
-    size_t n;
     int qlen = 1, qwid;
     char inquote = 0;
 
@@ -488,8 +487,10 @@ void ed_setup(Edit_t *ep, int fd, int reedit) {
                     int skip = 0;
                     ep->e_crlf = 0;
                     if (pp < ppmax) *pp++ = c;
-                    for (n = 1; *last; n++) {
-                        c = *last++;
+                    for (int n = 1; ; n++, last++) {
+                        c = *last;
+                        if (!c) break;
+
                         if (pp < ppmax) *pp++ = c;
                         if (c == '\a' || c == ESC || c == '\r') break;
                         if (skip || (c >= '0' && c <= '9')) {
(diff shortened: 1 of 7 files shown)


krader1961:
I'm not pushing for merging this over simply reverting the code to the original `for (n = 1; c = *last++; n++) {`. But I find it interesting that this version, which is equivalent to my pre-lint cleanup, makes the behavior clearer, IMHO. Especially with respect to the handling of where `last` points to if the last char inspected is not `\e` or `\r`. Which is just plain weird however it is implemented and deserves a comment.

src/cmd/ksh93/edit/edit.c
```diff
@@ -488,8 +487,10 @@ void ed_setup(Edit_t *ep, int fd, int reedit) {
                     int skip = 0;
                     ep->e_crlf = 0;
                     if (pp < ppmax) *pp++ = c;
-                    for (n = 1; *last; n++) {
-                        c = *last++;
+                    for (int n = 1; ; n++) {
+                        c = *last;
+                        if (!c) break;
```
siteshwar: `last` was incremented even when `c` is `null`. Are you sure it won't break any other use cases ?

krader1961: Yes, I'm certain this is okay. Specifically because of my comment on your PR. If we reach the nul char and increment `last` the `if (c ==0 || ...) last--;` after this loop undoes the increment. As I've said before, the structure of this is really weird. It's a little less weird with my change as it now looks more like what most people would write.

Note that when I woke up this morning I realized there is a case my change doesn't handle correctly. That is the explicit `continue` statement which would bypass the `last++`.  I'll push a fix for that in a second.

siteshwar:
@krader1961 I can't think of a condition where this will break. If you are confident about this change, go ahead and merge it. But please include a test case with it.

krader1961:
Okay, I've added a framework for `expect` based tests and included a regression test based on the reproducer in issue #776. Obviously I make no claim the `expect` framework can't be improved. But this at least provides a working example for how to add such tests.

krader1961:
I've merged this since a) I'm pretty confident the rewrite is equivalent to the original code and b) I've added our first `expect` based unit test to keep this from breaking in the future.

PR #777, opened 2018-08-09
The VENIX distro has been dead for two decades

krader1961:
Remove support for the build time `VENIX` symbol defined by that platform
since it has been dead for at least two decades. So remove references to
that symbol and the code predicated on it.


siteshwar:
LGTM

etscrivner:
LGTM

PR #781, opened 2018-08-13
Reduce the number of unit test scenarios

### LINKED ISSUE #780: Should we simplify the unit test framework by removing LANG=C for every test?
At the moment there is considerable duplication of effort when running unit tests due to running them with `LANG=C` and `LANG=en_US.UTF-8`. There are only a few tests that verify the correct behavior in both locales and they do so explicitly (e.g., the `locale.sh` unit test). All of the others can be expected to behave identically in those two locales because they only rely on behavior that is valid in the `C` locale.

Note that the `shcomp` variant is run with `LANG=en_US.UTF-8`. So it makes sense to drop that locale from the other test variants and just use `LANG=C` for the non-compiled script scenario. Or vice-versa. This will significantly reduce test run time without any loss of test coverage AFAICT.

krader1961:
The tests which depend on the locale (e.g., `locale.sh`) explicitly set
the locale to test compliance under different locales. For all other
tests running them under `LANG=C` and `LANG=en_US.UTF-8` is largely redundant.

Resolves #780

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -113,18 +113,28 @@ foreach testspec : all_tests
     testname = testspec[0]
     timeout = (testspec.length() == 2) ? testspec[1] : default_timeout
     parallel = serial_tests.contains(testname) ? false : true
+    if testname.endswith('.exp')
+        # The interactive `expect` based tests are highly sensitive to timing variations.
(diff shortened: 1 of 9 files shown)


krader1961:
I've merged this because it works and the current behavior does not justify the cost.

Issue #783, opened 2018-08-14
Re-enable the Travis CI "ubuntu" environment


krader1961:
As I write this Travis CI testing has been failing for approximately 24 hours due to issues beyond our control. Specifically, the `ubuntu` docker image installation has changed its dialog for configuring the timezone. Resulting in this install log output:

```
Setting up tzdata (2018d-1) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------
Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

PR #784, opened 2018-08-14
Skip interactive dialogue during package installation

siteshwar:
Certain packages like tzdata may ask for user input during installation,
this will block builds on Travis. We should skip it by settting
DEBIAN_FRONTEND to noninteractive.
    
Resolves: #783


krader1961:
LGTM but I can't help but think Docker should be automatically setting that env var.

PR #785, opened 2018-08-15
Remove `asolock()`

krader1961:
The `asolock()` function is theoretically useful. But we no longer
allow non-ksh code to link against the libast code used by ksh and thus
don't need to maintain the current API. Replace it with the POSIX thread
lock API.

The tproc.c and tthread.c tests are removed because they only test
behavior of the ASO code we're removing.


siteshwar:
LGTM

PR #787, opened 2018-08-15
Fix ASAN bug in `dthashchain()`

### LINKED ISSUE #786: ASAN bug in `dthashchain()`
This bug was found by the *src/cmd/tests/cdt/tannounce.c* unit test. The problem lies with the sequence of these two statements in `dthashchain()`:

```
_dtfree(dt, ll, type);
DTRETURN(obj, _DTOBJ(disc, ll));
```

The `_DTOBJ()` macro dereferences the `ll` pointer which was just freed. This is a result of trying to use a consistent idiom for returning from CDT functions. The pattern is reasonable for all the other return paths. It's not, however, correct for the case where `type & (DT_DELETE | DT_DETACH | DT_REMOVE)` is true. Whether simply deferring the free is sufficient is unclear but that's the solution I'm going to implement.

krader1961:
Fixes #786


siteshwar:
LGTM

PR #788, opened 2018-08-15
Add expect based tests for emacs keybindings

siteshwar:


krader1961:
FWIW, the unit test files at https://github.com/fish-shell/fish-shell/tree/master/tests which end in `.expect` are worth looking at. I wrote a lot of those. They show why it can be useful to explicitly write messages to stdout and stderr depending on whether a match was found or not. Since it is usually possible to resync on the next shell prompt that technique means the first failing match doesn't stop subsequent tests in the same script from being run.

See http://wiki.tcl.tk/10259 for help with special backslash sequences.
See http://wiki.tcl.tk/41378 for a discussion of strings enclosed in double-quotes versus braces.

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,210 @@
...
+
+# Send ctrl-p to check previous command
+send -- "\x10"
+expect -exact "set -o emacs"
+
+send -- "\x10"
+expect -exact "echo hello"
+
+# Send ctrl-n to check next command
+send -- "\xE"
```
krader1961: I'm pretty sure all chars specified using `\x` require two hexadecimal characters.

krader1961:
Rebase on top of the current master branch. I pushed a change that makes the stdout and stderr files for `expect` based tests optional. For me on macOS it fails with this stderr output:

```
<E> run_test[143]: Stderr for emacs.exp should have been empty:
eof; waiting on child process to exit
process 43366 exited with status 0 (CHILDKILLED SIGABRT SIGABRT)
eof
    while executing
"error $msg"
    (procedure "abort" line 2)
    invoked from within
"abort "eof""
    invoked from within
"expect "foobar""
    (file "/Users/krader/projects/3rd-party/ast/src/cmd/ksh93/tests/emacs.exp" line 165)
```

The starting point is the last line which tells us the `expect` script file name and the line number within that script which is responsible for the failure.

krader1961:
Also, even if TCL allows `\x` sequences to have a single digit (which is a perfectly reasonable "be permissive in what you accept" IETF RFC maxim) we should consistently write them with a leading zero.

src/cmd/ksh93/tests/emacs.exp
krader1961: Also, the `--` arg to stop parsing for args should be avoided unless needed, IMHO. I'm all for being explicit rather than implicit but this is one situation where the extra chars are more likely to cause the reader to ask "Why is this present?".

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,210 @@
...
+send -- "echo foobar\xF"
+expect -re ".*foobar"
+#
+# M-.       The last word of the previous command is inserted on the line.
+send -- "."
+expect "foobar"
+
+#
+## M-_       Same as M-.
+send -- "\x3"
```
krader1961: Your char code is incorrect. If you meant an [esc] char that should be `"\x1B"` or `"\033"`.

krader1961: And, of course, even though your current formulation gets the hex string for [esc] wrong it should not result in a SIGABRT from the ksh being tested. So please open an issue showing how to induce this failure mode.

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,210 @@
...
+send -- "_"
```
krader1961: This line contains a literal escape code point (e.g., `\x1B`). So do many other lines. Don't do that. Always specify such chars using a meta notation.

siteshwar:
I suspect there are timings issues in the `expect` based tests. Some tests are failing randomly.

krader1961:
I see the same failure on my macOS server that Travis reported for that environment. The `ksh` instance is dying. So probably not a timing issue.

krader1961:
If I start ksh interactively and press [meta-_] it dies due to signal SIGABRT. The abort is from libc `strcpy()`. Specifically, this statement in *history.c*:

```
if (s1 != string) strcpy(string, s1);
```

This means `s1` and `string` overlap. The backtrace:

```
  * frame #0: 0x00007fffa5c3867e libsystem_kernel.dylib`__kill + 10
    frame #1: 0x000000010f23ebe7 ksh`sh_fault(sig=6, info=0x00007fff509fe5b8, context=0x00007fff509fe620) at fault.c:167
    frame #2: 0x00007fffa5d19b3a libsystem_platform.dylib`_sigtramp + 26
    frame #3: 0x00007fffa5c38d43 libsystem_kernel.dylib`__pthread_kill + 11
    frame #4: 0x00007fffa5d26457 libsystem_pthread.dylib`pthread_kill + 90
    frame #5: 0x00007fffa5b9e420 libsystem_c.dylib`abort + 129
    frame #6: 0x00007fffa5b9e592 libsystem_c.dylib`abort_report_np + 181
    frame #7: 0x00007fffa5bc4f28 libsystem_c.dylib`__chk_fail + 48
    frame #8: 0x00007fffa5bc4f38 libsystem_c.dylib`__chk_fail_overlap + 16
    frame #9: 0x00007fffa5bc4f5a libsystem_c.dylib`__chk_overlap + 34
    frame #10: 0x00007fffa5bc5132 libsystem_c.dylib`__strcpy_chk + 64
    frame #11: 0x000000010f22579f ksh`hist_word(string="$KSH_VERSION", size=1024, word=-2) at history.c:912
    frame #12: 0x000000010f22e07c ksh`textmod(vp=0x00007fc20a408e00, c=95, mode=0) at vi.c:1867
    frame #13: 0x000000010f22b1d6 ksh`cntlmode(vp=0x00007fc20a408e00) at vi.c:529
    frame #14: 0x000000010f22a397 ksh`vigetline(vp=0x00007fc20a408e00, mode=-10) at vi.c:1032
    frame #15: 0x000000010f229075 ksh`ed_viread(context=0x00007fc20a800000, fd=0, shbuf="", nchar=1020, reedit=0) at vi.c:385
    frame #16: 0x000000010f24b069 ksh`slowread(iop=0x000000010f395a80, buff=0x00007fc20a801400, size=65536, handle=0x00007fc20a4030d0) at io.c:2054
    frame #17: 0x000000010f2cd65b ksh`sfrd(f=0x000000010f395a80, buf=0x00007fc20a801400, n=65536, disc=0x00007fc20a4030d0) at sfrd.c:235
    frame #18: 0x000000010f2c31b3 ksh`_sffilbuf(f=0x000000010f395a80, n=-1) at sffilbuf.c:97
    frame #19: 0x000000010f2cea7e ksh`sfreserve(f=0x000000010f395a80, size=0, type=0) at sfreserve.c:133
    frame #20: 0x000000010f26c5ed ksh`exfile(shp=0x000000010f39e8e0, iop=0x000000010f395a80, fno=0) at main.c:475
    frame #21: 0x000000010f26dbd1 ksh`sh_main(ac=1, av=0x00007fff50a073f8, userinit=0x0000000000000000) at main.c:317
    frame #22: 0x000000010f29c1a6 ksh`main(argc=1, argv=0x00007fff50a073f8) at pmain.c:41
    frame #23: 0x00007fffa5b0a235 libdyld.dylib`start + 1
    frame #24: 0x00007fffa5b0a235 libdyld.dylib`start + 1
```

krader1961:
The failure I saw occurs when the last command in my history is `$KSH_VERSION`. Adding some debug prints to show `s1` then `string` shows this:

```
WTF string 0x7fff56a049a5 12|$KSH_VERSION|
WTF string 0x7fff56a049a0 4|echo|
```

So, yes, this is yet another case where ksh is moving overlapping buffers and should not be using `strcpy()` or `memcpy()`. Yay for unit tests!

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,279 @@
+# Check emacs keybindings
+set pid [spawn $ksh]
+expect_prompt
+
+send "echo hello\r"
```
krader1961: You can drop this test. I only had it in mine because it was the very first such test we've written. We should probably move it, and any other test that does nothing more than establish baseline behavior, into a separate unit test whose name ensures it is run before any other interactive test.

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,279 @@
...
+#
+# ^L        Line feed and print current line
+#
+# M-^L      Clear the screen
+#
+
+# ^J        (New line) Execute the current line.
+#
+# ^M        (Return) Execute the current line.
+send "\x03"
```
krader1961: In general every test should ensure a new prompt is displayed and accounted for when it is done. That means every test should end with an `expect_prompt` call to ensure that is the case. We shouldn't be using [ctrl-C] to force a fresh prompt.

src/cmd/ksh93/tests/emacs.exp
siteshwar: It just acts as sample data for testing `ctrl-n` and `ctrl-p` bindings. I can drop the line below it however.

src/cmd/ksh93/tests/emacs.exp
siteshwar: Would it work if command on prompt is not executed ?

src/cmd/ksh93/tests/emacs.exp
krader1961: Ahh, yes, I see. Okay, just put add a comment that this is to inject a command in the history. Keep the `expect_prompt`.

src/cmd/ksh93/tests/emacs.exp
krader1961: The idea is to ensure we keep in sync with what's happening by inserting a `expect_prompt` everyplace we expect a new prompt to be issued. That way if we get out of sync, either because an unexpected prompt appears or the one we expect does not appear, we find out ASAP. So if your test is doing something like erasing the current char/word/line then we don't expect a new prompt so `expect_prompt` should not be called. But something like [ctrl-C] should cause a new prompt so we should verify that actually occurs.

krader1961: Also, I don't yet understand why line 255 is failing but when it does the input buffer looks like this:

```
[DEBUG] timeout; buffer= \r\nKSH PROMPT:15: ^[^\x08
```

Notice the escape char followed by a literal carat, then a backspace char.

siteshwar: Does macOS handle `alt-ctrl-v` differently ?

krader1961: No. The very handy `fish_key_reader` program (part of the fish shell) shows the expected output:

```
Press a key

              hex:   1B  char: \c[  (or \e)
(  0.225 ms)  hex:   16  char: \cV
bind \e\cV 'do something'
```

And it works fine when I type that sequence at an interactive shell prompt. There is something about the unit test that seems to cause the escape produced by [meta] to not be tied to the following [ctrl-V]. Which causes the [ctrl-V] to have it's usual "quote the next special char" meaning. So, yes, superficially this looks like a timing issue but that should be impossible given how the test is written.

If I delete all the tests prior to the [meta-ctrl-V] test it then works reliably.

krader1961: This seems to have something to do with how [ctrl-C] is handled.

src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,279 @@
...
+
+# Send ctrl-n to check next command
+send "\x0E"
+expect -exact "set -o emacs" {
+    puts "ctrl-n works"
+}
+
+# Take cursor at beginning of line with ctrl-a
+send "\x01"
+expect -exact "" {
```
krader1961: This is always true regardless of what operation was just done. The only way to test that [ctrl-a] works is to modify the buffer (e.g., by inserting a new char at the start of the line) then verifying the line looks like we expect.

krader1961:
Hold off on doing any more revisions, @siteshwar. I've added an `expect` command to compose control chars. And I'm changing your test code to conform to the test pattern I mentioned earlier. Each test should be of this form:

```
Send text
Redraw the line (optional)
Check the expected pattern is seen
Force a new prompt (optional)
Check for a new prompt (optional only if prev step skipped)
```

That pattern means that each test can assume a fresh prompt was issued and no input sent to the shell. This means that if a [ctrl-C], for example, is needed it's the responsibility of the previous test to do so and ensure a new prompt is present before the next test is run. Obviously tests of cursor movement and command line editing will need to skip the last two steps between each sub-test. But without that explicit coordination a test should be able to assume a fresh prompt with nothing on the command line.

siteshwar:
These changes have been polished in #792. So I am going to close this pull request.

Issue #789, opened 2018-08-16
ksh man page again - questionable doubled description --


aabatpurdue:
**Description of problem:**
There are 2 `set -B` description lines.  I do not use brace expansion so I do not know which, if either, is the correct one.

**Ksh version:**
Version A 93v-1328-g2e6acdf2

**How reproducible:**
Always.

**Steps to reproduce:**
1. man ksh
2. Skip to the description of the `set` command.
3.

**Actual results:**
` set`
`         ...`
`       -B  description 1`
`       -B  description 2`

**Expected results:**
Only 1 `-B`

**Additional info:**
Ain't got none (:-}).

PR #790, opened 2018-08-16
Remove duplicate description of 'set -B'

siteshwar:
Resolves: #789

### DIFF
--- a/src/cmd/ksh93/ksh.1
+++ b/src/cmd/ksh93/ksh.1
@@ -7045,10 +7045,6 @@ is used, the variable
 is not unset first.
 .TP 8
 .B \-B
-Enable brace pattern field generation.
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/ksh.1
```diff
@@ -7045,10 +7045,6 @@ is used, the variable
 is not unset first.
 .TP 8
 .B \-B
-Enable brace pattern field generation.
-This is the default behavior.
-.TP 8
-.B \-B
 Enable brace group expansion.  On by default.
```
krader1961: FWIW `set --help` has `Enable {...} group expansion. On by default.`. Which is why we need a single source of truth for such things.

krader1961:
LGTM

PR #791, opened 2018-08-16
Fix copying overlapping buffers in hist_word()

krader1961:
This came to light as a consequence of @siteshwar adding unit tests for
interactive emacs mode editing behavior.


siteshwar:
LGTM

PR #792, opened 2018-08-18
Add interactive tests for emacs mode

krader1961:
This is based on PR#788 by @siteshwar but with extensive changes to make
the tests more robust. In particular this structure of the unit test makes
it easier to add a new test. It also introduces a couple of helper
functions to make reading the log files easier.

### DIFF
--- a/src/cmd/ksh93/tests/emacs.exp
+++ b/src/cmd/ksh93/tests/emacs.exp
@@ -1,0 +1,422 @@
+# Check emacs keybindings
+set pid [spawn $ksh]
+expect_prompt
+
+# ==========
+# Enable emacs keybindings
+send "set -o emacs\r"
+expect_prompt
+
+# ==========
+# Populate the command history with a couple of commands.
+send "echo hello\r"
+expect "\r\nhello\r\n" {
+    puts "echo 1 works"
+}
+expect_prompt
+send "echo buh bye\r"
+expect "\r\nbuh bye\r\n" {
+    puts "echo 2 works"
+}
+expect_prompt
+
+# ==========
+# Send ctrl-p to retrieve previous command
+log_test_entry
+send [ctrl P]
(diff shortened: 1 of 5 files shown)


src/cmd/ksh93/tests/emacs.exp
```diff
@@ -0,0 +1,422 @@
...
+expect "\r\nbuh bye\r\n" {
+    puts "echo 2 works"
+}
+expect_prompt
+
+# ==========
+# Send ctrl-p to retrieve previous command
+log_test_entry
+send [ctrl P]
+send [ctrl L]
```
siteshwar: Why are you sending `ctrl-L` here ?

siteshwar: We can probably wrap this control sequence in a procedure called `redraw_prompt` for better readability.

siteshwar:
There was a timing issue in my original test, how was it fixed ?

src/cmd/ksh93/tests/emacs.exp
krader1961: Okay. Will do.

krader1961:
> There was a timing issue in my original test, how was it fixed ?

I can't answer that specific question because of the significant restructuring did. But I can answer the more general question. Part of the challenge to writing these types of tests is ensuring the tests are in sync with the buffered output. This is why many of the tests check for `: some text` or `\r\nsome text\r\n`. The first pattern is used after sending [ctrl-L] to redraw the command line and helps ensure we are matching the command line by anchoring it to the last two characters of the prompt (the `: `). The second pattern is bracketed by `\r\n` to ensure it matches the output of executing the `echo` or whatever command was run rather than the same text on the command line. Keep in mind that such matches implicitly consume any prior characters in the buffer.

Issue #793, opened 2018-08-18
Remove support for the `universe` command and related APIs


krader1961:
The concept of an OS "universe" dates back four decades to the 1980's when the AT&T and BSD implementations of UNIX were battling for dominance. Many vendors, such as Sequent Computer Systems and Sun Microsystems, naturally did not want to alienate any customers who preferred a particular flavor of UNIX. So they complicated their operating systems by adding the concept of a "universe". That is, a mode of behavior where both commands and APIs would behave like AT&T or BSD UNIX. You could change the behavior by changing the current "universe". This is reflected in the `ksh` implementation via symbols like `_cmd_universe`.

The cheap and easy availability of mechanisms like virtual machines and containers has made the concept of an OS "universe" irrelevant. Too, we're not even doing features tests for things like whether a `universe` command is available on the platform. Removing "universe" support is technically a non-backward compatible change. However it should be safe to do so since any environment that still requires explicit universe support by ksh is unlikely to ever include a `ksh` after the 93u+ code freeze back in the year 2012. The probability is so close to zero it's not worth measuring. And if they decide to do so it should be with the understanding that `ksh` no longer has any special support for the universe mechanism.

krader1961:
Not surprisingly there is a [Wikipedia article](https://en.wikipedia.org/wiki/Unix_wars) about the UNIX standards war that began in the 1980's.

Note: I worked for Sequent Computer Systems then IBM during the period being discussed. I supported both Sequent's DYNIX/3 (which favored BSD over AT&T) and DYNIX/ptx (which favored AT&T SysV over BSD). I am probably more intimately familiar with this issue than anyone else currently contributing to the maintenance of `ksh`. And I say good riddance to a feature which, like ISO-8859, caused more problems than it solved.

krader1961:
This [mailing list message](http://article.gmane.org/gmane.comp.standards.posix.austin.general/12525) by @ stephane-chazelas is relevant to this issue. That's because removing universe support may affect the default behavior of the `echo` command.

krader1961:
All "universe" support has been removed with the exception of the magic $PATH parsing for */xpg*, */bsd*, or */ucb* to select att or bsd emulation of the `echo` command.

PR #794, opened 2018-08-18
Remove symbol `_cmd_universe`

krader1961:
Related #793


siteshwar:
LGTM. I had not heard of universes until I started working on this project. This change should not affect any new operating systems that we care to support.

PR #795, opened 2018-08-18
Switch to semantic version numbering

siteshwar:
We have agreed to switch to semantic version numbering for next version
of ksh. This commit maintain backward compatibility with previous
versions of ksh by multiplying translated version number by 100 while
using version number in arithmetic context.


siteshwar:
I made a mistake while translating version string to number and need to redo this commit.

PR #796, opened 2018-08-19
Update code to convert version string to number

siteshwar:
We have switched to semantic version numbering and this commit modifies
code to convert version string to number in arithmetic context. Also,
add a test case for it and update changelog.

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -907,25 +907,52 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
 // This is invoked when var `.sh.version` is used in a numeric context such as
 // `$(( .sh.version ))`.
 //
-// TODO: Figure out what we want this to do with a semantic version string and
-// a version string that may include information about the git commit used for
-// the build. The original version converted the YYYY-MM-DD date stamp into the
-// number represented by YYYYMMDD. But note it would return a nonsensical value
-// if a date stamp in the correct form did not start exactly 10 characters
-// before the end of the string referred to by`e_version`.
 static Sfdouble_t nget_version(Namval_t *np, Namfun_t *fp) {
     UNUSED(np);
-    const char *cp = SH_RELEASE;
-    int c;
+    char *cp = strdup(SH_RELEASE);
     Sflong_t t = 0;
+    char *dash;
+    char *dot;
+    char *major_str, *minor_str, *patch_str;
+    int major, minor, patch;
     UNUSED(fp);
 
-    while ((c = *cp++)) {
-        if (c >= '0' && c <= '9') {
-            t *= 10;
-            t += c - '0';
-        }
+    // Version string in development version could be set like 2017.0.0-devel-1509-g95d59865
+    // If a '-' exists in version string, set it as end of string i.e. version string becomes
+    // 2017.0.0
+    dash = strchr(cp, '-');
+    if (dash) *dash = 0;
+
+    // Major version number starts at beginning of string
+    major_str = cp;
+
+    // Find the first '.' and set it to NULL, so major version string is set to 2017
+    dot = strchr(cp, '.');
+    if (!dot) {
+        // If there is no . in version string, it means version string is either empty, invalid
+        // or it's using old versioning scheme.
+        return 0;
     }
+
+    *dot = 0;
+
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/sh/init.c
```diff
@@ -907,25 +907,50 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
...
-// before the end of the string referred to by`e_version`.
 static Sfdouble_t nget_version(Namval_t *np, Namfun_t *fp) {
     UNUSED(np);
-    const char *cp = SH_RELEASE;
-    int c;
+    char *cp = strdup(SH_RELEASE);
     Sflong_t t = 0;
+    char *dash;
+    char *dot;
+    char buff[256];
```
krader1961: 256? It shouldn't need to be much larger than 4+2+2+1.

src/cmd/ksh93/sh/init.c
```diff
@@ -907,25 +907,50 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
...
+
+    // Patch number starts after second '.'
+    patch_str = dot + 1;
+
+    major = atoi(major_str);
+    minor = atoi(minor_str);
+    patch = atoi(patch_str);
+
+    // Pad minor and patch numbers with 0's if they are less than 10
+    snprintf(buff, sizeof(buff), "%d%02d%02d", major, minor, patch);
```
krader1961: This approach is fine but did you consider just adding up the values? That is, `t = major * 10000 + minor * 100 + patch;`?

krader1961:
LGTM although I would probably have avoided the int => string => int conversion at the end.

src/cmd/ksh93/sh/init.c
siteshwar: I will convert it to `32` bytes buffer.

src/cmd/ksh93/sh/init.c
siteshwar: You are right. We can get rid of `buff` array.

src/cmd/ksh93/sh/init.c
```diff
@@ -907,25 +907,46 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
...
+    // If a '-' exists in version string, set it as end of string i.e. version string becomes
+    // 2017.0.0
+    dash = strchr(cp, '-');
+    if (dash) *dash = 0;
+
+    // Major version number starts at beginning of string
+    major_str = cp;
+
+    // Find the first '.' and set it to NULL, so major version string is set to 2017
+    dot = strchr(cp, '.');
```
krader1961: I just noticed this sequence of statements isn't safe if the `SH_RELEASE` string doesn't contain a period in which case `dot` is NULL. This needs to be hardened against that scenario. We should probably just return zero because I can't imagine any other value that would not mislead or cause other problems.

krader1961: And I just tested this change and it does, as expected, generate a SIGSEGV on the next statement using my repo which does not have a semantic version tag.

siteshwar: I modified code to use old style of conversion if `.` is not present in version number.

src/cmd/ksh93/sh/init.c
```diff
@@ -907,25 +907,59 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
...
+    if (!dot) {
+        // If there is no . in version string, it means version string is either empty or
+        // it's using old versioning scheme.
+        int c;
+        while ((c = *cp++)) {
```
krader1961: This seems like overkill and won't produce the correct result unless someone overrides the automatic setting of `SH_RELEASE` to an old style version string. As it stands the code will use `93v` if there is no semantic version tag resulting in 93 for the numeric version. If there are no tags you get the git commit ID. For example, `#define SH_RELEASE "e4ca1b49"`. In which case this computes version 4149. That seems more likely to cause problems than utility. I know we can't protect against deliberate attempts to subvert this mechanism but it seems to me if that as of today if there is no semantic version string it should just return zero.

siteshwar: Fixed.

PR #797, opened 2018-08-21
Add interactive tests for vi bindings

siteshwar:


src/cmd/ksh93/tests/vi.exp
```diff
@@ -0,0 +1,670 @@
...
+control_mode
+send "j"
+redraw_prompt
+expect -ex ": echo buh bye" {
+    puts "j fetches next history item"
+}
+
+# [count]-  Equivalent to k.
+
+# [count][A If cursor is at the end of the line it is equivalent to / with string^set to the contents of the current line.  Otherwise, it is equivalent to k.
```
krader1961: Minor nit but it would be preferable for comment lines to be no longer than 100 chars to match the line length limit for C source.

src/cmd/ksh93/tests/vi.exp
```diff
@@ -0,0 +1,670 @@
...
+send "2"
+redraw_prompt
+expect -ex "1xzecho 21buh byXYe" {
+    puts "W takes cursor one alpha-numeric word forward"
+}
+
+exec mkdir foo1 foo2 foo3
+
+# [count]e  Cursor to end of word
+clear_prompt
```
krader1961: Every test that ends without leaving a fresh prompt should call `clear_prompt`. In other words, a test like this one should not begin with `clear_prompt`. It should be able to assume the shell has issued a prompt and is waiting for input. If `clear_prompt` is needed it should be tied to the end of the previous test. Also, you need to call `expect_prompt` after, or as part of, `clear_prompt` to ensure we're correctly keeping track of the prompts.

Lastly, each independent test like this one should begin with `log_test_entry` to inject markers in the *interactive.tmp.log* that help when debugging failures.

krader1961:
FYI, this fails on Cygwin and OpenBSD.

On OpenBSD the test to turn the command into a comment by inserting a `#` causes a SIGSEGV:

```
1   nv_associative + 18202
2   nv_associative + 18791
3   0x605e004
4   ed_viread + 6765
5   ed_viread + 1695
6   sh_trap + 46141
7   sh_run + 80087
8   sh_run + 36074
9   sfreserve + 1553
10  job_switch + 94622
11  job_switch + 101720
12  sh_addbuiltin + 131976
13   + 86
fish: 'src/cmd/ksh93/ksh' terminated by signal SIGABRT (Abort)
```

Yay for unit tests!

src/cmd/ksh93/tests/vi.exp
siteshwar: Calling `expect_prompt` inside `clear_prompt` starts causing tests to fail. I am still not sure why.

siteshwar:
It seems the test for `^` binding is consistently failing on macOS.

src/cmd/ksh93/tests/vi.exp
```diff
@@ -0,0 +1,927 @@
...
+proc insert_mode args {
+    global is_control_mode
+    send "i"
+    set is_control_mode 0
+}
+
+# Delete current line and change input mode to insert mode
+proc clear_prompt args {
+    send [ctrl C]
+    control_mode
```
krader1961: This causes problems, at least on macOS. What I'm seeing is that the [esc] is eaten, has no effect, which means the `i` sent by `insert_mode` is inserted in the command line. If I replace these two statements with `expect_prompt` the test reaches line 724 before failing on macOS. It fails on line 738 on Linux.

krader1961: Also, it should always be the case that a [ctrl-C] results in a new prompt that is already in insert mode. It should not be necessary to toggle between control and insert modes in this way. If doing this does make a difference that suggests a bug.

PR #799, opened 2018-08-21
Fix couple of defects identified by coverity

siteshwar:


src/cmd/ksh93/sh/init.c
```diff
@@ -942,6 +942,7 @@ static Sfdouble_t nget_version(Namval_t *np, Namfun_t *fp) {
 
     // Find the second '.' and set it to NULL, so minor version string is set to 0
     dot = strchr(minor_str, '.');
+    assert(dot != 0);
```
krader1961: Just `assert(dot)` is sufficient. If you want to be explicit please use `NULL` rather than zero.

krader1961:
LGTM

Issue #779, opened 2018-08-10
Macro `mbchar()` is incorrectly used


krader1961:
I noticed that `cppcheck` issues warnings such as this one:

```
[src/lib/libast/regex/regnexec.c:657] error (preprocessorErrorDirective):
 failed to expand 'mbchar', Wrong number of parameters for macro 'mbchar'.
```

There are at least two problems that warning brings to our attention. The most significant one is that there is code in the project which uses the new four arg call and code that uses the old one arg call. Instances of the latter being in *src/lib/libcmd/cut.c* and *src/lib/libcmd/wclib.c*.

The cppcheck warning tells us the places that should be invoking `mbchar()` with four args is treated as an error because the `#if ASTAPI(20130913)` test is false when it should be true. That is because we currently define it to an earlier date/value:

```
meson.build:feature_data.set('_API_ast', 20130625)
```

It looks like the beta branch was defining it as `99999999` but I need to confirm that.

krader1961:
Closer inspection shows that when building ksh `-D_API_ast=20130625` is passed to the compiler. But when building the rest of the AST code it defaults to `99999999`. But that only affects one file, *src/lib/libast/include/ast_intercept.h*, which I removed with commit  0390b4a9 to remove the syscall intercept mechanism.

TBD is whether we want to finish the work to use the API changes introduced `20130913` (c.f., *src/lib/libast/include/ast.h*) or drop those changes. And fix the call sites that expect to use those new APIs but which are in fact using the old APIs which results in lint warnings.

krader1961:
The specific error from `cppcheck` that caused me to open this issue seems to be incorrect. It appears to be using the second `#define mbchar` definition in *ast.h* despite the `#if ASTAPI(20130913)` statement being true and thus should cause only the first definition to be used. Nonetheless, there is still a problem. Some of the project code is using the new API(s) and some is using the old depending on if that `#if` is true or false. And that depends on whether the module is in the *src/lib/libast* tree or some other tree. That is in part due to how the `ASTAPI` macro is defined:

```
#define ASTAPI(rel) (_BLD_ast || !_API_ast || _API_ast >= rel)
```

The `_BLD_ast` symbol is true for the *src/lib/libast* and false elsewhere.

PR #800, opened 2018-08-22
Remove mb2wc, mbnchar, and mbnsize macros

krader1961:
As part of resolving issue #779 these three legacy macros can be removed or replaced by equivalent APIs to simplify the code.


krader1961:
I've gone ahead and merged this since the changes are borderline mechanical and the sooner we remove the legacy locale macros the better.

PR #802, opened 2018-08-22
Locale shcomp

krader1961:
Fail the test if `shcomp` fails. Workaround `iconv()` failing in the `C` locale on some platforms.


krader1961:
I've merged this since the two changes are trivial and self-evidently correct.

PR #807, opened 2018-08-25
Avoid recalculating version number string to number

siteshwar:
Version number string should be converted to number only once as it is
not going to change in subsequent calls of `nget_version`.

Related: #335

### DIFF
--- a/src/cmd/ksh93/sh/init.c
+++ b/src/cmd/ksh93/sh/init.c
@@ -909,8 +909,12 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
 //
 static Sfdouble_t nget_version(Namval_t *np, Namfun_t *fp) {
     UNUSED(np);
+
+    // We should not need to convert version number string every time this function is called
+    static Sflong_t version_number  = -1;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/init.c
```diff
@@ -910,52 +910,56 @@ static_fn char *get_version(Namval_t *np, Namfun_t *fp) { return (nv_getv(np, fp
...
-
-    // Find the first '.' and set it to NULL, so major version string is set to 2017
-    dot = strchr(cp, '.');
-    if (!dot) {
-        // If there is no . in version string, it means version string is either empty, invalid
-        // or it's using old versioning scheme.
-        free(cp);
-        return 0;
-    }
+    if (version_number == -1) {
```
krader1961: Personally, I think an early return is cleaner:

```
if (version_number != -1) return version_number;
```

Then 99% of the function no longer needs to be an indented block.

krader1961: Also, if you put this at the top of the function with the other vars defined below it then you avoid the need to pointlessly `strdup(SH_RELEASE)` and free the buffer even though we've already calculated and cached the return value.

PR #815, opened 2018-08-28
Lint

krader1961:


src/cmd/ksh93/sh/deparse.c
```diff
@@ -305,6 +306,7 @@ static_fn void p_tree(const Shnode_t *t, int tflags) {
                 }
                 p_arg(&(t->lst.lstlef->arg), ' ', 0);
                 if (t->tre.tretyp & TBINARY) {
+                    assert(cp != NULL);
```
siteshwar: Why ?

src/cmd/ksh93/sh/parse.c
```diff
@@ -864,8 +865,14 @@ static_fn Shnode_t *funct(Lex_t *lexp) {
     }
     shp->st.staklist = (struct slnod *)slp;
     last = fctell();
-    fp->functline = (last - first);
-    fp->functtre = t;
+    // It should be impossible for `fp` to be NULL since a longjmp() should not occur before it has
+    // been assigned a non-NULL value in the `if (jmpval == 0) {...}` block above.
+    if (!fp) {
```
siteshwar: We can avoid this `if` block by adding `assert(fp != NULL)`.

src/cmd/ksh93/sh/deparse.c
krader1961: Because the next statement, `sfputr(outfile, cp, ' ');`, uses that var and expects it to be non-NULL. However, the way the code is structured there is theoretically a path that reaches here wherein `cp` is not set to a non-NULL value. The assert tells the compiler we expect it to be non-NULL and silences the warning.

src/cmd/ksh93/sh/parse.c
krader1961: Unfortunately cppcheck isn't smart enough to recognize that keeps the next two statements from dereferencing a NULL pointer. So rather than add two `// cppcheck-suppress` comments (one for each reference) I opted for this structure.

krader1961:
Merged

PR #816, opened 2018-08-29
Remove astapi

krader1961:


src/lib/libast/cdt/cdtlib.h
```diff
@@ -110,9 +86,6 @@ typedef struct _dtlib_s {
          ? (*(dt)->disc->eventf)((dt), DT_ANNOUNCE | (ty), (ob), (dt)->disc)            \
          : 0)
 
-/* map bits for upward compabitibility */
-#define DTTYPE(dt, ty) ((dt)->typef ? (*(dt)->typef)((dt), (ty)) : (ty))
```
siteshwar: What was the function of this macro ?

krader1961: It supported code written before the CDT API changed in 2011-11-11. See the `_dttype2005()` function that is also removed. It's a no-op identity function for ksh.

PR #821, opened 2018-08-30
Extend tests for read builtin

siteshwar:
Split tests for read builtin in separate file and add new tests for
different options.


src/cmd/ksh93/tests/read.exp
```diff
@@ -30,6 +30,18 @@ expect -ex "\r\ngimme input > "
 expect -ex "\r\ngot ''\r\n"
 expect_prompt
 
+#-p prompt Write prompt on each line before reading.
+send "read -p 'custom_prompt> ' foo\r"
+expect -ex "custom_prompt> "
```
krader1961: Writing `expect` based tests is hard. Here you need to consider that the input buffer will contain `custom_prompt>` twice. This will match the first instance which is from the previous `send`, not the actual prompt from executing the `read`. What I often do is deliberately cause the test to fail. For example, by introducing a typo in the expect pattern. Then looking at the log file to see what was actually in the buffer and tailoring the pattern to match what we're really interested in. Which in this case is the actual prompt.

What you really want here is `expect -re "\r\ncustom_prompt> $"`.

src/cmd/ksh93/tests/read.exp
```diff
@@ -30,6 +30,18 @@ expect -ex "\r\ngimme input > "
...
+send [ctrl C]
```
krader1961: You need an `expect_prompt` after this to verify we get a new prompt.

src/cmd/ksh93/tests/read.exp
```diff
@@ -30,6 +30,18 @@ expect -ex "\r\ngimme input > "
...
+
+#-v When reading from a terminal the value of the first variable is displayed and used as a default
+# value.
+send "set foo bar\r"
+send "read -v foo\r"
+expect -ex "bar"
```
krader1961: Same as above. Here you're actually matching the "bar" from `set foo bar` which does not do what you intended. I'd also recommend using `read -p 'custom_prompt> ' -v foo` to make matching the prompt and default value clearer.

krader1961:
Other than the problems I noted with the *read.exp* tests the rest of the change looks fine.

PR #822, opened 2018-08-31
Remove `mb_xfrm`

krader1961:
The `mb_xfrm` indirection var is no longer needed. The one use in the
`mbcoll()` macro can be eliminated as the test is always true and the
value is now a constant.


krader1961:
Merged because it's trivial and I want to move ahead with removing the now pointless `mbt...()` macros that simply wrap their `mb...()` counterpart.

Issue #825, opened 2018-09-01
printf "%#d" is broken


siteshwar:
**Description of problem:**
`printf "%#d" 1000` gives wrong output.

**Ksh version:**
2017.0.0-devel-1584-g9e9c97a3-dirty

**How reproducible:**
Always

**Steps to reproduce:**

```
printf "%#d" 1000
```

**Actual results:**
10k

**Expected results:**
1.0k

**Additional info:**
This bug is not present in `93v- 2014-12-24` release.

siteshwar:
This is a regression introduced by d2aa35f16946e1f5a9ac92851f91d8d4a4e9a5ff. Decimal character is not being printed.

PR #826, opened 2018-09-02
Fix format string to print decimal character correctly

siteshwar:
Resolves: #825


krader1961:
LGTM. /facepalm on my part but obviously the original code did not fully support non-ASCII using localess since it assumed the radix char was an ASCII char (presumably just `.` or `,`). It's a safe bet there are lots of similar bugs.

PR #827, opened 2018-09-02
Fix obvious malloc size calculation bug

### LINKED ISSUE #824: Malloc bug found by cppcheck in `sh_argopts()`
Cppcheck warned us:

```
[src/cmd/ksh93/sh/args.c:459] warning (signConversion):
 Suspicious code: sign conversion of - in calculation, even though - can have a negative value
```

Which is from this (slightly abridged) block of code from `sh_argopts()`:

```
int keys = 0;
size_t nodesize;
if (keylist) {
        for (cp = keylist; (c = *cp); cp++) {
        if (c == ',') keys++;
        }
        keys++;
} else {
        keylist = Empty;
}
nodesize = sizeof(struct Node) + (keys - 1) * sizeof(Namval_t *);
```

The problem is that `keys` may be zero in in that last statement. I've confirmed empirically by adding a debug print that this does actually occur. Resulting in `nodesize` being 16 when it should be a minimum of 24 (the `sizeof(struct Node)`). Note that `nodesize` is used in the next statement (not shown) to malloc a buffer. Here is one way to induce this bug (from the *arrays.sh* unit test):

```
ar=(9 11 6 3.5 22)
set -s -A ar
```

TBD is the correct fix:

1) Don't subtract one from `keys` when calculating `nodesize`.
1) Treat `keys - 1` as zero  when `keys` is zero.
1) Something else.

krader1961:
The affected code is a poster child for why allocating a single buffer
then setting pointers in the structure at the head of the buffer to
locations within the buffer is a bad idea. Figuring out whether the
allocation is the correct size is borderline impopssible.

Resolves #824

### DIFF
--- a/src/cmd/ksh93/sh/args.c
+++ b/src/cmd/ksh93/sh/args.c
@@ -236,7 +236,7 @@ int sh_argopts(int argc, char *argv[], void *context) {
     int setflag = 0, action = 0, trace = (int)sh_isoption(shp, SH_XTRACE);
     Namval_t *np = NULL;
     const char *sp;
-    char *keylist = 0;
+    char *keylist = NULL;
     int verbose, f, unsetnp = 0;
     Optdisc_t disc;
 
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/sh/args.c
```diff
@@ -441,25 +441,28 @@ int sh_argopts(int argc, char *argv[], void *context) {
...
                 int i, c, keys = 0;
-                size_t nodesize;
+
                 if (keylist) {
                     for (cp = keylist; (c = *cp); cp++) {
                         if (c == ',') keys++;
                     }
                     keys++;
                 } else {
-                    keylist = Empty;
```
siteshwar: There are some places in code which directly compare against value of `Empty`. Have you verified such checks won't be affected by this change ?

src/cmd/ksh93/sh/args.c
```diff
@@ -476,7 +479,7 @@ int sh_argopts(int argc, char *argv[], void *context) {
                 strcpy(sp->keys[0], keylist);
                 cp = (char *)sp->nodes;
                 for (c = 0; c < arp->nelem; c++) {
-                    if (keylist != Empty && *keylist != ':') {
+                    if (*keylist && *keylist != ':') {
```
siteshwar: Ok, I see you have fixed a check here.

src/cmd/ksh93/sh/args.c
krader1961: Yes, for this specific change. In general it looks like most, if not all, such uses were an "optimization" to avoid having lots of empty strings created (which hasn't been an issue for C compilers for at least three decades) or simply used as a silly stylistic idiom that obfuscates more than it illuminates. The only reason I haven't removed it everywhere is I haven't taken the time to verify it isn't a sneaky way to distinguish between a dynamically allocated string versus a static string for purposes of deciding if it can be passed to `free()`. Also, there is `EmptyStr` which is functionally identical. So I've been nibbling away at removing these references as I come across them individually, like this one, and am able to determine the transformation is safe.

src/cmd/ksh93/sh/args.c
```diff
@@ -441,25 +441,28 @@ int sh_argopts(int argc, char *argv[], void *context) {
...
-                                           (sizeof(char *) + 1) * keys +
-                                           (arp->nelem + 1) * (nodesize + sizeof(void *)) + c + 3);
+                // This used to multiply by `(keys - 1)` but `keys` can be zero which means the
+                // nodesize can be less than `sizeof(struct Node)` which is obviously wrong.
+                // Whether multiplying by `keys` is correct is unclear.
+                // See issue #824.
+                size_t nodesize = sizeof(struct Node) + keys * sizeof(Namval_t *);
+                struct Sort *sp =
+                    malloc(sizeof(struct Sort) + strlen(keylist) + (sizeof(char *) + 1) * keys +
+                           (arp->nelem + 1) * (nodesize + sizeof(void *)) + c + 3);
```
krader1961: This entire size calculation is suspicious (e.g., `(sizeof(char *) + 1)`)but the final `+ 3` term especially so. Three what? As written it means three bytes but why does this need three additional bytes? There are many other places in the code with a similar pattern of allocating a huge buffer that holds many distinct structures with pointers from one to another within the malloced block.

krader1961:
Merged even though this is far from an optimal solution. The optimal solution is to change very place that uses this pattern to explicitly allocate each structure individually and correspondingly free all of them when the main structure is freed.

PR #829, opened 2018-09-03
Remove `alarm` builtin

### LINKED ISSUE #646: Undocumented `alarm` builtin -- whither its future?
While investigating why `read -t0.01` might not time out (see this PR [comment](https://github.com/att/ast/pull/645#discussion_r199972130)) I stumbled across the *src/cmd/ksh93/bltins/alarm.c* module which implements the undocumented `alarm` builtin. When I googled it I found only one relevant result: https://blog.fpmurphy.com/2008/12/mysterious-ksh93-alarm-builtin.html. Notice that was written in Dec 2008 and refers to ksh93s. The last sentence of the article is

> Use alarm with caution since Dave Korn has also stated that "alarm hasn’t been documented because it is not working in all cases."  The examples shown in this post were tested using ksh93 version s.

If a feature is that old and still undocumented it may be time to drop it. Or at least not enable it by default (which is currently the case) and add a "here be dragons" warning at the top of the module.

krader1961:
The `alarm` builtin has been present and undocumented for at least 10
years. Dating back to the ksh93s release if not earlier since the 93s
beta release appears to have been made on 2006-12-29. Since it is old,
undocumented, and a public statement by the AST team suggests it is not
ready for prime time it is time to remove it in the interest of simplifying
the project.

Resolves #646


src/cmd/ksh93/sh/fault.c
```diff
@@ -387,7 +387,6 @@ void sh_chktrap(Shell_t *shp) {
         }
     }
     if (!shp->sigflag) return;
-    if (shp->sigflag[SIGALRM] & SH_SIGALRM) sh_timetraps(shp);
```
siteshwar: Both `SH_SIGALRM` and `SH_SIGTSTP` are defined as `0200`. I suspect there are going to be side effects of this change.

krader1961:
Merged

PR #830, opened 2018-09-03
Lint: early exit from `sh_exec()`

krader1961:
This clarifies and simplifies the code in `sh_exec()` by performing an
early exit. This was brought to my attention by the clang static analysizer:

src/cmd/ksh93/sh/xec.c:893:5: prefer early exits and continue
[convention|P3] Use early exit/continue to simplify code and reduce indentation


krader1961:
Merged

PR #831, opened 2018-09-03
Extend tests for printf builtin

siteshwar:
Split tests for printf builtin in a separate file and add new tests.


krader1961:
FYI, even when applied on top of all changes merged since you opened this it still fails on macOS and other BSD based systems with errors like

```
<E> printf[5]: %b does not recognize \a
```

That's due to a difference in how `od` behaves. On BSD systems the lines are blank padded on the right. Changing the pattern checking for success to allow for those spaces fixes the problem. For examaple, `grep -q "007$"` => `grep -q "007 *$"`.

krader1961:
LGTM other than the BSD failures due to how `od` works.

PR #834, opened 2018-09-04
Extend tests for pwd builtin

siteshwar:


siteshwar:
One new test fails on macOS with this error:

```
<E> builtins[647]: /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/ksh.builtins.XXXXXXX.u7j5zkcn != /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/ksh.builtins.XXXXXXX.u7j5zkcn: pwd -P should print physical path.
```

Is that because there are more than one symlink in physical path ?

krader1961:
On macOS */var* is a symlink to *private/var*. And the value returned by `mktemp -d` returns a path with the symlink unresolved and that is what we `cd` into when running the tests. So it is expected that `pwd` and `pwd -P` will not be the same. You should be able to produce the same failure on Linux by making */tmp* a symlink to a different directory.

src/cmd/ksh93/tests/builtins.sh
```diff
@@ -629,6 +629,23 @@ then
     rmdir "$pwd/f1"
 fi
 
+OLDPWDSYMLINK="$TEST_DIR/oldpwdsymlink"
```
krader1961: If ksh had a `realpath` command I would recommend using it to make this test work even if `$TEST_DIR` contains symlinks (like it does on macOS).  Instead the simplest "fix" is to probably do `cd $(pwd -P)` here and then use `$PWD` where you have `$TEST_DIR`.

PR #838, opened 2018-09-05
Fix out of bound error while printing failures

siteshwar:


src/cmd/ksh93/tests/util/run_test.sh
```diff
@@ -214,7 +214,7 @@ then
         run_interactive $(( i == 3 ))
         status=$?
         [[ $status -eq 0 ]] && break
-        log_info "Iteration $i of interactive test '$test_name' failed on line ${failure_lines[-1]}"
+        log_info "Iteration $i of interactive test '$test_name' failed on line ${failure_lines[$i]}"
```
krader1961: Isn't this change a no-op? Subscript -1 is the last entry in the array which should always be the failure line of the most recent iteration. I have no objection if you want to merge this on the basis of being less confusing to someone not familiar with the behavior of negative indexes. But if it's actually fixing something then I'd like to know what you observed since this should be a no-op.

krader1961: I see only one way this could fail. If the `typeset line=$(sed ...)` failed to assign a value to `line`. But in that case both variants of this line will fail. The correct fix is to change `failure_lines+=($line)` to `failure_lines+=(${line:-0})` to ensure there is a, possibly dummy, line number for the iteration.

krader1961: Too this change isn't a no-op. It introduces a different problem because `$i` is starts at one but array indexes start at zero.

siteshwar: I saw an index range out of bound error while trying to write a test case. This patch may not be correct, but there is definitely a problem here.

krader1961: Yes, there is. As I said in my second comment if the test fails without reporting a line number then the array will be empty. The fix is to change `failure_lines+=($line)` to `failure_lines+=(${line:-0})`.

krader1961:
LGTM

Issue #682, opened 2018-07-13
Tab completion chokes on ~/some


siteshwar:
@kusalananda reports at https://github.com/att/ksh/issues/3:

Version AJM 93u+ 2012-08-01 running on OpenBSD 6.1 (but have been observed elsewhere).

Tab completion does not work in Vi mode (only) for completing a subdirectory name in one's home directory, if tab is pressed after the fourth character of the directory name:

$ cd ~/loca

Pressing tab at this point starts moving the cursor to the left, over the entered text, eventually leaving the command line reading

cd /home/

This does not occur if the tab completion is initiated after the third or fifth entered character.

This also affects completion of file names.

If ~username is used, then tab completion fails if pressing tab when the length of the username plus the entered length of the name of the file/dir is four characters:

ls ~kk/dm

kusalananda:
@siteshwar  Thanks for moving it, sorry for my confusion.

siteshwar:
@kusalananda No worries. fwiw https://github.com/att/ksh/ has been deleted to avoid any confusions in future.

kusalananda:
This bug seems to have been resolved in the development version.  The version I tried is `Version A 93v-1400-g3d8a9fe4`.

I'm not closing the issue since I don't know whether you'd like to do some other tests for this first (or write regression tests).

PR #839, opened 2018-09-05
Regression test for tab completion choking in vi mode

siteshwar:
Add regression test for tab completion choking on fourth character of
dirname in ~/<dirname> in vi mode.

Resolves: #682

### DIFF
--- a/src/cmd/ksh93/tests/vi.exp
+++ b/src/cmd/ksh93/tests/vi.exp
@@ -949,6 +949,15 @@ expect -re ".*\rKSH PROMPT:\[0-9\]+: $" { }
 send "\r"
 expect_prompt
 
+# Regression tests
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/tests/vi.exp
```diff
@@ -949,6 +949,14 @@ expect -re ".*\rKSH PROMPT:\[0-9\]+: $" { }
 send "\r"
 expect_prompt
 
+# Regrssion tests
+# https://github.com/att/ast/issues/682
+send "mkdir ~/local\r"
```
krader1961: Probably worth a comment that this depends on the *run_tests.sh* test runner setting `$HOME` to the temp test dir to avoid scribbling on the users home dir.

krader1961:
LGTM

PR #840, opened 2018-09-05
Add basic interactive tests for job handling

siteshwar:
Also, split tests for handling job related functions in a separate
file.

### DIFF
--- a/src/cmd/ksh93/tests/jobs.exp
+++ b/src/cmd/ksh93/tests/jobs.exp
@@ -1,0 +1,71 @@
+# Interactive tests for jobs, bg, fg and related builtins
+set pid [spawn $ksh]
+expect_prompt
+
+# ======
+# Ctrl-Z should stop current process
+log_test_entry
+send "cat\r"
+sleep 0.5
+send [ctrl Z]
+expect -re ".*Stopped.*cat\r\n" {
+    puts "Ctrl-z stops cat processs"
+}
+expect_prompt
+
+# ======
+# Test if jobs builtin lists stopped process
+log_test_entry
+send "jobs\r"
+expect -re ".*Stopped.*cat\r\n" {
+    puts "jobs lists stopped cat process"
+}
+expect_prompt
+
+# ======
+# bg should put last stopped process in background
+log_test_entry
+send "bg\r"
+expect -re ".*cat&\r\n"
+expect_prompt
+
+send "jobs\r"
+expect -re ".*Stopped.*SIGTTIN.*cat\r\n" {
+    puts "bg brings last stopped process to background"
+}
+expect_prompt
+
+# ======
+# fg should put last stopped process in foreground
+log_test_entry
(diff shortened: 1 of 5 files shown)


src/cmd/ksh93/tests/jobs.exp
```diff
@@ -0,0 +1,61 @@
...
+# ======
+# Ctrl-Z should stop current process
+log_test_entry
+send "cat\r"
+sleep 0.5
+send [ctrl Z]
+expect -re ".*Stopped.*cat.*" {
+    puts "Ctrl-z stops cat processs"
+}
+
```
siteshwar: @krader1961 Why putting `expect_prompt` here causes tests to fail ?

krader1961: Because your previous regex captures the prompt. Specifically, the trailing `.*`. Replace that wit h`\r\n`. Also, for reasons I haven't figured out the "Stopped" message isn't appearing in the expect buffer when I run this on macOS.

src/cmd/ksh93/tests/util/run_test.sh
```diff
@@ -214,7 +214,7 @@ then
         run_interactive $(( i == 3 ))
         status=$?
         [[ $status -eq 0 ]] && break
-        log_info "Iteration $i of interactive test '$test_name' failed on line ${failure_lines[-1]}"
+        log_info "Iteration $i of interactive test '$test_name' failed on line ${failure_lines[$i]}"
```
krader1961: Don't forget to revert this portion of the change.

src/cmd/ksh93/tests/jobs.exp
```diff
@@ -0,0 +1,71 @@
...
+set pid [spawn $ksh]
+expect_prompt
+
+# ======
+# Ctrl-Z should stop current process
+log_test_entry
+send "cat\r"
+sleep 0.5
+send [ctrl Z]
+expect -re ".*Stopped.*cat\r\n" {
```
krader1961: This is failing sometimes on my macOS systems and on Travis macOS. The "stopped" message isn't being written. But it works reliably when I do this by hand. We'll probably need to blacklist this on macOS.

krader1961: This fails on macOS because the  interrupts cat's `read()` on stdin and it doesn't retry. Which results in the `cat` dying. You can work around that difference by using a fifo:

```
send "mkfifo fifo\r"
expect_prompt
send "cat < fifo\r"
```

Then change the pattern here and below from `cat\r\n` to `cat < fifo\r\n`. However, this also requires other changes. I recommend just blacklisting this entire unit test (in meson.build) on macOS with a "TODO" comment for now.

krader1961:
LGTM

It's unfortunate we need to blacklist tests like this but until we get more people helping to improve the project this is going to be necessary.

PR #849, opened 2018-09-09
Extend tests for command builtin

siteshwar:
Split tests for command builtin in separate file and add new tests.


krader1961:
LGTM. Seems like a reasonable thing to do. The existing unit test scripts are a bit ad hoc and try to do too much so this type of refactoring is a net improvement IMO.

Issue #851, opened 2018-09-09
self-document getopts mangling alternative usages


perette:
When using self-documenting (long-form) getopts strings, alternate usage forms are now getting mangled.  This previously worked, I believe fairly recently so.

**Ksh version:**
Version A 2017.0.0-devel-1694-g4e123c2

**How reproducible:**
Seems to be reliably troublesome

**Steps to reproduce:**
1.  Use --help or --man option on script using self-documenting get-opts that has multiple usage forms.  Or, cut and paste this into a terminal:
```
getopts -a test $'
[-]
[Test?Test of getopts]
[t:test?Test option]

use form one
use form two

[-Perette?Perette Barella]' baka '--help'
```

**Actual results:**
```
Usage: test [ options ] use form one
   Or:use form one
testuse form one
[use form one
optionsuse form one
] use form two
OPTIONS
  -t, --test      Test option
```

**Expected results:**
```
Usage: test [ options ] use form one
   Or: test [ options ] use form two
OPTIONS
  -t, --test      Test option
```
(This output was produced by an earlier ksh, both Version M 1993-12-28 s+ and Version AJM 93u+ 2012-08-01, but I think it was working much more recently than those version numbers indicate or I think I'd have encountered this earlier.)

**Additional info:**
Broken by commit ce5dabe on August 7.

siteshwar:
@perette Thank you for the perfect bug report.

> Broken by commit ce5dabe on August 7.

To be more specific this part of change has introduced this regression:

```
-#define ERROR_translating() (error_info.translate && (ast.locale.set & (1 << AST_LC_MESSAGES)))
+#define ERROR_translating() error_info.translate
```

perette:
I'm looking at the original commit, and I'm also concerned about the changes to `init.c`; it looks to my eye like too much got dyked out-- the `lctype++` got removed, which appears to be part of encoding tracking so if character type changes on-the-fly, Ksh adapts existing shell variables to the new encoding.  Not incrementing this variable will break that.  Thoughts?

Additionally, if `sh_lexstates` is truly a variable, then the `else` clause that resets it to default got whacked.  However, I think now that the AST crap is gone, this is now a constant--in which case, this should get stripped out and `sh_lexrstates` should be used instead (or better, `sh_lexrstates` should become the new `sh_lexstates` to get rid of the unexplained 'r').

krader1961:
[...]

PR #852, opened 2018-09-09
Fix definition of ERROR_translating macro

siteshwar:
There was a regression introduced by changes in ce5dabe and d2aa35f that
deal with removing AST specific locale subsystem. Fix it by using
standard `LC_MESSAGES` symbol.

Resolves: #851


siteshwar:
@krader1961 If this change looks correct to you, I will add a test case with it.

@perette Your review will be greatly appreciated.

perette:
@siteshwar Your pending merge fixes reported #851 behavior.

Issue #856, opened 2018-09-11
Integrate POSIX tests from `yash` shell


siteshwar:
`yash` shell has a [set of tests](https://github.com/magicant/yash/tree/trunk/tests) that can be used to verify POSIX compatibility. We should investigate if they can be used to test ksh.

siteshwar:
Since the code is GPL'ed we can not directly integrate these tests in our code. I will try to integrate them in our [application compatiblity](https://github.com/att/ast/blob/10d173b93679f3697c926663bf564083443b4480/app-compat-travis.yml) tests.

siteshwar:
fwiw I should mention that ideally any broken tests should be fixed in `yash` upstream. If they are not willing to take patches from us, we will have to maintain a fork of these tests.

PR #857, opened 2018-09-11
Add `yash` tests in application compatibility tests

siteshwar:
`yash` shell has a set of tests that can be used to test POSIX
compatibility. We should start running them in our application
compatibility tests.

Currently all the tests are not passing, so I have enabled only subset
of these tests. Failing tests should be investigated separately.

Resolves: #856

### DIFF
--- a/app-compat-travis.yml
+++ b/app-compat-travis.yml
@@ -6,10 +6,18 @@ services:
 
 matrix:
     include:
-        - env:
+        - os: linux
+          env:
             - PROJECT_NAME=kshdb
               GIT_URL="https://github.com/rocky/kshdb.git"
               COMMANDS="git checkout att-ksh93v-testing; ksh ./autogen.sh; make test"
+
+        - os: linux
+          env:
+            - PROJECT_NAME=yash
+              # This fork contains list of tests and a test runner script
+              GIT_URL="https://github.com/siteshwar/yash.git"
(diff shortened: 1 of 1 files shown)


app-compat-travis.yml
```diff
@@ -6,19 +6,26 @@ services:
...
+        - os: linux
+          env:
             - PROJECT_NAME=kshdb
               GIT_URL="https://github.com/rocky/kshdb.git"
               COMMANDS="git checkout att-ksh93v-testing; ksh ./autogen.sh; make test"
 
+        - os: linux
+          env:
+            - PROJECT_NAME=yash
+              GIT_URL="https://github.com/siteshwar/yash.git"
```
krader1961: Should this have a "TODO" comment explaining why we're using your personal clone and that we hope to use the unmodified yash tests?

krader1961:
LGTM

app-compat-travis.yml
siteshwar: I have added a list of tests to run and testrunner through https://github.com/siteshwar/yash/commit/fbe0139c2dbe0242fce4f00978cefd35caba42a8. I can leave a note here to clarify why I am using my clone.

siteshwar:
Thanks @magicant for writing reusable tests.

app-compat-travis.yml
siteshwar: and opened #859 to investigate failing tests.

PR #865, opened 2018-09-12
Remove var `viraw`

### LINKED ISSUE #864: Eliminate `viraw` and `SH_VIRAW`
While working on lint cleanups I noticed that `viraw` is now a constant thanks to oclint warning about constant `if()` expressions. We no longer support the vi CLI editor mode without support for putting the tty in "raw" mode. So eliminate the symbol.

krader1961:
While working on lint cleanups I noticed that viraw is now a constant
thanks to oclint warning about constant if() expressions. We no longer
support the vi CLI editor mode without support for putting the tty in
"raw" mode. So eliminate the symbol. And update the documentation to
explain that the option no longer has any effect.

Fixes #864


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ None at this time.
 
 ## Notable non-backward compatible changes
 
+- Vi raw mode is now the default and cannot be disabled (issue #864).
```
siteshwar: It does not look like a backward incompatible change to me. I see this entry in `RELEASE` file:

```
01-11-26 +The default has compilation mode has been changes so that
              viraw mode will always be on.
```

I can not disable `viraw` mode in version of `ksh` that is shipped with RHEL.

krader1961: Note that "the default compilation mode" simply means that viraw was enabled by default. The second clause of that sentence seems to be independent of the first clause. And looking at the source in the legacy ast.master and ast.beta branches it is not at all obvious you can't disable viraw behavior. Nonetheless, I'm happy to move this statement about `viraw` to the "Other significant changes" section.

Issue #866, opened 2018-09-12
`ninja -v ...` shows 64 of 484 files are compiled twice --D


aabatpurdue:
**Description of problem:**
This is mostly a FYI.  It seems that 64 of the 484 files are compiled twice. Don't know if this can be changed or even if it should be changed.  I have almost nil knowledge of `meson` and `python3` so am not a good candidate for making any changes.

**Ksh version:**
The clone is from September 3.

**How reproducible:**
Every time.

**Steps to reproduce:**
1. `ninja -v ....`
2.
3.

**Actual results:**
64 of 484 files are compiled twice.

**Expected results:**
In a perfect world, only 420 files should be compiled. (:-})

**Additional info:**
I did a `cmp` of one pair of files and they compared equal.  Since the remainder are probably equal also, it would be nice if these could be put in a directory by themselves and used twice.

siteshwar:
@aabatpurdue Can you tell us names of these files ?

siteshwar:
fwiw on my linux machine `401` files are compiled (without api tests).

aabatpurdue:
@siteshwar The files that I see are in [dblcomp.txt](https://github.com/att/ast/files/2376947/dblcomp.txt).  I ran `ninja -v ...` and extracted the `-c file-path`.  I then sorted it twice,  the second time with the `-u` option, and then `sdiff``ed the two results.  The attached file contains the differences.

Ummm.  Should have said that I see this on Cygwin and MSYS2.

Does your linux box OS supply a number of the required functions that results in fewer compiles??

krader1961:
Those modules are built twice because they are built with a different value for a preprocessor symbol. That's how the legacy build system did it and no one has bothered to figure out if that's actually necessary. This is the relevant section from *src/cmd/ksh93/meson.build*. I added the comment nine months ago because I was wondering the same thing:

```
# TODO: Figure out why there is a different definition of these symbols for
# ksh93 versus shcomp and if there is a good reason add a comment why they are
# defined differently.
ksh93_c_args = shared_c_args + [
    '-DERROR_CATALOG=NULL',
]

shcomp_c_args = shared_c_args + [
    '-DERROR_CATALOG="libshell"',
]
```

krader1961:
[...]

PR #867, opened 2018-09-13
Remove preprocessor symbol ERROR_CATALOG

krader1961:
This symbol no longer has any effect on either the ksh or shcomp
programs. Remove it.

Related #866


krader1961:
This doesn't actually remove the second compilation of the modules affected by this preprocessor symbol. But it makes doing so possible. Removing the redundant compilation will happen in a future change.

PR #875, opened 2018-09-16
Coverity leak2

krader1961:


src/lib/libast/regex/regcomp.c
```diff
@@ -2130,23 +2130,27 @@ static_fn Rex_t *regcomp_grp(Cenv_t *env, int parno) {
...
+                return NULL;
             }
             e->re.group.size = c;
             e->re.group.expr.binary.left = f;
-            if (!(e->re.group.expr.binary.right = regcomp_alt(env, parno, 1))) {
+            e->re.group.expr.binary.right = regcomp_alt(env, parno, 1);
+            if (!e->re.group.expr.binary.right) {
                 drop(env->disc, e);
-                return 0;
+                free(e);
```
siteshwar: Won't this leak if enclosing condition is `false` ?

krader1961: Yes, I was focused on the specific leak reported by Coverity but you are correct there are others. But it's not quite as simple as just freeing `e` regardless. That's because `regcomp_rep()` may return `e` which needs to be propagated or it may return NULL in which case `e` needs to be freed. I've gone ahead and fixed the other two obvious leaks of `e`.

PR #877, opened 2018-09-16
Fix file descriptor leak

### LINKED ISSUE #876: File descriptor leak reported by Coverity hides a deeper, more serious, problem
Coverity issue [CID#294191](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=51930236&defectInstanceId=9618998&mergedDefectId=294191) warns us about a file descriptor leak in `sh_iosave()`. But if you look closely you'll see that `fd` (the leaked var) is **not actually used**. This isssue exists in the ksh93v-branch before we started making changes. The ksh93u+ code does not contain the block which creates the leak that begins with these statements:

```
if (savefd < 0 && (sp = shp->sftable[origfd]) && (sfset(sp, 0, 0) & SF_STRING)) {
    fd = open("/dev/null", O_RDONLY | O_CLOEXEC);
```

It looks like it should have done a `savefd = fd;` if the `if (fd < 10)` test inside that block of code was false. So it's not surprising no one has ever reported a problem attributable to this issue as it would require an unlikely sequence of events.

krader1961:
The file descriptor leak is actually the least important aspect of this
fix. The most important aspect is that without this change the code uses
the wrong value associated with `savefd` under unusual conditions.

CID#294191
Fixes #876

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -1711,7 +1711,7 @@ static_fn ssize_t tee_write(Sfio_t *iop, const void *buff, size_t n, Sfdisc_t *u
 void sh_iosave(Shell_t *shp, int origfd, int oldtop, char *name) {
     int savefd;
     Sfio_t *sp;
-    int fd, flag = (oldtop & (IOSUBSHELL | IOPICKFD));
+    int flag = (oldtop & (IOSUBSHELL | IOPICKFD));
     int savestr = 0;
 
     oldtop &= ~(IOSUBSHELL | IOPICKFD);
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -1753,18 +1753,19 @@ void sh_iosave(Shell_t *shp, int origfd, int oldtop, char *name) {
...
+            if (fd == -1) {
+                errormsg(SH_DICT, ERROR_system(1), e_open, "/dev/null");
                 __builtin_unreachable();
             }
 
             if (fd < 10) {
                 savefd = sh_fcntl(fd, F_DUPFD_CLOEXEC, 10);
                 close(fd);
+            } else {
+                savefd = fd;
```
siteshwar: I am not sure how to verify this fix is correct.

krader1961: I don't know how to test this since it clearly requires two unusual conditions. But it is fairly obvious that the intent is to ensure `savefd` is greater than 9 when the subsequent code is executed. So if the preceding `open()` assigns a fd >= 10 there isn't any need for the `sh_fcntl` to "move" the fd.

PR #880, opened 2018-09-16
Add interactive tests for hist builtin

siteshwar:


src/cmd/ksh93/tests/hist.exp
```diff
@@ -0,0 +1,118 @@
+# Interactive tests for hist builtin
+set pid [spawn $ksh]
+expect_prompt
+
+# ==========
+# hist -l lists commands on standard output
+log_test_entry
+send "hist -l\r"
+
+expect -ex "hist -l\r\n8\tcd foo2\r\n9\tcd ..\r\n10\tmkdir foo3\r\n11\tcd foo3\r\n12\tcd ..\r\n13\techo hello world\r\n14\techo hello world1 > foo\r\n15\techo hello world2 >> foo\r\n16\techo hello world3 >> foo\r\n17\techo hello world1 > bar\r\n18\techo hello world2 >> bar\r\n19\techo hello world3 >> bar\r\n20\techo hello world1 > baz\r\n21\techo hello world2 >> baz\r\n22\techo hello world3 >> baz\r\n23\thist -l\r\n" {
```
krader1961: Personally I would replace a lot of the stuff in the middle with a simple `\r\n.*\r\n` but this is fine too.

krader1961:
LGTM

src/cmd/ksh93/tests/hist.exp
siteshwar: The reason I did not want to refactor code for issue #387 is because we did not have good tests for history related functions. I want the tests to be as thorough as possible.

PR #883, opened 2018-09-17
Resolve Coverity Scan false positive

### LINKED ISSUE #882: Coverity false positive about a not null terminated string
Coverity issues warning [CID#294196](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=51985454&defectInstanceId=9627662&mergedDefectId=294196) about the use of a potentially unterminated string. This is a false positive but it's understandable given how the code is structured. However, the offending function really should ensure the string is null terminated. Restructure the code to make it clear to lint tools the string is null terminated and make the code safer to use.

krader1961:
CID#294196
Fixes #882

### DIFF
--- a/src/lib/libast/path/pathprog.c
+++ b/src/lib/libast/path/pathprog.c
@@ -38,7 +38,6 @@
 #endif
 
 #include "ast.h"
-#include "ast_api.h"
 
 #if __CYGWIN__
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #887, opened 2018-09-18
Remove code that is a no-op

### LINKED ISSUE #886: Coverity resource leak warning points to a more serious isssue
Coverity Scan reported resource leak [CID#253698](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=52075654&defectInstanceId=9648767&mergedDefectId=253698&fileStart=751&fileEnd=1000) in *src/cmd/ksh93/sh/path.c*. The warning appears to be a false positive. But it caused me to take a close look at the code:

```
    path = path_relative(shp, opath);
    if (*path != '/' && path != opath) {
        // The following code because execv(foo,) and execv(./foo,) may not yield the same results.
        char *sp = (char *)malloc(strlen(path) + 3);
        sp[0] = '.';
        sp[1] = '/';
        strcpy(sp + 2, path);
        path = sp;
    }
    if (spawn /* && !sh_isoption(shp,SH_PFSH) */) {
        pid = _spawnveg(shp, opath, &argv[0], envp, spawn >> 1);
    } else {
        pid = path_pfexecve(shp, opath, &argv[0], envp, spawn);
    }
    if (xp) *xp = xval;
    if (*path == '.' && path != opath) {
        free(path);
        path = path_relative(shp, opath);
    }
```

Note that the first and last `if (...)` blocks used to be protected by `#ifdef SHELLMAGIC`. Since that was always defined to be one we removed the preprocessor test. The block of code above exists in the ksh93u+ branch.

**What is interesting is that the munged path isn't actually used.** So while I have no idea what problem that path munging was intended to solve (presumably a workaround for a broken platform) it is clearly a no-op now. And removing that code will eliminate the Coverity resource leak warning.

Also, it's not at all obvious the last `if (...)` can't call `free(path)` even if the first `if (...)` block wasn't executed. Looking at the implementation of `path_relative()` it looks like it should be impossible but I can't help but wonder what would happen if someone deliberately specified a path with an embedded `/./` segment after a `PWD` prefix. Regardless, the code in question is not needed.

krader1961:
Coverity Scan warned about a theoretical memory leak. It looks like that
warning is a false positive but it caused me to notice the offending
code is a no-op. So remove it.

CID#253698
Fixes #886

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -1057,24 +1057,12 @@ pid_t path_spawn(Shell_t *shp, const char *opath, char **argv, char **envp, Path
     sfsync(sfstderr);
     sh_sigcheck(shp);
     path = path_relative(shp, opath);
-    if (*path != '/' && path != opath) {
-        // The following code because execv(foo,) and execv(./foo,) may not yield the same results.
-        char *sp = (char *)malloc(strlen(path) + 3);
-        sp[0] = '.';
-        sp[1] = '/';
-        strcpy(sp + 2, path);
-        path = sp;
-    }
     if (spawn /* && !sh_isoption(shp,SH_PFSH) */) {
         pid = _spawnveg(shp, opath, &argv[0], envp, spawn >> 1);
     } else {
         pid = path_pfexecve(shp, opath, &argv[0], envp, spawn);
     }
     if (xp) *xp = xval;
-    if (*path == '.' && path != opath) {
-        free(path);
-        path = path_relative(shp, opath);
-    }
     if (pid > 0) return pid;
 retry:
     shp->path_err = errno;


kdudka:
How are you confirming that the code is a no-op?

I can see that `path` is passed to `exscript()` in the code below the mentioned block.

krader1961:
@kdudka If you look closely at the logic it is obvious that the dynamically allocated path is not used by the statement you reference. So either the `_spawnveg()` and `path_pfexecve()` calls between the two blocks of code I removed are supposed to use `path` rather than `opath` or that code is a no-op.

krader1961:
My bet is that long ago the munged path was intended to be used by the two function calls between the blocks of code I removed. Whether or not the code ever did so is now irrelevant.

krader1961:
Too, the fact that mere mortals can't agree on something as simple as this is a pretty strong signal that the code needs significant refactoring.

kdudka:
That is not the point.  If you push the proposed change, `exscript()` may run with different value of `path`, resulting in unwanted changes of behavior.  Calling such code a no-op is misleading and not at all helpful.

krader1961:
> That is not the point...

It is precisely the point. Look at the code before my change. The structure makes it quite clear that mutated `path` is not meant to be used by the statement you are worried about. The only question is if the two functions between the code I removed should use `path` rather than `opath`.

kdudka:
Given the fact that nobody complains about the current behavior suggests that ksh does what users (or rather existing ksh scripts) expect.  **Changing the behavior** to make the code look better (or compile with fewer warnings) is not what users of ksh expect.  On top of that, describing such change as `Remove code that is a no-op` is a really bad idea from maintenance point of view.

kdudka:
*Refactoring* denotes a process of restructuring existing code without changing its external behavior, which is definitely not what this PR is about.

If you think you are fixing a bug, you should support it by some examples, regression tests, etc.  Then we can evaluate how serious the bug is and what is the cost of the fix in terms of backward compatibility.  However, blindly changing the code could hardly help anybody.

kdudka:
Yes, @siteshwar has explained me why the value passed to `exscript()` as `path` would be the same with/without the proposed change.  So there should be no change in behavior caused by this patch really.  Sorry for not understanding it originally.

And, as @siteshwar says, it would be good to describe the actual Coverity report to be fixed in the commit message.  Not everybody has access to the ksh project at scan.coverity.com service.  Moreover, the link in https://github.com/att/ast/issues/886#issue-361122756 did not work for me despite I have the access.

PR #889, opened 2018-09-18
Remove Coverity dead code warning

### LINKED ISSUE #888: Coverity Scan warns about dead code which points to a deeper problem.
See Coverity Scan [CID#319591](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=52075629&defectInstanceId=9648636&mergedDefectId=319591) which warns about dead code in `nv_setarray()`. Which is a false positive AFAICT but does lead us to recognize other problems with the code. Such as this sequence of statements in that function (the last one being the cause of the Coverity warning):

```
if (!(fp = nv_isvtree(np))) value = nv_getval(np);
if (!fun) return NULL;
if (ap) return NULL;
```

Notice that neither `fp` nor `value` is used if `!fun`or `ap` is true. And `nv_istree()` has no side-effects. This bogosity was much harder to recognize in the original ksh93v- code because it was written like this:

```
        if(!(fp=nv_isvtree(np)))
                value = nv_getval(np);
        if(fun && !ap && (ap = (Namarr_t*)((*fun)(np, NIL(char*), NV_AINIT))))
```

krader1961:
CID#319591
Fixes #888

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -1057,24 +1057,12 @@ pid_t path_spawn(Shell_t *shp, const char *opath, char **argv, char **envp, Path
     sfsync(sfstderr);
     sh_sigcheck(shp);
     path = path_relative(shp, opath);
-    if (*path != '/' && path != opath) {
-        // The following code because execv(foo,) and execv(./foo,) may not yield the same results.
-        char *sp = (char *)malloc(strlen(path) + 3);
-        sp[0] = '.';
-        sp[1] = '/';
-        strcpy(sp + 2, path);
-        path = sp;
-    }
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/sh/path.c
```diff
@@ -1057,24 +1057,12 @@ pid_t path_spawn(Shell_t *shp, const char *opath, char **argv, char **envp, Path
     sfsync(sfstderr);
     sh_sigcheck(shp);
     path = path_relative(shp, opath);
-    if (*path != '/' && path != opath) {
```
siteshwar: This change has been already merged into `master`. This commit should not be in this pull request.

krader1961: Correct. I accidentally opened the PR before pushing to master the branch this was based on. It's harmless.

PR #891, opened 2018-09-19
Don't return a pointer to a stack var

### LINKED ISSUE #890: Coverity Scan: base64encode(): Pointer to local outside scope
Coverity Scan [CID#253637](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=52144380&defectInstanceId=9664463&mergedDefectId=253637) reports that `base64encode()` may return with the value assigned to parameter `tn` pointing to stack local var `tmp`.

There is only one caller that passes a pointer for parameter `tn`: `nv_getval()`. And that function throws away the assigned value. So rather than try to fix this bug just remove the `tn` parameter from the `base64encode()` interface.

P.S., Interestingly the unit test for `base64encode()` passes NULL for the `tn` parameter. So whomever wrote that test didn't think it was worth testing that returned value was correct.

krader1961:
Coverity Scan pointed out that it is possible for `base64encode()` to
assign the address of `tmp` to the `tn` argument. Since no users of this
function actually use the `tn` parameter just remove it.

This also fixes a related failure mode that clang static analysis
identified after removing `tn`:

src/lib/libast/string/base64.c:124:51: The left operand of '==' is a
garbage value

So predicate that test on `tp != tmp`.

This also enables running the API/base64.c unit test.

CID#253637
Fixes #890

### DIFF
--- a/src/lib/libast/string/base64.c
+++ b/src/lib/libast/string/base64.c
@@ -32,6 +32,8 @@
 #include <string.h>
 #include <sys/types.h>
 
+#include "ast.h"  // IWYU pragma: keep
+
 #define PAD '='
 
 #define B64_UC 3
@@ -49,7 +51,7 @@ static const char alp[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0
  * mime base64 encode
  */
 
-ssize_t base64encode(const void *fb, size_t fz, void **fn, void *tb, size_t tz, void **tn) {
+ssize_t base64encode(const void *fb, size_t fz, void **fn, void *tb, size_t tz) {
     unsigned char *fp;
     unsigned char *tp;
     unsigned char *fe;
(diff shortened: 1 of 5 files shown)


src/lib/libast/string/base64.c
```diff
@@ -119,13 +118,12 @@ ssize_t base64encode(const void *fb, size_t fz, void **fn, void *tb, size_t tz,
         *tp++ = (fz == 2) ? m[(b >> 6) & 077] : PAD;
         *tp++ = PAD;
     }
-    if (n)
+    if (n) {
         n += (tp - tmp) - 1;
-    else {
-        if (tp > (unsigned char *)tb && *(tp - 1) == '\n') tp--;
+    } else {
+        if (tp != tmp && tp > (unsigned char *)tb && *(tp - 1) == '\n') tp--;
```
siteshwar: Why are you comparing ` tp != tmp` ? I would appreciate a comment in source code here.

krader1961: I decided to turn it into an assert. Here's the comment I added:

```
// Static analysis warns that there is a theoretical path where the next `if` can be reached
// with tp == tmp. If that happens the `*(tp - 1)` would access the byte before the start of
// the tmp array.
```

It's due to this warning I put in the commit comment:

```
src/lib/libast/string/base64.c:124:51: The left operand of '==' is a garbage value
```

Basically this is yet another example of code that is poorly structured and plays fast and loose with the rules.

PR #893, opened 2018-09-19
Coverity Scan: resource leak

### LINKED ISSUE #892: Coverity Scan: sfpoll() leaks memory when a subsequent allocation fails
Coverity Scan [CID#253559](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=52144268&defectInstanceId=9664548&mergedDefectId=253559) warns that if the allocation of a `struct pollfd` array fails at line 107 of *src/lib/libast/sfio/sfpoll.c* the earlier allocation of an array of ints assigned to `status` will be leaked.

This is a poster child example for why I opened issue #853. What in the hell is the caller supposed to do if such a fundamental allocation fails thus causing `sfpoll()` to return an error? Pretty much anything it does at that point will soon result in another allocation failure and probably terminate the process. Only now at a distant remove from the earlier allocation failure. I'm not going to change the logic other than to eliminate the theoretical, but who the hell cares, leak. I'm just pointing out that this sort of anal approach to "always let the caller decide what to do on failure" is often pointless.

krader1961:
Returning when the malloc of a `struct pollfd` array fails leaks the
memory allocated earlier and assigned to `status`.

CID#253559
Fixes #892

### DIFF
--- a/src/lib/libast/sfio/sfpoll.c
+++ b/src/lib/libast/sfio/sfpoll.c
@@ -40,14 +40,15 @@
 int sfpoll(Sfio_t **fa, int n, int tm) {
     int r, c, m, np, eintr;
     Sfio_t *f;
-    int *status, *check;
+    int *check;
 
     if (n <= 0 || !fa) return -1;
 
-    if (!(status = (int *)malloc(2 * n * sizeof(int)))) return -1;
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #894, opened 2018-09-20
Remove symbol _shtest_c

krader1961:
While working on another change I noticed preprocessor symbol
`_shtest_c`. It is completely superfluous. It's sole purpose is to
ensure calls to `access()` don't get translated to `sh_access()` inside
that function, thus creating unwanted recursion. It turns out there are
no places which need

    #define access(a, b) sh_access(a, b)

All the places which need it simply call `sh_accss()` directly. So just
remove that macro as well as the puzzling `_shtest_c` symbol.

### DIFF
--- a/src/cmd/ksh93/include/shell.h
+++ b/src/cmd/ksh93/include/shell.h
@@ -389,12 +389,6 @@ extern Shell_t sh;
 #undef stat
 #define stat(a, b) sh_stat(a, b)
 #endif
-#ifndef _shtest_c
-#ifndef _SH_PRIVATE
-#undef access
-#define access(a, b) sh_access(a, b)
-#endif
-#endif  // !_shtest_c
 #ifndef _shio_h
 #undef chdir
 #define chdir(a) sh_chdir(a)
@@ -414,8 +408,6 @@ extern Shell_t sh;
 #endif
 #endif
 #ifndef _SH_PRIVATE
-#undef access
-#define access(a, b) sh_access(a, b)
 #undef close
 #define close(a) sh_close(a)
 #if SHELLAPI(20120720)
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/include/shell.h
```diff
@@ -414,8 +408,6 @@ extern Shell_t sh;
 #endif
 #endif
 #ifndef _SH_PRIVATE
-#undef access
-#define access(a, b) sh_access(a, b)
```
siteshwar: I suspect it might break https://github.com/att/ast/blob/8d69ea8fdc6c02d317afac76bd6b26f3915e86e9/src/lib/libcmd/mkdir.c#L148 if effective and real user id is different.

siteshwar: Also we should probably replace https://github.com/att/ast/blob/8d69ea8fdc6c02d317afac76bd6b26f3915e86e9/src/cmd/ksh93/sh/path.c#L1016 by manually calling `sh_access`.

krader1961: Note that neither of those use that macro. I verified this is safe by changing it to `#define access(a, b) WTFsh_access(a, b)`. If there were any places using that macro that change would have broken the build by introducing an unresolved symbol. Whether those two uses should be using `sh_access` and they are currently broken is not clear. The answer for the *path.c* instance is almost certainly yes since there are two places in that module which use `sh_access`.

krader1961: The `access()` call in *mkdir.c* is suspicious in as much as it looks to be pointless. In any case I don't think it needs to be changed to `sh_access()`. The one in *path.c*, on the other hand, does look like it should be using `sh_access()` so I'll change that one.

PR #897, opened 2018-09-20
Remove preprocessor symbol USE_REALLOC

### LINKED ISSUE #896: Remove preprocessor symbol USE_REALLOC
While fixing other problems with *src/lib/libast/misc/stk.c* I noticed preprocessor symbol `USE_REALLOC`. We don't define it and neither did the legacy master and beta branches. It is so old (dating back to at least ksh93u+) that we should just remove it.

TBD is whether to pretend it was defined or not. That's because constructs like this seem dodgy:

```
if (sp->stkflags & STK_SMALL)
#ifndef USE_REALLOC
    if (flags & STK_SMALL)
        bsize = roundof(bsize, STK_FSIZE / 16);
    else
#endif /* USE_REALLOC */
        bsize = roundof(bsize, STK_FSIZE);
```

Specifically the division by 16 if the `STK_SMALL` flag is set. Why 16? And the man page says this:

```
A \fIflags\fP argument of \f5STK_SMALL\fP indicates that unused
space on the stack should be freed whenever this stack ceases
to be the active stack.
```

Which doesn't seem to say anything about reducing the size of the allocation by a factor of 16. Too, we're no longer using the AST vmalloc subsystem. So this behavior may have been to accommodate some quirk of that subsystem.

krader1961:
Also simplify the code to ignore requests for small stacks and instead
just use a smaller, fixed, size to round the stack size. This leaves
the STAK_SMALL and STK_SMALL symbols just in case we decide that the
distinction really does matter.

Fixes #896

### DIFF
--- a/features/meson.build
+++ b/features/meson.build
@@ -56,6 +56,8 @@ feature_data.set10('_hdr_execinfo', cc.has_header('execinfo.h', args: feature_te
 feature_data.set10('_hdr_unistd', cc.has_header('unistd.h', args: feature_test_args))
 feature_data.set10('_hdr_stdlib', cc.has_header('stdlib.h', args: feature_test_args))
 feature_data.set10('_hdr_malloc', cc.has_header('malloc.h', args: feature_test_args))
+feature_data.set10('_hdr_filio', cc.has_header('filio.h', args: feature_test_args))
+feature_data.set10('_sys_filio', cc.has_header('sys/filio.h', args: feature_test_args))
 
 feature_data.set10('_lib_sigqueue',
     cc.has_function('sigqueue', prefix: '#include <signal.h>', args: feature_test_args))
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -154,6 +154,7 @@ static_fn bool iousepipe(Shell_t *shp) {
     if (!sh_iovalidfd(shp, subpipe[2])) abort();
     shp->fdstatus[subpipe[2]] = shp->fdstatus[1];
     close(fd);
+    // cppcheck-suppress deallocuse // it's okay to use fd in the fcntl() even though it is closed
     fcntl(subpipe[1], F_DUPFD, fd);
     shp->fdstatus[1] = shp->fdstatus[subpipe[1]] & ~IOCLEX;
     sh_close(subpipe[1]);
@@ -183,6 +184,7 @@ void sh_iounpipe(Shell_t *shp) {
         goto done;
     }
     close(fd);
+    // cppcheck-suppress deallocuse // it's okay to use fd in the fcntl() even though it is closed
     fcntl(subpipe[2], F_DUPFD, fd);
     shp->fdstatus[1] = shp->fdstatus[subpipe[2]];
     if (subdup) {
--- a/src/cmd/ksh93/tests/types.sh
+++ b/src/cmd/ksh93/tests/types.sh
@@ -711,7 +711,8 @@ typeset -T baz_t=(
 )
 baz_t z
 [[ ${z.z} == 789 ]] || log_error "z.z is ${z.z} should be 789"
-[[ ${ z.out } == "$out" ]] 2> /dev/null || log_error "z.out should be $out"
+[[ ${ z.out } == "$out" ]] 2> /dev/null || log_error "z.out wrong" "$out" "${ z.out }"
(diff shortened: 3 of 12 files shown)


krader1961:
Thanks, @etscrivner. The change obviously can't have any impact other than possibly the speed of the shell. The reason @siteshwar hasn't given a thumbs up is that we're waiting to see if Joerg Schilling, a noted proponent of ksh on Solaris who is concerned about what we're doing with the code, has to say about the change.

So I'm curious if you have any concerns at all about simplifying the code in a way that theoretically might cause heavy use of regexes to incur a performance penalty. I say "theoretically" because AFAICT (based on the unit tests) the effect, if any, is so close to zero that it doesn't matter.

krader1961:
Also, I recognize that the unit tests are not intended to evaluate the speed of the code. I was simply pointing out in my previous comment that this doesn't introduce an egregious (i.e., obvious) slow down. And, AFAICT from reviewing the code, it should not cause a slow down other than for an artificially engineered test case. Any real world ksh script that exercises this code should not be slowed by the slightly larger allocations by more than can otherwise be attributed to random variation in run time.

etscrivner:
Fair enough, AFAIK this produced no detectable slowdown on my modern hardware - though the case may be different on older systems.

My personal take on this is that simplifying ksh is probably necessary to make it at all maintainable. Any performance hits taken on regexps along the way would (hopefully) be minimal, and the simplifications made would open the door for potential larger optimizations to regexp and other subsystems in the future. In other words, I'm okay with things getting slightly (within reasonable limits) worse before they get better.

krader1961:
I've merged this with the change mentioned in my last comment to issue #896. Specifically, ensure `STK_FSIZE` is the system page size if it can be determined at build time else `4096`.

siteshwar:
Looks like this change broke one test on 32-bit Ubuntu:

```
194/210 types/shcomp                            FAIL     3.30 s
```

krader1961:
The `types` test (both variants) is by far the flakiest of all the unit tests. It easily accounts for at least half the test failures. Both on my systems and Travis. This change actually increases the initial size of a couple of allocations and doesn't reduce the size of any allocation on 32-bit systems. So if anything it should decrease the likelihood of many types of malloc errors (e.g., allocating a buffer that is too small).

I'm not seeing this failure on my Ubuntu 32-bit system. Note that my original version where I reduced the rounding to 2048 did cause failures on that system. So there is something deeply wrong with the relevant code. I'll investigate this but we may have to just disable that test for the moment in that environment.

krader1961:
Something is wrong with our Travis config or Travis itself. As part of debugging this issue I added the following just before the failing test:

```
log_warning "$(uname -s) $(uname --processor)"
log_warning "$(uname -a)"
cat /etc/lsb-release >&2
```

That reports that the `i386/ubuntu` environment is actually x86_64 rather than the expected `i686`.

krader1961:
It looks to me like `docker pull ${DISTRO_TYPE}` is pulling from public repositories not under the control of Travis. And that invalid distros are not hard failures but rather resolve to a "best match" or some such. I have a rough idea what Docker is about but haven't dealt with Docker first hand before now. So anything I'm saying regarding the unexpected processor type vis-a-vis our Travis 32-bit environment could be complete nonsense.

krader1961:
Changing `i386/ubuntu` to `i686/ubuntu` results in a different failure during the installation of the OS:

```
Meson encountered an error:
Could not detect Ninja v1.5 or newer
```

That appears to be due to the Docker `i686/ubuntu` platform being the rather old `Ubuntu 14.04.5 LTS` release. But since we want to support platforms that are not bleeding edge this needs to be resolved.

siteshwar:
> It looks to me like docker pull ${DISTRO_TYPE} is pulling from public repositories not under the control of Travis.

We are pulling official 32-bit images of ubuntu from [here](https://hub.docker.com/r/i386/ubuntu/). These are provided by Docker.

krader1961:
> We are pulling official 32-bit images of ubuntu from here. These are provided by Docker.

@siteshwar Thanks for the explanation. Would you mind amending our *.travis.yml* config file to include a comment to that effect? Also commenting on why we are using Docker image names that are not versioned and including links to the Docker documentation for each image we depend on? I know this is a bit pedantic to someone familiar with Docker. But one problem with the project we inherited is too much was not documented because the authors assumed any reader of the code would share their background knowledge that informed the creation of the code.

PR #899, opened 2018-09-20
Lint: SFOPEN() definition is confusing

### LINKED ISSUE #898: Macro `SFOPEN` should be redefined
Oclint is issuing a lot of warnings like this one:

```
src/lib/libast/sfio/sfstack.c:104:5: constant conditional operator [basic|P2]
```

Which is due to this statement:

```
SFOPEN(f1, 0);
```

The reason for the warning is the second argument is a constant and the macro tests if it is zero or non-zero:

```
#define SFOPEN(f, l) (void)((l) ? 0 : ((f)->mode &= ~(SF_LOCK | SF_RC | SF_RV), _SFOPEN(f), 0))
```

The majority of `SFOPEN()` uses pass a constant zero for the second, `l`, parameter. The few that don't should do an explicit `if (l) SFOPEN(f);`. Which would also obviate the need for the otherwise pointless cast to `void`.

krader1961:
The SFOPEN() macro causes lots of lint warnings like this one:

src/lib/libast/sfio/sfstack.c:104:5: constant conditional operator [basic|P2]

That's because the majority of uses pass a constant zero for the second,
`l`, parameter which turns the macro into a no-op. It also obscures the
purpose of the second parameter. Redefine the macro to take just a
single parameter and redefine the few places which don't want it to
always open a SFIO stream by turning

    SFOPEN(f, local)

into

    if (!local) SFOPEN(f);

This both eliminates the warnings and clarifies the behavior.

Fixes #898

### DIFF
--- a/src/lib/libast/sfio/_sfputd.c
+++ b/src/lib/libast/sfio/_sfputd.c
@@ -65,7 +65,7 @@ int _sfputd(Sfio_t *f, Sfdouble_t v) {
     }
 
     /* write out the signs and the exp */
-    SFOPEN(f, 0);
+    SFOPEN(f);
     if (sfputc(f, n) < 0 || (w = sfputu(f, w)) < 0) SFMTXRETURN(f, -1);
     SFLOCK(f, 0);
     w += 1;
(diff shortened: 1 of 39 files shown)


siteshwar:
LGTM

PR #900, opened 2018-09-20
Don't retry close() on EINTR

### LINKED ISSUE #885: Stop retrying `close()` on EINTR
While working on a Coverity Scan resource leak I noticed this line in *xec.c*:

```
while (close(fd) < 0 && errno == EINTR) errno = err;
```

That practice is pointless. See these links for why:
 
https://lwn.net/Articles/576478/
http://austingroupbugs.net/view.php?id=529
https://sourceware.org/bugzilla/show_bug.cgi?id=14627

krader1961:
Per issue #885 there is no point in checking if `close()` results in
EINTR and retrying. So simplify the code.

Resolves #885

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -466,9 +466,8 @@ int sh_close(int fd) {
     }
 
     if (!(sp = shp->sftable[fd]) || (sffileno(sp) != fd) || (sfclose(sp) < 0)) {
-        int err = errno;
         if (fdnotify) (*fdnotify)(fd, SH_FDCLOSE);
-        while ((r = close(fd)) < 0 && errno == EINTR) errno = err;
+        close(fd);
     }
 
     if (fd > 2) shp->sftable[fd] = 0;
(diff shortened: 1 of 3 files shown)


krader1961:
There may be other places which retry `close()` on `EINTR` but that seems unlikely given the style of the code. And such code is unlikely to cause problems so it doesn't hurt to leave it in place if it wasn't caught by my simple `grep 'close.*EINTR'` that only detects when both occur on the same line.

siteshwar:
This should be safe on Linux. However reading the discussion on opengroup:

```
If close() returns -1 with errno setto [EINTR], it is unspecified whether fildes can subsequently bepassed to any function except close( ) or posix_close( ) without error.
--
```

So not reissuing `close()` will cause undefined behavior. @krader1961 Am I missing something here ? Is removing this code guaranteed to be safe on other OSs ?

krader1961:
What that comment is saying is that if the close fails due to EINTR (something that is impossible in practice) you cannot use the fd in any call other than another close. In other words, you can't continue to use the fd just because the close failed with EINTR precisely because a EINTR failure has undefined semantics. Well, duh! The person who wrote that was just being pedantic. Something I'm prone to do as well. :smile:

PR #903, opened 2018-09-21
Coverity: mem leak

### LINKED ISSUE #902: Coverity: mem leaks in dllscan.c
Coverity Scan [CID#288384](https://scan4.coverity.com/reports.htm#v37777/p13914/fileInstanceId=52274581&defectInstanceId=9679646&mergedDefectId=288384) shows a memory leak in `dllsopen()`. In addition to the leak it uses `calloc()` and `memcpy()` to duplicate a string when a simple `strdup()` would have done the job.

krader1961:
This should eliminate a memory leak identified by Coverity. The changes
of `calloc()` to `malloc()` are because the memory doesn't need to be
zeroed. That's a vestage from when the code used `newof()` which we
replaced with `calloc()` to retain the newof semantics. Surprise! Some
of those `newof()` uses didn't need to clear the memory.

CID#288384
Fixes #902

### DIFF
--- a/src/lib/libdll/dllscan.c
+++ b/src/lib/libdll/dllscan.c
@@ -25,6 +25,7 @@
 
 #include <ctype.h>
 #include <errno.h>
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

Issue #881, opened 2018-09-16
readonly builtin not working correctly in functions


perette:
**Description of problem:**
When the builtin `readonly` is used inside functions with an assignment to a freshly-created variable, the variable is created readonly but is given a null value.

**Ksh version:**
Version M 1993-12-28 s+ (OS/X Snow Leopard), Version A 2017.0.0-devel-1694-g4e123c2-dirty (recent) (Linux Mint)

**How reproducible:**
Reliably.

**Steps to reproduce:**
Run this script:
```
#!/bin/ksh
# These are outside function, so work...
readonly x=fish
readonly a=another
echo "[outside function] x is $x"



function t {
	echo "[in t] x is $x"

	# Inside function, readonly with value assignment makes read-only with null value.
	readonly y="hello, ${x}!"
	readonly z="mangle"

	# However, creating the variable then readonlying it works
	typeset w="hello, ${x}!"
	readonly w

	# And if it's created, then readonlyed with a value, it also works
	typeset v="hello, ${x}!"
	readonly v="a different value"

	# This errors out because it sees global 'a', which is
	# already readonly and can't be reassigned.
	# readonly a=inside

	# Though this works, creating a local constant a
	typeset a=foobar
	readonly a

	# Dump values
	echo "In function:"
	readonly -p
}

t


echo "After function:"
readonly -p
```

**Actual results:**
```
[outside function] x is fish
[in t] x is fish
In function:
readonly a=foobar
readonly v='a different value'
readonly w='hello, fish!'
readonly x=fish
readonly y
readonly z
After function:
readonly a=another
readonly x=fish

```

**Expected results:**
```
[outside function] x is fish
[in t] x is fish
In function:
readonly a=foobar
readonly v='a different value'
readonly w='hello, fish!'
readonly x=fish
readonly y="hello, fish!"
readonly z=mangle
After function:
readonly a=another
readonly x=fish
```

**Additional info:**

krader1961:
There has been extensive discussion about the expected behavior of the `readonly` builtin in PR #906. The question is whether it should be equivalent to `typeset -r` or if it should always put the var in the global scope. At the present time it does the latter. Which is consistent with how `export` behaves. Note that `export` and `readonly` are implemented by the same function so it is reasonable to conclude the author intended that behavior (the bug described by this issue notwithstanding).  But note that the ksh man page says this:

> The `export` command is the same as `typeset -x` except that if you use export within a function, no local variable is created.

[...]

PR #906, opened 2018-09-21
Set correct scope for readonly variables

siteshwar:


perette:
I have not managed to grok the name-value code, so all I can really say is, these changes fix the behavior.

krader1961:
This also seems to leave a related bug. Shouldn't `readonly x=y` be equivalent to `typeset -r x=y`? Which means that when run inside a function it shouldn't mutate the value in outer scopes? Run the following with and without the proposed fix:

```
abc=ABC
tro=TRO
foo=FOO
function test_func {
    readonly foo="bar"
    typeset -r tro=temp
    typeset abc=def
    echo inside foo $foo
    echo inside tro $tro
    echo inside abc $abc
}
test_func
echo outside foo $foo
echo outside tro $tro
echo outside abc $abc
```

Notice how `$foo` outside the function gets mutated but `$tro` doesn't.

perette:
@krader Good finds.  I agree this behavior is inconsistent, and probably a different bug.

Your test script exposes interesting behavior:
```
function foo {
   readonly x=fish
    echo inside $x
}
echo outside $x
x=whale
```

So it looks like the original implementation *creates a local variable*, but *put the contents into a global of the same name*.  That global lacks the readonly flag, which is assigned to the local.

So the question is, what *should* readonly do when given an assignment?  Create a local, equivalent to `typeset -r`, or create a global variable?  The local feels right, but I can see an argument that if you're really making a constant, it might as well go in the global scope.  `bash` does the global scope, so that weighs heavily in favor of the global.

perette:
If it's supposed to be equivalent to `typeset -r` (and that does seem like the natural thing), why isn't it just implemented with an `alias readonly=typeset -r`?  Given `float`, `integer`, and several other types are really just aliases, I'd have expected that approach unless `readonly` was meant to do something different.

If we do want to refactor it to be `typeset -r`, is it better to simplify by dropping the code and adding a predefined alias?

If we want to go with the bash-compatible behavior (which I agree with Krader may be a non-backward compatible fix), the man pages should get updated to be explicit about this deviant behavior.

krader1961:
Is `readonly` supposed to mimic `export`? The latter is not equivalent to `typeset -x` since it does not create a locally scoped exported var.  But note that `export` is explicitly documented not to be equivalent:

> The export command is the same as typeset -x  except  that  if you  use  export  within  a  function, no local variable is created.

No such caveat appears for the `readonly` documentation. Oversight in the documentation? Or deliberate in which case the implementation is wrong.

krader1961:
The fact `readonly` isn't implemented as an alias for `typeset -r` sort of suggests the documentation of `readonly` should have the same caveat as `export`. This is reinforced by the fact both commands are implemented by the same function.

perette:
I'm leaning toward readonly mimicking export.  Imagine the use case of a function load_program_configuration that reads some config (say, paths and stuff), and wants to stick it in global readonly variables for the rest of the program to use.  But there's no way to do that inside the function if `readonly` === `typeset -r`, since it would instead create locals, which would disappear on return.  You'd have to read the variables inside function, then readonly them in the main--but that breaks the modularity of having the function.

krader1961:
@perette Just to be clear.... You seem to be arguing for the documentation of `readonly` to include the same caveat as `export`. That is okay as far as I am concerned. As I mentioned earlier the fact both `readonly` and `export` are implemented by the same function (something I consider a bogosity, they should be separate functions) strongly suggests that `readonly` is meant to create a global var rather than a function scoped var like `typeset -r`. However, it is still inconsistent and broken in other ways.

What happens if you do `export x=y` inside and outside a function? Now consider what happens if you do `readonly x=y` inside and outside a function (i.e., marking it readonly before the function is run). What happens if you run the function more than once? This is pretty clearly something the original ksh team either didn't think deeply about or some fundamental implementation errors were made.

krader1961:
@siteshwar I'm inclined to move this discussion to issue #881 and give it a few weeks, better would be a couple of months, to see what others think should be the correct behavior. And since it is currently broken in a manner that makes it unusable (in ksh93u+ as well) we presumably don't have to worry about backward compatibility.

krader1961:
Closing this since there is a debate regarding the desired behavior in issue #881 with the current consensus being that the behavior implemented by this PR may be correct but is not optimal.

PR #908, opened 2018-09-22
Extend tests for `print` builtin

siteshwar:
Split `print` builtin tests in a separate file and add new tests.


krader1961:
LGTM. Thank you very much for prefacing each distinct test with `# ======` and a comment describing the test. The legacy tests are far harder to read and understand than necessary because whomever wrote them couldn't be bothered to add such markers and explanatory text.

PR #913, opened 2018-09-25
Lint: nullPointerRedundantCheck

krader1961:
From cppcheck:

[src/lib/libast/path/pathtemp.c:217] warning (nullPointerRedundantCheck):
 Either the condition '!d' is redundant or there is possible null
 pointer dereference: d.

The issue is that the person who wrote much of this code seemed to delight
in writing clever, compact, code. So refactor it to eliminate the lint
and clarify the behavior.

### DIFF
--- a/src/lib/libast/path/pathtemp.c
+++ b/src/lib/libast/path/pathtemp.c
@@ -113,6 +113,18 @@ static struct Tmp_s {
     char *tmppath;
 } tmp = {.mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH};
 
+static_fn char *get_tmp_dir() {
+    char *d = astconf("TMP", NULL, NULL);
+    if (*d && eaccess(d, W_OK | X_OK) == 0) return d;
+
+    d = TMP1;
+    if (eaccess(d, W_OK | X_OK) == 0) return d;
+    d = TMP2;
+    if (eaccess(d, W_OK | X_OK) == 0) return d;
+
+    return NULL;
+}
+
 char *pathtemp(char *buf, size_t len, const char *dir, const char *pfx, int *fdp) {
     char *d;
     char *b;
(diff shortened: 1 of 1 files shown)


src/lib/libast/path/pathtemp.c
```diff
@@ -113,6 +113,18 @@ static struct Tmp_s {
     char *tmppath;
 } tmp = {.mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH};
 
+static_fn char *get_tmp_dir() {
+    char *d = astconf("TMP", NULL, NULL);
+    if (*d && eaccess(d, W_OK | X_OK) == 0) return d;
+
+    d = TMP1;
```
siteshwar: This assignment looks unnecessary.

src/lib/libast/path/pathtemp.c
```diff
@@ -113,6 +113,18 @@ static struct Tmp_s {
...
+    if (eaccess(d, W_OK | X_OK) == 0) return d;
```
siteshwar: if (eaccess(TMP1, W_OK | X_OK) == 0) return TMP1;

src/lib/libast/path/pathtemp.c
```diff
@@ -113,6 +113,18 @@ static struct Tmp_s {
     char *tmppath;
 } tmp = {.mode = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH};
 
+static_fn char *get_tmp_dir() {
```
siteshwar: This name might clash with other names in debug builds in future. May be rename it to `get_pathtmp_dir`.

src/lib/libast/path/pathtemp.c
krader1961: Haha, yes. It was late and just unpacking the mess that was the original `if(...)` used up my remaining energy.

PR #916, opened 2018-09-25
Add tests for `set` builtin

siteshwar:


krader1961:
The test in *set.sh* at line 64 fails under shcomp on all my platforms. Which is probably a bug in the shcomp code. So you'll need to special-case that test to not run it under shcomp (don't forget to add a TODO to re-enable it). Other than that LGTM.

siteshwar:
@krader1961 `set -x` test fails on macOS. Is it something broken in tests or on macOS ?

siteshwar:
It seems an empty space is not printed after `PS4` in macOS.

PR #917, opened 2018-09-26
Lint: Value stored to 'size' is never read

### LINKED ISSUE #915: Unused `size` var in `pathcanon()`
I don't normally open an issue for lint errors reported by `cppcheck` or `oclint` (which includes warnings from clang static analysis). But this one by the compiler via oclint is interesting:

```
src/lib/libast/path/pathcanon.c:69:16: Value stored to 'size' is never read
```

It isn't just that the assignment is pointless. It's that the `size` parameter is unused. That parameter was used back when this code emulated/supported 3DFS. Every single invocation passes a value equal to `strlen(path) + 1` or a larger value. This parameter makes sense since the definition of the function is that it modifies the `path` buffer. But `man src/lib/libast/man/path.3` makes no mention of what happens if the returned value would be larger than the size of `path`.

The `size` values larger than `strlen(path) + 1` are what concerns me since there are statements like `strcpy(path, s);` in that function which don't obviously respect the original size of the buffer passed into the function. So did I introduce a bug in my changes to remove 3DFS support or is the API FUBAR? 

The answer appears to be "FUBAR" since the `pathcanon()` in the ksh93u+ legacy master branch also completely ignores the argument AFAICT.  In the ksh93v- beta branch it does appear to be used but only due to changes introduced by that version we deliberately decided to drop.

Note that I instrumented the code and it appears, while running unit tests on all the platforms I test such changes, that `pathcanon()` is never called with `size == 0` or that the `path` string (not buffer as defined by `size`) is ever overwritten by a longer path.

Also, Please explain to me why this

```
if (!*(path + 1)) return path + 1;
```

is clearer than

```
if(!path[1]) return "";
```

krader1961:
Fixes #915

### DIFF
--- a/src/lib/libast/path/pathcanon.c
+++ b/src/lib/libast/path/pathcanon.c
@@ -49,6 +49,7 @@
 #include <sys/stat.h>
 
 #include "ast.h"
+#include "ast_assert.h"
 
 #ifndef ELOOP
 #define ELOOP EINVAL
@@ -73,7 +74,7 @@ char *pathcanon(char *path, size_t size, int flags) {
                 path++;
             } while (*path == '/' && *(path + 1) == '/');
         }
-        if (!*(path + 1)) return path + 1;
+        if (!path[1]) return path + 1;
     }
     p = r = s = t = path;
     for (;;) {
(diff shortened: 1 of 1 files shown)


krader1961:
I considered replacing the `strcpy()` with `strlcpy()` rather than add `assert()`. But that would have resulted in potentially returning garbage to the caller. Better to just die if we're about to either corrupt memory or truncate a pathname.

PR #923, opened 2018-10-01
Remove irrelevant symbol `ERROR_VERSION`

### LINKED ISSUE #922: Remove symbol `ERROR_VERSION`
Preprocessor symbol `ERROR_VERSION` is another one that caught my eye and appears to be unnecessary. It is defined as the constant `20070319L`, both now and in the ast.beta and ast.master branches. So all of the `#if ERROR_VERSION >= 20000101L` like tests can be eliminated and presumed true or false according to the specific test.

Note that the specific values being tested include `20000101L`, `20000317L`, and `20030214L`. The current value of that symbol is `20070319L`. So obviously there was a 2+ year period when the code affected by that symbol was in flux. There is no obvious reason why we need to retain this symbol 15 years later.

krader1961:
Symbol `ERROR_VERSION` is a constant. The places it is tested are
therefore known to be true or false at compile time and those tests can
be eliminated.

Resolves #922

### DIFF
--- a/src/cmd/ksh93/include/fault.h
+++ b/src/cmd/ksh93/include/fault.h
@@ -95,11 +95,7 @@ struct checkpt {
     int mode;
     int vexi;
     struct openlist *olist;
-#if (ERROR_VERSION >= 20030214L)
     Error_context_t err;
-#else
-    struct errorcontext err;
-#endif
 };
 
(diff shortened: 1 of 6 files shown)


siteshwar:
LGTM

PR #925, opened 2018-10-01
Remove unnecessary `#ifdef ERROR_NOTIFY` test

### LINKED ISSUE #924: `#ifdef ERROR_NOTIFY` is pointless
Preprocessor symbol `ERROR_NOTIFY` is always defined. So the one place which does `#ifdef ERROR_NOTIFY` in *src/cmd/ksh93/sh/fault.c* can be simplified. Presumably that test is leftover from a time when the symbol was experimental. There are several places which use the symbol unconditionally. So it isn't possible to build ksh without it being defined.

krader1961:
Resolves #924


siteshwar:
LGTM

Issue #854, opened 2018-09-11
Eliminate pathtemp() or replace its custom RNG with standard libc RNG functionality.


krader1961:
The function `pathtemp()` in *src/lib/libast/path/pathtemp.c* has it's own custom RNG to "generate a pseudo-random name". This is both a bad idea and unnecessary. We should either replace the custom RNG with standard libc functionality or replace the handful of uses of `pathtemp()` with something like `mktemp()`.

krader1961:
From the *src/lib/libast/man/path.3* man page:

```
       pathtemp  generates  in  path  a   temporary   file   path   name   of   the   form
       dir/pfx<pid>.<suf>  where the length of pfx, if !=0, is limited to 5, the length of
       <pid> (the base 64 representation of the current process id) is limited to  3,  and
       <suf>  (an  internally generated suffix that avoid file conflicts) is limited to 3.
       The generated path name conforms to the classic UNIX 14 char and the DOS 8.3  limi-
       tations.  Both dir and pfx may be 0.  access(2) is used to avoid file conflicts but
       the generated path name is not created, so you could lose in a race.
```

We can't possibly care about those limitations in the year 2018. And to the extent they are relevant the libc equivalent functions will deal with them.

krader1961:
There are only seven uses of `pathtemp()` now. And there are only three patterns. None of which require the features provided by the AST `pathtemp()`.

krader1961:
I've merged the two commits that resolve this issue by replacing the very complex and problematic `pathtemp()`  with a much simpler `ast_temp_file()`.

PR #926, opened 2018-10-03
Replace `pathtemp()` with a simpler function

krader1961:
This leaves the existing implementation just to keep the focus on the
important parts of resolving issue #854. I'll follow this with a cleanup
change that removes the now unused code.

Related #854

### DIFF
--- a/src/cmd/ksh93/bltins/hist.c
+++ b/src/cmd/ksh93/bltins/hist.c
@@ -21,6 +21,7 @@
 
 #include <ctype.h>
 #include <fcntl.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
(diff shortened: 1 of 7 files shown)


krader1961:
I've merged this with a second commit to remove the now unused code.

siteshwar:
I get lot of these warnings after this change:

```
upstream/ast/build/../src/lib/libast/path/pathtemp.c:103: warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
```

krader1961:
Yeah, I noticed that after I ran `bin/lint` after pushing this PR. In the cleanup commit that followed this one I added some suppressions. This is one of those, somewhat rare, instances where there really isn't a good alternative to using the problematic function. It may be that the one place that uses the temp file name to create a named fifo rather than a file (hence the need for `mktemp()`) should be changed to use a different mechanism for generating that name. But I wanted to keep this change as low risk as possible so decided using `mktemp()` for that one caller and dealing with the lint warnings was safer.

PR #930, opened 2018-10-05
Replace sh_lexstates with sh_lexrstates

### LINKED ISSUE #929: Replace `sh_lexstates` with `sh_lexrstates`
While resolving issue #851 contributor @perette made the following observation:

> Additionally, if sh_lexstates is truly a variable, then the else clause that resets it to default got whacked. However, I think now that the AST crap is gone, this is now a constant--in which case, this should get stripped out and sh_lexrstates should be used instead (or better, sh_lexrstates should become the new sh_lexstates to get rid of the unexplained 'r').

The `r` in `sh_lexrstates` is apparently meant to mean read-only. Having said that it looks to me like @perette is correct and `sh_lexstates` is just a copy of `sh_lexrstates` now and is never modified. There is one place in *src/cmd/ksh93/sh/macro.c* where it superficially looks like it might modify `sh_lexstates` indirectly. But that is just because the people who wrote the code assigned multiple meanings to variable `state`. Specifically, assigning both `sh_lexstates` and `mp->shp->ifstable` to it in `mac_copy()`. Even though the two state tables have literally nothing in common. A bad practice we see repeatedly in this project. The former alias is not modified the latter alias is modified.

AFAICT the only reason `sh_lexstates` existed as a separate array from `sh_lexrstates` was to support non-ASCII (i.e., EBCDIC) systems where a few entries in the state table are different due to differences in the codepoint assigned to symbols. The implementation clearly treated systems using EBCDIC as an inferior class because it could have used the same logic to construct the ASCII lexer state table rather than a hand-crafted set of arrays. Since we have dropped support for non-ASCII systems we can fold the two vars.

krader1961:
After dropping support for non-ASCII character encodings (e.g., EBCDIC)
we no longer need to copy `sh_lexrstates` to `sh_lexstates`. Just
renamed the former to the latter.

This also cleans up a couple of related bogosities such as the magic
`sh_lexstates[2]` reference in src/cmd/ksh93/edit/completion.c with
`ST_NORM` in place of `2`.

Resolves #929

### DIFF
--- a/src/cmd/ksh93/data/lexstates.c
+++ b/src/cmd/ksh93/data/lexstates.c
@@ -334,9 +334,9 @@ static const char sh_lexstate9[256] = {
     S_NOP,   S_NOP, S_NOP,   S_BRACE, S_PAT, S_ENDCH, S_NOP, S_NOP,  // 120 - 127
 };
 
-const char *sh_lexrstates[ST_NONE] = {sh_lexstate0, sh_lexstate1, sh_lexstate2, sh_lexstate3,
-                                      sh_lexstate4, sh_lexstate5, sh_lexstate6, sh_lexstate7,
-                                      sh_lexstate8, sh_lexstate9, sh_lexstate5};
+const char *sh_lexstates[ST_NONE] = {sh_lexstate0, sh_lexstate1, sh_lexstate2, sh_lexstate3,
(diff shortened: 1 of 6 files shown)


etscrivner:
LGTM. tested locally. Significantly simplifies and cleans up several things at once.

Issue #853, opened 2018-09-10
Why do so many functions return NULL if a memory allocation fails?


krader1961:
Please ponder why this function returns NULL if the memory allocation fails (from *src/cmd/ksh93/sh/nvdisc.c*):

```
static_fn void *newnode(const char *name) {
    size_t s;
    Namval_t *np = newof(0, Namval_t, 1, s = strlen(name) + 1);

    if (np) {
        np->nvname = (char *)np + sizeof(Namval_t);
        memcpy(np->nvname, name, s);
    }
    return (void *)np;
}
```

Also ponder the first assignment. Hint: it contains another assignment embedded within the ostensible function call but which is actually a macro. And because it is a macro the expression might be executed more than once. But that is a different rant than the one I want to focus on in this issue.

The question is whether these explicit returns of a null pointer on a memory allocation failure make any sense. My opinion is **hell no** in general. There may be specific instances where this makes sense. But in this specific instance the two call sites require a non-NULL return value. Thus the explicit test for a non-NULL value is pointless and only delays the failure. And AFAICT this is true for 99% of the equivalent situations in this project.

krader1961:
FWIW, There are many bogosities in this project. But two really stand out:

1) Wrapping every `return` expression in parentheses. Which leads to such bogosities as `return (0);`.

1) The explicit `if (!(ptr = (malloc(1)) return 0;`, or equivalent, statements. In every single instance I've examined that just delays the failure. I have yet to see an instance where ksh would do something useful if the malloc/calloc/realloc failed and the called function short circuited as a result.

krader1961:
The code is also not consistent about dealing with dynamic allocations that fail. For example, [this block](https://github.com/att/ast/blob/master/src/cmd/ksh93/sh/io.c#L536-L545) silently ignores the second `strdup()` failure rather than return `-1`. And note that it also leaks the previous `strdup()` buffer. Lastly, the first `strdup()` isn't even necessary unless the path contains a `/` character.

krader1961:
Another great example for why this pattern needs to be changed. See `nv_clone_disc()` in *src/cmd/ksh93/sh/nvdisc.c*. It returns NULL if the allocation fails but every caller blindly assumes the pointer is not NULL. I have yet to find a single place in the ksh code that does anything useful if a function it calls fails due to a dynamic allocation failure. That is, any place that does not immediately report a failure via `errormsg()` or similar mechanism when an allocation fails.

krader1961:
[...]

PR #933, opened 2018-10-07
Remove lint from module io.c

krader1961:
Clang static analysis reports:

src/cmd/ksh93/sh/io.c:541:14: Potential leak of memory pointed to by 's'

This eliminates that lint warning and also restructures the code to clarify
the behavior. As always, this type of change is not meant to address
every shortcoming of the code being modified. It's only meant to be an
incremental step to improving the quality by eliminating a lint warning.

Related #853

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -532,21 +532,26 @@ static int inetopen(const char *path, int flags, Inetintr_f onintr, void *handle
             return -1;
         }
     }
+
     if (flags == O_NONBLOCK) return 1;
-    if (!(s = strdup(path))) return -1;
-    t = strchr(s, '/');
-    if (t) {
-        *t++ = 0;
-        if (!strcmp(s, "local")) s = strdup("localhost");
-        fd = getaddrinfo(s, t, &hint, &addr);
-    } else {
-        fd = -1;
+
+    t = strchr(path, '/');
+    if (!t) return -1;
+
+    s = strdup(path);
+    s[t - path] = 0;
+    if (!strcmp(s, "local")) {
+        free(s);
+        s = strdup("localhost");
     }
+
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/io.c
```diff
@@ -532,21 +532,26 @@ static int inetopen(const char *path, int flags, Inetintr_f onintr, void *handle
...
-        *t++ = 0;
-        if (!strcmp(s, "local")) s = strdup("localhost");
-        fd = getaddrinfo(s, t, &hint, &addr);
-    } else {
-        fd = -1;
+
+    t = strchr(path, '/');
+    if (!t) return -1;
+
+    s = strdup(path);
```
siteshwar: Shall we check if `strdup()` fails ?

src/cmd/ksh93/sh/io.c
```diff
@@ -532,21 +532,26 @@ static int inetopen(const char *path, int flags, Inetintr_f onintr, void *handle
...
-    if (!(s = strdup(path))) return -1;
-    t = strchr(s, '/');
-    if (t) {
-        *t++ = 0;
-        if (!strcmp(s, "local")) s = strdup("localhost");
-        fd = getaddrinfo(s, t, &hint, &addr);
-    } else {
-        fd = -1;
+
+    t = strchr(path, '/');
```
siteshwar: Rename `t` to `slash` ?

src/cmd/ksh93/sh/io.c
krader1961: Not really much point in an explicit `assert(s)`. If the following assignment doesn't fail (say because we're running on a system which doesn't make page zero invalid) the `strcmp()` will definitely fail.

src/cmd/ksh93/sh/io.c
krader1961: Good idea. Done.

PR #934, opened 2018-10-07
Remove LIBICONV_PLUG

### LINKED ISSUE #932: Undefined symbols _iconv_close
**Description of problem:**

Hi, I'm a developer with MacPorts, trying to fix the problem that we haven't been able to build ksh93 for years. I see you've switched to the meson build system, but when I try to build the latest code on macOS High Sierra, I get failures due to undefined symbol `_iconv_close`. I know that libiconv builds in different ways on different systems or in different configurations, and sometimes, as is the case in MacPorts, the symbol is actually called `_libiconv_close`. I don't know how projects are supposed to structure their code to accommodate both possibilities.

**Ksh version:**

Latest code as of 2018-10-05, 1faa299aa9bf68f31f0ff9283b683ced5daf0e81

**How reproducible:**

Always

**Steps to reproduce:**
1. On macOS, with MacPorts installed, check out the code, run `meson` and `ninja`.

**Actual results:**

```
[338/493] Linking target src/cmd/tests/aso/tlock.c.
FAILED: src/cmd/tests/aso/tlock.c 
ccache /usr/bin/clang  -o src/cmd/tests/aso/tlock.c 'src/cmd/tests/aso/src@cmd@tests@aso@@tlock.c@exe/tlock.c.o' -L/opt/local/lib -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -pipe -Os -arch x86_64 -Wl,-headerpad_max_install_names -arch x86_64 src/lib/libast/libast.a src/lib/libast/libenv.a -lpthread -lm -liconv -ldl -Wl,-headerpad_max_install_names -Wl,-rpath,/path/to/build/src/lib/libast  
  "_iconv_close", referenced from:
      _init_ast_struct in libast.a(comp_setlocale.c.o)
```

**Expected results:**

Successful compilation

**Additional info:**

krader1961:
There was a period, roughly a year ago, when we needed to define
preprocessor symbol LIBICONV_PLUG on some platforms to build ksh. That
causes problems when building ksh via MacPorts. And testing on the
same platforms I was using a year ago when I introduced this hack shows
that it is apparently no longer needed. Presumably the real problem lay
elsewhere. Most likely in the various AST wrappers and `#define` symbols
that we've removed in the intervening time.

Resolves #932

### DIFF
--- a/config_ast.h.in
+++ b/config_ast.h.in
@@ -37,7 +37,6 @@
 #mesondefine CONF_LIBPATH
 #mesondefine _PROC_PROG
 #mesondefine _DARWIN_C_SOURCE
-#mesondefine LIBICONV_PLUG
 
 #mesondefine _ast_sizeof_int
 #mesondefine _ast_sizeof_long
--- a/features/meson.build
+++ b/features/meson.build
@@ -37,15 +37,7 @@ libexecinfo_dep = cc.find_library('execinfo', required: false, dirs: lib_dirs)
 # On some systems (e.g., OpenBSD) `iconv()` is in libiconv.
 libiconv_dep = cc.find_library('iconv', required: false, dirs: lib_dirs)
 
-# Under Cygwin the iconv shared library exports the symbols we need with a
-# `lib` prefix. So we don't want the LIBICONV_PLUG behavior on that platform.
-# On others we either want this or it's a no-op.
-if not (system == 'cygwin' or system == 'openbsd')
-    feature_data.set('LIBICONV_PLUG', 1)
-endif
-
-# On Cygwin the message catalog functions (e.g., `catopen()`) are in this
-# library.
+# On Cygwin the message catalog functions (e.g., `catopen()`) are in this library.
 libcatgets_dep = cc.find_library('catgets', required: false, dirs: lib_dirs)
 
 feature_data.set10('_hdr_execinfo', cc.has_header('execinfo.h', args: feature_test_args))


siteshwar:
There is a reference to issue #280 about adding `LIBICONV_PLUG` symbol. AFAICT this change was done for macOS and gives priority to non-system version of `libiconv`.

@catull Are you able to build from this pull request on your system ?

krader1961:
@siteshwar Note that #280 was merged by me with reservations and I had to subsequently amend the change to conditionally define that symbol. I have no trouble building and testing ksh with this change on macOS 10.12 and 10.13. Albeit without a non-standard libiconv installed.

ryandesign:
My understanding, from [the libiconv homepage](https://www.gnu.org/software/libiconv/) and from [this FreeBSD commit message](https://reviews.freebsd.org/rP398996) and other sources, is that `LIBICONV_PLUG` is not intended to be used by any software other than libiconv itself. It is only an option that may be used when compiling libiconv. Compiling libiconv with `LIBICONV_PLUG` means "don't use the `lib` prefix on `iconv` functions"; this "is a way to get good iconv support without having glibc-2.1". macOS doesn't use glibc; it uses libSystem, which doesn't contain any iconv functions. I assume Apple has used `LIBICONV_PLUG` when building the bundled copy of libiconv it ships in /usr. Compiling libiconv without `LIBICONV_PLUG` means "do use the `lib` prefix"; this is intended for most other libiconv installations and ensures that the libiconv functions don't interfere with the system's iconv functions. Whether libiconv is compiled with that option or not, as long as you `#include` the `<iconv.h>` matching the library you use, it should all work out, since the header will contain the right `#define`s so that you can always use the `iconv` functions without the `lib` prefix in your code.

krader1961:
I checked the beta (93v-) branch and am pretty sure we introduced this hack because the code had stuff like this in *src/lib/libast/features/iconv*:

```
printf("#undef  iconv_open\n");
printf("#define iconv_open      %siconv_open\n", lib);
printf("extern %siconv_t        %siconv_open(const char*, const char*);\n", lib, lib);
```

That ended up dynamically constructing the *ast_iconv.h* header which contained entries like this which varies depending on whether or not the feature test determined that the  "lib" prefix was needed:

```
#undef      iconv_open
#define iconv_open  _ast_iconv_open
```

For expediency we generated the *ast_iconv.h* using the legacy IFFE mechanism one time on Linux then checked it into the project. It was only when I started working on getting ksh to build on other platforms that we found the problem with that `#define` above. Since then I removed that header as part of my removal of all the AST locale code that ksh didn't need or that we didn't want to support.

krader1961:
Merged with commit comment amended.

PR #942, opened 2018-10-08
Add tests for `shift` builtin

siteshwar:


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -478,3 +478,30 @@ builtin -d alias 2>/dev/null && log_error "Deleting a special builtin should fai
...
+set -- This is a message
+shift 3
+actual="$1"
+expect=$'message'
+[[ "$actual" = "$expect" ]] || log_error "shift [n] does not shift n parameters" "$expect" "$actual"
+
+# ==========
+# shift <arithmetic expression> should shift result of arithmetic expression
+set -- This is a message
+shift 1+2
```
krader1961: Wow! I did not know that was legal. And why special-case `shift` that way when `shift $((1 + 2))`  works fine? Oh well, since we're stuck with it now adding a test to ensure it works is a good idea.

krader1961:
LGTM

PR #947, opened 2018-10-11
Implement an `eaccess()` fallback

### LINKED ISSUE #941: Build failure on OS X 10.9 and older: use of undeclared identifier 'AT_FDCWD'
**Description of problem:**

On the somewhat older OS X 10.9 and earlier, [building ksh fails](https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/68202/steps/install-port/logs/stdio) due to the use of `faccessat` and `AT_FDCWD` which weren't part of OS X until version 10.10.

**Ksh version:**

20181007

**How reproducible:**

Always

**Steps to reproduce:**
1. Build ksh on OS X 10.9 or older

**Actual results:**

```
../ast-7a4881196dac4ea2264ec0e8b95e6db7b4965ef9/src/lib/libast/port/astconf.c:483:61: warning: implicit declaration of function 'faccessat' is invalid in C99 [-Wimplicit-function-declaration]
                                if ((d = sfstruse(tmp)) && !eaccess(d, X_OK)) {
                                                            ^
./config_ast.h:238:23: note: expanded from macro 'eaccess'
#define eaccess(p, f) faccessat(AT_FDCWD, p, f, AT_EACCESS)
                      ^
../ast-7a4881196dac4ea2264ec0e8b95e6db7b4965ef9/src/lib/libast/port/astconf.c:483:61: error: use of undeclared identifier 'AT_FDCWD'
./config_ast.h:238:33: note: expanded from macro 'eaccess'
#define eaccess(p, f) faccessat(AT_FDCWD, p, f, AT_EACCESS)
                                ^
1 warning and 1 error generated.
```

**Expected results:**

Successful build.

**Additional info:**

krader1961:
Some systems we want to support don't provide `faccessat()` or `eaccess()`
or `eaccessuid()`. So implement build time feature tests and provide
fallback implementations.

Resolves #941

### DIFF
--- a/src/lib/libast/misc/fallbacks.c
+++ b/src/lib/libast/misc/fallbacks.c
@@ -6,10 +6,13 @@
 
 // We keep all these includes because it's simpler and cleaner to do this than wrap them in
 // `#if !_lib_mkostemp` type pragmas.
-#include <stdio.h>   // IWYU pragma: keep
-#include <stdlib.h>  // IWYU pragma: keep
-#include <string.h>  // IWYU pragma: keep
-#include <unistd.h>  // IWYU pragma: keep
+#include <errno.h>     // IWYU pragma: keep
+#include <fcntl.h>     // IWYU pragma: keep
+#include <stdio.h>     // IWYU pragma: keep
+#include <stdlib.h>    // IWYU pragma: keep
+#include <string.h>    // IWYU pragma: keep
+#include <sys/stat.h>  // IWYU pragma: keep
+#include <unistd.h>    // IWYU pragma: keep
 
 #include "ast.h"         // IWYU pragma: keep
 #include "ast_assert.h"  // IWYU pragma: keep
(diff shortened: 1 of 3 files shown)


src/lib/libast/misc/fallbacks.c
```diff
@@ -22,6 +25,65 @@ void do_not_use_this_fallback() {
...
+    return faccessat(AT_FDCWD, pathname, mode, AT_EACCESS);
+}
+#else
+// The platform doesn't have eaccess(), euidaccess(), or faccessat() so we have to roll our own.
+// This may not be optimal in that it calls access() when it might be possible to return an answer
+// based on what we already know. Since this is a fallback function we hope not to use we're
+// striving for simplicity and clarity.
+int eaccess(const char *pathname, int mode) {
+    // In case the user passed bogus bits for the mode tell them we can't determine what they want.
+    if (mode & ~(X_OK | W_OK | R_OK)) {
```
siteshwar: Add `F_OK` to this list. It's set to `0` on my Linux system, but unless you are sure that it's always `0`, it should be here.

Other option is to remove bogus bits like [glibc implementation](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/euidaccess.c;h=ec44b0f6961edcb19b8bf67df3be530437b4902c;hb=HEAD#l146).

src/lib/libast/misc/fallbacks.c
```diff
@@ -22,6 +25,65 @@ void do_not_use_this_fallback() {
...
+        errno = EINVAL;
+        return -1;
+    }
+
+    int uid = getuid ();
```
siteshwar: Use types `uid_t` and `gid_t` to store return value of `getuid()`, `getgid()` and related functions.

src/lib/libast/misc/fallbacks.c
```diff
@@ -22,6 +25,65 @@ void do_not_use_this_fallback() {
...
+    int gid = getgid ();
+    int euid = geteuid ();
+    int egid = getegid ();
+
+    // If we aren't suid or sgid then we can just use access().
+    // Similarly if we're just checking if the file exists just use access().
+    if ((euid == uid && egid == gid) || mode == F_OK) return access(pathname, mode);
```
siteshwar: if `mode == F_OK` then we don't need to call `access`, we have already checked file exists above and can return `0`.

siteshwar: Sorry. I see we are checking it after this line, so this change is fine.

src/lib/libast/misc/fallbacks.c
krader1961: I would expect it to be zero on all UNIX like systems but I'll include it here rather than remove bogus bits. If the user passes a bogus value it isn't safe to assume the what is left after stripping the invalid bits is meaningful.

src/lib/libast/misc/fallbacks.c
krader1961: Yes, of course.

PR #952, opened 2018-10-14
Use builtins in unit tests by default

### LINKED ISSUE #944: Enable all ksh builtins by default for unit tests
The `oclint` tool warned about code in *src/lib/libcmd/dirname.c* that could be simplified. Which caused me to notice that we have zero explicit tests of builtin `dirname` behavior. Also, there is only one use in all the ksh unit tests and it would only indirectly tell us if it was malfunctioning. This is probably true for some other builtins.

At a minimum we should a) add some basic tests of the `dirname` builtin and b) enable all builtins by default. The latter may or may not be trivial because a year ago when active maintenance resumed a lot of ksh builtin uses in unit tests were explicitly disabled.

krader1961:
The default behavior of unit tests should be to use builtins when
available and using external commands of the same name should be done
explicitly. This will help ensure builtins behave sensibly even if there
are insufficient explicit tests of each builtin.

Resolves #944

### DIFF
--- a/src/cmd/ksh93/tests/basic.sh
+++ b/src/cmd/ksh93/tests/basic.sh
@@ -257,7 +257,10 @@ fi
 
 cd $TEST_DIR || { log_error "cd $TEST_DIR failed"; exit 1; }
 print ./b > ./a; print ./c > b; print ./d > c; print ./e > d; print "echo \"hello there\"" > e
-chmod 755 a b c d e
+for x in a b c d e
+do
+    chmod 755 $x  # chmod builtin at the moment only handles one filename
+done
 x=$(./a)
 if [[ $x != "hello there" ]]
 then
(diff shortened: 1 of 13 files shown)


krader1961:
FWIW, There are lots of places in the tests which explicitly invoke an external command for no obvious reason. Such as the `printf=$(whence -p printf)` and the subsequent use of `$printf` in *subshell.sh*. Since this change is already quite large I did not attempt to figure out why those explicit invocations of an external command exist. I simply modified the test to use the external command that was found before the builtin of the same name was enabled.

dannyweldon:
While I haven't reviewed the code, I think this is an excellent idea.

siteshwar:
LGTM

krader1961:
I merged this with a workaround for issue #960.

PR #955, opened 2018-10-14
Add tests for `whence` builtin

siteshwar:


src/cmd/ksh93/tests/whence.sh
```diff
@@ -0,0 +1,63 @@
...
+#    off the -v option.
+builtin cat
+[[ $(whence -p cat) =~ "bin/cat" ]] || log_error "whence -p should search in PATH"
+builtin -d cat
+
+# ==========
+# -q Quiet mode. Returns 0 if all arguments are built-ins, functions, or are programs found on the
+#    path.
+actual=$(whence -q cat)
+[[ ${#actual} -eq 0 ]] || log_error "whence -q should have empty output"
```
krader1961: Probably worth adding an explicit test of the exit status prior to this line: `[[ $? -eq 0 ]] || ...`. Similarly I would add an explicit test for a command that doesn't exist:

```
actual=$(whence -q no-such-command)
[[ $? -eq 1 ]] || log_error "whence -q of non-existent command should report failure status"
```

krader1961:
LGTM other than the suggestion to add a couple more tests.

src/cmd/ksh93/tests/whence.sh
siteshwar: Added.

PR #958, opened 2018-10-16
Add tests for restricted shell

siteshwar:


krader1961:
LGTM

P.S., Thanks for embracing the style I prefer for such tests :smile: When I look at the legacy tests I grind my teeth due to lack of structure and poorly structured error messages.

Issue #959, opened 2018-10-16
`command -p` fails in loops


siteshwar:
**Description of problem:**
`command -p` fails in loops

**Ksh version:**
2017.0.0-devel-1949-gbeaf28cf

**How reproducible:**
Sometimes

**Steps to reproduce:**
```
set -e

rm -f /tmp/foo
touch /tmp/foo

for i in $(seq 679); do
  command -p file /tmp/foo
done
```

**Actual results:**
```
/tmp/test.sh: line 12: file: not found
```

**Expected results:**
There should not be any error.

**Additional info:**
```
# Works with:
#   command file /tmp/foo
# or when using a full path:
#   command -p /usr/bin/file /tmp/foo
# or when using a loop of 678.
```

=====
P.S.,  See my, krader1961, comment below about being able to reproduce this using the last official, ksh93u+, release on OpenSuse 42.

krader1961:
This only fails for me reliably on OpenSuse 42. Not macOS, FreeBSD, OpenBSD, or Ubuntu even with up to 10,000 iterations. Once it fails all external commands fail; e.g., just typing `ls /tmp/foo` results in `src/cmd/ksh93/ksh: _ls: line 139: ls: not found`.

orbea:
I can reproduce this on Slackware64 current as well with the `ksh93-20180412_71e6fdad-x86_64-1` package.

krader1961:
Weird. Earlier today I bisected the failure to commit c52d451e which doesn't make any sense since that change couldn't possibly cause this failure. And reverting that change, unsurprisingly, doesn't fix the problem. What's stranger is that now the test is failing with earlier commits.

So I did a new round of bisecting and am now quite confident it is due to commit 566ec823 to "Use fork()/exec() instead of posix_spawn() by default". Which **strongly suggests** the non-posix_spawn() variant of the project has not been stress tested in a long time and has at least one bug. That's because that change did exactly one thing. It added `-D_AST_no_spawnveg` when building ksh. Prior to commit 566ec823 I can't induce this failure. With commit 566ec823 it fails slightly more often than it passes.

krader1961:
After reverting commit 566ec82 the test now fails reliably for me on OpenSuse 42, Fedora 28 and Ubuntu 16.04 32-bit at iteration 604. It seems likely the issue is some sort of dynamic memory allocation bug. Of which the code has a surprisingly large number of instances as found by various tools such as ASAN, cppcheck, and the clang static analyzer.

[...]

PR #961, opened 2018-10-18
Fix a memory corruption issue in default path

siteshwar:
`astconf()` function internally uses `fmtbuf()` to allocate memory.
Memory allocated by this function is short-lived and should not be
directly used in global variables, so duplicate the string returned by
`astconf()` function.

Resolves: #959

### DIFF
--- a/src/cmd/ksh93/sh/path.c
+++ b/src/cmd/ksh93/sh/path.c
@@ -394,7 +394,13 @@ Pathcomp_t *path_nextcomp(Shell_t *shp, Pathcomp_t *pp, const char *name, Pathco
 static_fn Pathcomp_t *defpath_init(Shell_t *shp) {
     if (!std_path) {
         std_path = astconf("PATH", NULL, NULL);
-        if (!std_path) std_path = e_defpath;
+        if (std_path) {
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/path.c
```diff
@@ -393,7 +393,9 @@ Pathcomp_t *path_nextcomp(Shell_t *shp, Pathcomp_t *pp, const char *name, Pathco
 
 static_fn Pathcomp_t *defpath_init(Shell_t *shp) {
     if (!std_path) {
-        std_path = astconf("PATH", NULL, NULL);
+        // Value returned by astconf() is short lived, duplicate the string.
+        // https://github.com/att/ast/issues/959
+        std_path = strdup(astconf("PATH", NULL, NULL));
```
krader1961: Shouldn't the original, non strdup, assignment be done then this becomes:

```
if (std_path) {
    std_path = strdup(std_path);
    assert(std_path);
} else {
    std_path = e_defpath;
}
```

siteshwar: Fixed.

krader1961:
LGTM

PR #966, opened 2018-10-21
Add tests for math functions

siteshwar:


src/cmd/ksh93/tests/math.sh
```diff
@@ -178,3 +178,267 @@ function test_has_iszero
 # run tests
 test_arithmetric_expression_accesss_array_element_through_nameref
 test_has_iszero
+
+# ==========
+# Math functions
+
+# acos
+expect=3.14159265358979324
```
krader1961: This is very fragile. It assumes things that may not be true. For example, the use of IEEE 754 representation, 64 bit floats, and specific compiler behavior.  Those assumptions are likely to be true on most of the platforms we're currently using to validate the behavior but we shouldn't assume they will always be true. In fact, on WSL (Windows Subsystem for Linux) these tests fail:

```
<E> math[209]: atan failed
<E> expect: |1.5596856728972892|
<E> actual: |1.55968567289728908|
<E> math[281]: expm1 failed
<E> expect: |1.71828182845904524|
<E> actual: |1.71828182845904509|
```

And under Cygwin this failure is seen:

```
<E> math[239]: copysign failed
<E> expect: |-1.0|
<E> actual: |-9.09346574546873788e-656|
```

See https://bitbashing.io/comparing-floats.html for an exploration of the topic.

This level of precision (i.e., number of significant digits) also isn't needed to validate that these functions are behaving sensibly. Too, even if all the above assumptions are valid doing a simple equality check (as is done two lines below) is always wrong when floating point values are involved. We don't need to go as far as the recommendations in the link I provided above. For our purposes it is sufficient to run the output through `printf` with a reasonable scale. For example,

```
expect=3.14159265
actual=$(printf '%.8f' $(( acos(-1) )) )
```

Whether that `printf` subcommand is worth abstracting into a function that is used for generating the string representation of both the expected and actual values is an open question. For example,

```
function round {
    printf '%.8f' $1
}

expect=$(round 3.14159265358979324)
actual=$(round $(( acos(-1) )) )
```

krader1961:
LGTM. I'm okay with not using the `round` function on both expected and actual values but it would be safer and clearer to apply it to both values.

Issue #967, opened 2018-10-21
`signbit` prints 512 if number is negative


siteshwar:
**Description of problem:**
`signbit` prints `512` if number is negative

**Ksh version:**
2017.0.0-devel-1953-g89d7b802

**How reproducible:**
Always

**Steps to reproduce:**
```
echo $(( signbit(-1) ))
```

**Actual results:**
`512`

**Expected results:**
`1`

**Additional info:**
This bug is reproducible with `ksh 93u+`.

krader1961:
Also, this isn't technically a bug since the underlying math(3) function only guarantees to return a non-zero value if the input is negative. It does not guarantee to return one in that case.


@siteshwar What leads you to believe this should output `1` rather than `512` or any other non-zero value?

siteshwar:
> What leads you to believe this should output 1 rather than 512 or any other non-zero value?

Strictly speaking it may not be a bug, but it makes behavior more consistent with other math functions. For e.g. `iszero` returns either 0 or 1.

PR #968, opened 2018-10-22
Return 1 from `signbit` function if sign bit is set

siteshwar:
`signbit` function prints 512 on Linux if sign bit is set. It should
print 1 to make the result more clear.

Resolves: #967

### DIFF
--- a/src/cmd/ksh93/bltins/math.c
+++ b/src/cmd/ksh93/bltins/math.c
@@ -113,7 +113,7 @@ static Sfdouble_t local_nexttoward(int type_1, Sfdouble_t arg_1, int type_2, Sfd
 }
 
 static int local_signbit(Sfdouble_t a1) {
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/bltins/math.c
```diff
@@ -113,7 +113,7 @@ static Sfdouble_t local_nexttoward(int type_1, Sfdouble_t arg_1, int type_2, Sfd
 }
 
 static int local_signbit(Sfdouble_t a1) {
-    return signbit(a1);  //!OCLINT(constant conditional operator)
+    return !!signbit(a1);  //!OCLINT(constant conditional operator)
```
krader1961: Up to you whether you leave it this way but I'm not a fan of this trick. It confuses people and causes lint tools to complain about the double negation. I prefer the less confusing `signbit(a1) != 0`.

krader1961:
LGTM

Issue #969, opened 2018-10-23
enabling builtins for unit tests by default causes almost every test to fail under ASAN


krader1961:
After merging my change to enable builtins like `cat` and `chmod` by default for all unit tests I noticed that enabling ASAN causes almost every unit test to fail with feedback like below. This is on macOS.

```
==69479==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000251e at pc 0x0
0010734f161 bp 0x7fff58d580b0 sp 0x7fff58d580a8
READ of size 1 at 0x60200000251e thread T0
    #0 0x10734f160 in dllsread dllscan.c:363
    #1 0x10734a5fe in dllplugin dllplug.c:52
    #2 0x1070e96ea in path_absolute path.c:748
    #3 0x1070e6c2d in path_search path.c:607
    #4 0x106ef5ad7 in whence whence.c:267
    #5 0x106ef699e in b_whence whence.c:168
    #6 0x10713499d in sh_exec xec.c:1104
    #7 0x10713ff87 in sh_exec xec.c:1794
    #8 0x10704cb19 in exfile main.c:516
    #9 0x107050b2b in sh_main main.c:314
    #10 0x106e9d625 in main pmain.c:41
    #11 0x7fffc8353234 in start (libdyld.dylib:x86_64+0x5234)

0x60200000251e is located 0 bytes to the right of 14-byte region [0x602000002510,0x60200000251e)
allocated by thread T0 here:
    #0 0x107edcd43 in wrap_malloc (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x54d43)
    #1 0x1072f256b in ast_malloc vmbusy.c:25
    #2 0x10734c654 in dllsopen dllscan.c:237
    #3 0x10734a5e1 in dllplugin dllplug.c:50
    #4 0x1070e96ea in path_absolute path.c:748
    #5 0x1070e6c2d in path_search path.c:607
    #6 0x106ef5ad7 in whence whence.c:267
    #7 0x106ef699e in b_whence whence.c:168
    #8 0x10713499d in sh_exec xec.c:1104
    #9 0x10713ff87 in sh_exec xec.c:1794
    #10 0x10704cb19 in exfile main.c:516
    #11 0x107050b2b in sh_main main.c:314
    #12 0x106e9d625 in main pmain.c:41
    #13 0x7fffc8353234 in start (libdyld.dylib:x86_64+0x5234)

[...]

PR #970, opened 2018-10-23
Fix a bug found by ASAN after enabling builtins

krader1961:
Properly null terminate the dynamically allocated string. This reduces
unit test failures under ASAN on my system from 128 to 43.

Related #969


krader1961:
Merged

PR #973, opened 2018-10-24
Remove unused `cmdrecurse` function

siteshwar:
and related `SH_RUNPROG` macro.

This function and macro is unused, so remove it from sources.


krader1961:
LGTM. What's weird is that function isn't used in the ksh93u or ksh93v branches either. So it's not like it became unneeded as a result of the changes we've made.

PR #975, opened 2018-10-25
Remove unused function sh_waitnotify()

siteshwar:
This function was used by tksh and mkservice builtin. Since it's not
being used anymore, remove it from sources.


siteshwar:
> Is this not still useful to be kept for builtin developers as part of libshell? 

@dannyweldon That's a good question. It installs a callback that is executed while waiting for jobs to finish and might be useful for other plugins that are not shipped by upstream.

@krader1961 What's your take on this ?

krader1961:
I'm ambivalent. It's a complicated interface. It doesn't even appear that mkservice used it correctly since it made no attempt to handle the case where a handler was already installed. And it isn't even clear that it makes sense to have a stack of handlers given the shell doesn't support true concurrency of builtins. Too, it appears the sole use case for `sh_waitnotify()` is something like `mkservice` of which there can be only one instance running at any point in time. Given the earlier discussion regarding the removal of `mkservice` I'll bet real money it never gets reinstated. At least not unless there are additional architectural changes such as the introduction of support for threads.

So on balance my vote is to remove it. And also remove the related `waitevent` var and the code predicated on it since it is otherwise unused.

krader1961:
Note that my objection to retaining the `sh_waitnotify()` interface is not due solely to its complexity. Let alone its use of the callback pattern. As a grey beard who has used a huge number of frameworks that use callbacks I have no objection to the pattern. Callbacks are an extremely useful pattern. My objection is because it appears `sh_waitnotify()` was introduced solely to support a specific builtin: `mkservice`. And the definition of the "waitnotify" API is problematic. What does it mean to install a new notify handler when one is already in effect? A google search reveals no uses outside the `mkservice` builtin we have removed.

[Nuke it from orbit.](https://www.urbandictionary.com/define.php?term=nuke%20it%20from%20orbit)

dannyweldon:
> What does it mean to install a new notify handler when one is already in effect?

Looking at `mkservice.c` and how it uses `sh_fdnotify()`, I'm guessing that the correct usage would have been to save the handle returned and call it first inside the new callback, but for some reason that was not done in `mkservice`, so I could be wrong.

If you are going to remove it then, can you remove it from `shell.h` as well?

krader1961:
LGTM

PR #978, opened 2018-10-28
Require a compatible version of clang-format

krader1961:
Another contributor, @siteshwar at RedHat, recently implemented some
automation to run `bin/style --all` on a regular basis. Unfortunately the
system he ran it on has an old version of `clang-format` which produces
incorrect formatting. Or at least inconsistent with newer versions of
`clang-format` which causes unwanted thrashing of the code style. So
impose a minimum acceptable version of the `clang-format` tool.

### DIFF
--- a/bin/style
+++ b/bin/style
@@ -15,10 +15,24 @@
 # newline makes it safe to handle file names with spaces.
 IFS=$'\n'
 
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #981, opened 2018-10-29
Fix a ASAN stack-use-after-return bug

### LINKED ISSUE #980: ASAN finds a stack var use after return bug in `arith()`
From the `subshell.sh` unit test:

```
==97376==ERROR: AddressSanitizer: stack-use-after-return on address 0x000104a27370 at pc 0x000102c02e25 bp 0x7ffeed277140 sp 0x7ffeed277138
READ of size 8 at 0x000104a27370 thread T0
    #0 0x102c02e24 in nv_restore subshell.c:326
    #1 0x102c08eec in sh_subshell subshell.c:612
    #2 0x102c2d280 in sh_exec xec.c:1649
    #3 0x102c2f32c in sh_exec xec.c:1794
    #4 0x102b3721b in exfile main.c:516
    #5 0x102b3b37f in sh_main main.c:314
    #6 0x10297f585 in main pmain.c:41
    #7 0x7fff6545d084 in start (libdyld.dylib:x86_64+0x17084)

Address 0x000104a27370 is located in stack of thread T0 at offset 112 in frame
    #0 0x102a685bf in arith arith.c:556

  This frame has 3 object(s):
    [32, 48) 'n.addr'
    [64, 72) 'str' (line 559)
    [96, 160) 'node' (line 749) <== Memory access at offset 112 is inside this variable
```

The problem is this block of code in the `arith()` function:

```
            } else if ((Namfun_t *)lvalue->ptr && !nv_hasdisc(np, &ENUM_disc) &&
                       !nv_isattr(np, NV_INTEGER)) {
                Namval_t *mp, node;
                mp = ((Namfun_t *)lvalue->ptr)->type;
                memset(&node, 0, sizeof(node));
                nv_clone(mp, &node, 0);
                nv_offattr(&node, NV_RDONLY | NV_NOFREE);
                nv_putval(&node, np->nvname, 0);
                if (nv_isattr(&node, NV_NOFREE)) return nv_getnum(&node);
            }
```

The problem here is the stack var `node` which is associated with `np->nvname` and thus needs to persist after `arith()` returns.

krader1961:
There is a code path in the `arith()` function which can result in
returning a pointer to a stack var. This fixes that at the cost of a
possible memory leak.

Fixes #980

### DIFF
--- a/src/cmd/ksh93/sh/arith.c
+++ b/src/cmd/ksh93/sh/arith.c
@@ -746,13 +746,19 @@ static_fn Sfdouble_t arith(const char **ptr, struct lval *lvalue, int type, Sfdo
                 lvalue->ptr = (void *)nv_hasdisc(np, &ENUM_disc);
             } else if ((Namfun_t *)lvalue->ptr && !nv_hasdisc(np, &ENUM_disc) &&
                        !nv_isattr(np, NV_INTEGER)) {
-                Namval_t *mp, node;
-                mp = ((Namfun_t *)lvalue->ptr)->type;
-                memset(&node, 0, sizeof(node));
-                nv_clone(mp, &node, 0);
-                nv_offattr(&node, NV_RDONLY | NV_NOFREE);
-                nv_putval(&node, np->nvname, 0);
-                if (nv_isattr(&node, NV_NOFREE)) return nv_getnum(&node);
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #982, opened 2018-10-29
Coverity "Dereference after null check" bug

krader1961:
It appears that `shp->jmplist` has to be non-NULL and that the predicate
for it being non-NULL that this change removes is a holdover from a time
when it could be NULL.  Looking at the entire project there are a couple of
places with similar tests but far more that simply assume it is non-NULL.

Coverity CID#279531

### DIFF
--- a/src/cmd/ksh93/sh/fault.c
+++ b/src/cmd/ksh93/sh/fault.c
@@ -39,6 +39,7 @@
 
 #include "ast.h"
 #include "ast_aso.h"
+#include "ast_assert.h"
 #include "error.h"
 #include "fault.h"
 #include "fcin.h"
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/sh/fault.c
```diff
@@ -432,13 +433,14 @@ void sh_chktrap(Shell_t *shp) {
 // Exit the current scope and jump to an earlier one based on pp->mode.
 //
 void sh_exit_20120720(Shell_t *shp, int xno) {
-    struct checkpt *pp = (struct checkpt *)shp->jmplist;
-    int sig = 0;
     Sfio_t *pool;
+    int sig = 0;
+    struct checkpt *pp = (struct checkpt *)shp->jmplist;
+    assert(pp);
```
krader1961: Note that this `assert(pp)` isn't really needed. If `pp` is NULL the program should die when the dereference three lines below occurs. This is here solely to make it blindingly obvious we require it to be non-NULL even though the prior version of the code makes it appear it could be NULL at this juncture.

siteshwar:
LGTM

PR #985, opened 2018-10-30
Remove unused function `sh_utol()`

siteshwar:


src/cmd/ksh93/sh/string.c
```diff
@@ -218,27 +218,6 @@ void sh_trim(char *sp) {
...
 
-//
-// copy <str1> to <str2> changing upper case to lower case
-// <str2> must be big enough to hold <str1>
-// <str1> and <str2> may point to the same place.
-//
-// assume str1!=0 && str2!=0
-// return x satisfying strlen(in str1)==strlen(in str2);
-//
-void sh_utol(char const *str1, char *str2) {
```
krader1961: Not only unused but broken for any locale other than ASCII or a ISO 8859 variant.

krader1961:
LGTM

Issue #855, opened 2018-09-11
Cleanup `if (nv_isattr(np, NV_DOUBLE) == NV_DOUBLE)`


krader1961:
There are too many places in the code which do something like this:

```
if (nv_isattr(np, NV_DOUBLE) == NV_DOUBLE)
```

The explicit equality test is pointless since `nv_isattr()`, by definition, only returns true if the flag is set. Thus the explicit equality test does nothing more than cause the reader to wonder what the point of the comparison is and whether there is something unexpected about the behavior of `nv_isattr()`.

Note that there are other places which test if just one of two flags is set. Such as this one:

```
if (nv_isattr(np, NV_INT16P | NV_DOUBLE) == NV_INT16
```

I'll bet someone saw that and concluded it made sense to use the same pattern even when a single flag was being tested. When what they should have concluded is the multiple flag test should have been rewritten.

P.S., I would love to be able to interview the people who established this idiom to understand why they thought it was a good idea.

krader1961:
FFS! This idiom appears to be due to preprocessor symbols like this one:

```
define NV_NOTSET (NV_INTEGER | NV_BINARY)
```

If you just do `nv_isattr(np, NV_NOTSET)` it evaluates to true whether either flag is set rather than if both are set. Again, I would love to interview the people who established this idiom to understand why they thought it was a good idea. Because the "nv_isattr()" naming implies that it returns true if both of the flags are set. The names of too many functions/macros obscure the actual purpose. For example, `roundof()`. Most English speakers would have named that `round_to()`.

krader1961:
Note that there are a large number of places which use this idiom:

```
if (nv_isattr(np, NV_NOFREE))
```

You can't have it both ways. Either it should always be `if (nv_isattr(ptr, flags) == flags)` or just `if (nv_isattr(ptr, flags))`. Because you shouldn't have to know if `flags` is a single or multiple bits.

krader1961:
This may be the poster child example for why the existing `nv_isattr()` implementation needs to be changed. From *src/cmd/ksh93/sh/name.c*:

```
        if (nv_isattr(np, NV_LJUST | NV_RJUST) &&
            nv_isattr(np, NV_LJUST | NV_RJUST) != (NV_LJUST | NV_RJUST)) {
            tofree = NULL;
        }
```

Quick! Tell me what that `if()` condition means.

Answer: It's testing if either attr flag `NV_LJUST` or `NV_RJUST` are set but not both.

[...]

PR #987, opened 2018-10-31
Coverity out-of-bounds access error

krader1961:
Incorrect loop constructs in `check_limits()` causes dereference past
the end of the `fltconst` and `intconst` arrays.

Coverity CID#253767

### DIFF
--- a/src/cmd/ksh93/sh/arith.c
+++ b/src/cmd/ksh93/sh/arith.c
@@ -108,7 +108,7 @@ struct Mathconst {
 #define M_SQRT1_2l 0.7071067811865475244008443621048490L
 #endif
 
-// The firs three entries cann't be moved or it will break the code.
+// The first three entries cann't be moved or it will break the code.
 static const struct Mathconst Mtable[] = {
     {"1_PI", M_1_PIl}, {"2_PI", M_2_PIl},   {"2_SQRTPI", M_2_SQRTPIl},
     {"E", M_El},       {"LOG2E", M_LOG2El}, {"LOG10E", M_LOG10El},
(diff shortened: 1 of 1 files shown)


krader1961:
I pushed again to rebase on top of the `master` branch.

src/cmd/ksh93/sh/arith.c
```diff
@@ -295,71 +304,79 @@ static const Fltconst_t fltconst[] = {
...
+static_fn Namval_t *check_limits(Shell_t *shp, char *cp) {
+    char *ep = strrchr(cp, '.');
+    if (!ep) return NULL;  // no attribute name
+
+    *ep = 0;
+    Namval_t *np = nv_open(cp, shp->var_tree, NV_VARNAME | NV_NOADD | NV_NOFAIL);
+    *ep++ = '.';
+    if (!np || !nv_isattr(np, NV_INTEGER)) return NULL;  // var type not integer or float
+
+    if (nv_isattr(np, NV_DOUBLE) == NV_DOUBLE) {
```
siteshwar: IIRC you questioned this pattern, why are you using it ?

krader1961: Yes I did. See issue #855. But changing it is outside the scope of this fix. And omitting the `== NV_DOUBLE` is only safe if you know the symbol has a single bit set.

krader1961: Which is precisely why I want to change the definition of `nv_isattr()`. Notice that the preceding  use omits the `== NV_INTEGER` test. The problem here is that this is the definition for `NV_DOUBLE`:

```
#define NV_DOUBLE (NV_INTEGER | NV_ZFILL)
```

This is without a doubt one of the worst APIs I've seen I've seen in this project.

Issue #983, opened 2018-10-30
Rename the _20120720 API variants and remove the legacy interfaces


krader1961:
We intend to maintain the ability to build existing 3rd party ksh plugin source. But with all the changes that have been made since ksh93u+ was released there should be zero expectation that existing plugin binaries will be compatible. This means we can simplify the project by removing legacy interfaces such as `sh_exit()` prior to the introduction of `sh_exit_20120720()`. This also allows us to eliminate *src/cmd/ksh93/include/shellapi.h*.

krader1961:
When I say

> ... there should be zero expectation that existing plugin binaries will be compatible.

I'm referring to changes such as removing the dependency on AST Vmalloc. If an existing plugin uses that subsystem the dynamically allocated buffers it may allocate can't be passed to the system's `free()` function. Which the ksh code might do.

DavidMorano:
> I'm referring to changes such as removing the dependency on AST Vmalloc.

Although, in my view, VMALLOC should not be missed by anyone (serious about clean code), someone familiar with various advanced linking techniques should be able to continue to bind (or pre-bind) their plugin code to an AST library (or any extra library) if they really want to, even in the face of using different subroutines to satisfy dependencies using the same names (like |malloc()| and |free()|). In some linking systems this can be done using "symbolic" linking, a sort of pre-binding before run-time dynamic name lookup. But investigate the linking options for your target system for possibilities. But in the specific case of dynamic memory allocation, my real advice would be to lose your dependency on VMALLOC entirely if you think you still have one.

For myself, almost none of LIBAST is especially interesting to use for anything. Also, a substantial amount of LIBAST is not thread-safe, and that makes it instantly unusable when all other code, including plugins and their dependent libraries, is multithread hot. Also, for those having or desiring to create multithread hot plugins, I found VMALLOC to be an instant process crasher. There was something (forget the details now) that prevented VMALLOC from working properly in a multithread environment, so it had to go. I compiled VMALLOC out of KSH many years ago to get multithreaded plugins (and background threads).

PR #988, opened 2018-11-01
Remove unneeded redefinition of `sh_funscope`

siteshwar:


krader1961:
This is a good example of why I opened issue #983. Rather than just removing these unused legacy interfaces one at a time I'd prefer to remove them all at the same time and rename the variants that include a date stamp to the simple name. If we ever again need to make a non-backward compatible API change I'd prefer to do it in a different manner.

siteshwar:
@krader1961 Ok, shall I merge this ? I usually review code for each file one by one and find these unused functions. If you want to remove them all at once, I will stop opening such pull requests.

krader1961:
Since we're talking about what I call a "mechanical" change that doesn't affect functionality it makes more sense to do it all as a single commit. Just remove all the legacy interfaces, rename the new ones introduced in 2012 to their legacy name, and remove *shell_api.h*.

siteshwar:
Ok, will close this for now.

PR #989, opened 2018-11-02
Remove unused dll_lib.c code

krader1961:
I went to fix this warning from oclint:

src/lib/libdll/dll_lib.c:173:9: String copy function overflows destination buffer

But since none of the code in that module is used, not even in the old
beta branch, there isn't much point maintaining it so remove it.


siteshwar:
LGTM

PR #995, opened 2018-11-04
Update readme regarding mailing lists

krader1961:


siteshwar:
I think we should message Lefteris to learn about status of mailing list before merging this.

krader1961:
> I think we should message Lefteris to learn about status of mailing list before merging this.

Okay, email sent. But since the last message to either list was ten months ago according to marc.info I think it is safe to conclude those mailing lists have run their course. I haven't seen a message in my inbox from those mailing lists since then. At this point even if AT&T said they intend to fix whatever problem is causing email to be undeliverable I think we should stop suggesting people use those lists to communicate with us.

siteshwar:
If these mailing lists are set to be retired, then we will have to set up at least one list to communicate with the community (for announcements etc.).

krader1961:
Still no response from the AT&T team regarding the mailing lists.

siteshwar:
We have received response from AT&T that they were not aware these lists were not working. Also, it's unlikely that these lists will be kept well maintained in future. Please merge this and It's time to set up a new list.

Issue #1002, opened 2018-11-07
Change all structure allocations that include room for additional data


krader1961:
I decided to tackle the ASAN bug shown below that occurs when running the `substring` unit test. The problem is due to the last member of `struct index_array` being a variable size array. When it is resized the structure may be relocated. Which invalidates any cached pointers to the original structure. But most of the code assumes that won't happen. I'm willing to bet most of the heap-use-after-free bugs are due to this pattern. Specifically this pattern:

```
struct index_array {
    // fixed size members....
    // followed by a variable size structure member
    union Value val[1];   // array of value holders
}
```

The idea is that such a definition avoids having to do two allocations and free calls and is thus more efficient. The problem is that this isn't safe if the trailing member is resized after the initial allocation as it can cause the entire structure to be moved to a different memory location. Which, given how many places in the code cache pointers to such structures even if just in a stack local var, is guaranteed to cause problems.

```
==92875==ERROR: AddressSanitizer: heap-use-after-free on address 0x614000022c70 at pc 0x00011009abe4 bp 0x7ffedfcf9b50 sp 0x7ffedfcf9b48
READ of size 8 at 0x614000022c70 thread T0
    #0 0x11009abe3 in varsub macro.c:1306
    #1 0x1100879c6 in copyto macro.c:599
    #2 0x11008d423 in sh_macexpand macro.c:250
    #3 0x1100a427a in sh_macpat macro.c:414
    #4 0x1101b45c9 in sh_exec xec.c:2320
    #5 0x1101aa097 in sh_exec xec.c:1784
    #6 0x1100b238b in exfile main.c:515
    #7 0x1100b6502 in sh_main main.c:313
    #8 0x10fefad95 in main pmain.c:41
    #9 0x7fff7108008c in start (libdyld.dylib:x86_64+0x1708c)

0x614000022c70 is located 48 bytes inside of 408-byte region [0x614000022c40,0x614000022dd8)
freed by thread T0 here:
    #0 0x1105c21bd in wrap_free (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x551bd)
    #1 0x110363e6b in ast_free vmbusy.c:50
    #2 0x10fff2590 in array_grow array.c:651
    #3 0x10ffefb94 in nv_putsub array.c:949
    #4 0x11009aac7 in varsub macro.c:1303
    #5 0x1100879c6 in copyto macro.c:599
    #6 0x11008d423 in sh_macexpand macro.c:250
[...]

PR #1003, opened 2018-11-08
Struct index array

krader1961:
This PR does not address the issue described by #1002. It merely lays the groundwork for doing so by simplifying the code and removing some bogosities.


krader1961:
I've merged these changes since they are a) trivial, and b) don't actually change any behavior since they only change spelling of a symbol or are similarly innocuous and safe.

PR #1007, opened 2018-11-11
[kshbugbot] Fix code styling

kshbugbot:


siteshwar:
I am not sure why you are closing these pull requests without merging. But in case if there are any issues with them, just leave them open so that the bot does not redo the pull request. I will fix it when I get back from holidays.

Issue #1011, opened 2018-11-20
Shall  `cd` builtin support `@` command line option ?


siteshwar:
`cd` builtin supports `@` command line option that can be used to open files under `/dev/file/xattr` path. This option was not present in `ksh93u+` release. This issue tracks investigating if it is worth keeping this option in next release.

krader1961:
On my system that path doesn't exist:

```
$ ls /dev/file
ls: /dev/file: No such file or directory
```

In fact it doesn't exist on any of the my VMs that I use to test things on UNIX like OS's. If this feature is part of 3DFS support, or something similar, then it should be excised.

PR #1012, opened 2018-11-20
Remove `@` command line option from `cd` builtin

siteshwar:
This option is used to reference files under '/dev/file/xattr' which is
a special path handled by ksh. Since we have already dropped support for
such paths through previous changes, we should drop this command line
option too.

Resolves: #1011


krader1961:
LGTM

PR #1014, opened 2018-11-22
Extend tests for `enum` builtin

siteshwar:

PR #1015, opened 2018-11-22
Fix a crash while running `enum` builtin

siteshwar:
Running `enum` builtin without any arguments causes a crash. Fix it by
restoring if condition from last stable release.

### DIFF
--- a/src/cmd/ksh93/bltins/enum.c
+++ b/src/cmd/ksh93/bltins/enum.c
@@ -302,11 +302,12 @@ int b_enum(int argc, char **argv, Shbltin_t *context) {
     }
 
     argv += opt_info.index;
-    if (error_info.errors) {
+    argc -= opt_info.index;
+    if (error_info.errors || argc != 1) {
         error(ERROR_USAGE | 2, "%s", optusage(NULL));
         return 1;
     }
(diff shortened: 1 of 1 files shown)


siteshwar:
Added test case for it in #1014.

src/cmd/ksh93/bltins/enum.c
```diff
@@ -302,11 +302,11 @@ int b_enum(int argc, char **argv, Shbltin_t *context) {
     }
 
     argv += opt_info.index;
-    if (error_info.errors) {
+    if (error_info.errors || !*argv || *(argv + 1)) {
```
krader1961: This is an atypical, and confusing, way to do the test for the wrong number of arguments. This should be

```
argc -= opt_info.index;
if (error_info.errors || argc != 1) {
```

src/cmd/ksh93/bltins/enum.c
```diff
@@ -302,11 +302,11 @@ int b_enum(int argc, char **argv, Shbltin_t *context) {
...
         error(ERROR_USAGE | 2, "%s", optusage(NULL));
         return 1;
     }
-    if (!*argv) sh_outenum(shp, sfstdout, NULL);
```
krader1961: What in the world do you think this was meant to do?

PR #1019, opened 2018-11-26
Eliminate use of two obsolete preprocessor symbols

### LINKED ISSUE #1018: Preprocessor symbols `u370` and `cray`
While looking into issue #1008 I noticed preprocessor symbols `u370` and `cray`. AFAICT there is no reason for either symbol to exist today in this project source code. The former because we no longer support running IBM mainframes using EBCDIC; only IBM mainframe VMs that use ASCII. And the latter because Cray Computer Corporation no longer exists.

krader1961:
Resolves #1018


siteshwar:
LGTM

Issue #240, opened 2017-12-19
Remove support for SHOPT_AUDIT and SHOPT_AUDITFILE?


krader1961:
While working on issue #210 I noticed that the legacy build defined the following two symbols but the new Meson based build does not:

```
-DSHOPT_AUDIT
-DSHOPT_AUDITFILE="/etc/ksh_audit"
```

The source code implies this has something to do with recording keystrokes. But googling turns up a few pages such as https://www.ibm.com/developerworks/aix/library/au-korn93/ which says:

> New features of the Korn Shell provide system administrators and management with the ability to monitor, track, record, and audit every command executed by any user of a system. This is different from the normal shell history, and provides detailed information that includes date, time, tty, user, and the command. This information can be stored locally or transmitted in real time to a remote logging system.

This [blog post](https://blog.fpmurphy.com/2008/12/ksh93-auditing-and-accounting.html) contains more detail. The author of that article concludes with this paragraph:

> Most of the information provided in this post is not documented in a single place anywhere that I can find by searching the Internet.  The ksh93 man page does not mention either the accounting or auditing facilities.  Even the ksh93 source code is somewhat vague.  I gleened most of this information by studying the code in ../src/cmd/ksh93/edit/history.c.  If I got anything wrong, please let me know so that I can update this post.

I recommend removing this feature. The code protected by those symbols is so miniscule it is reasonable to assume the feature itself is a security hole. Furthermore, this "feature" is trivial to bypass. Too, I've never seen this feature in any other shell I've used or even proposed. Therefore the added complexity can't be justified.

krader1961:
Not surprisingly you can also find people reporting problems building ksh with this feature enabled: https://access.redhat.com/solutions/234843

See also https://administratosphere.wordpress.com/2011/05/20/logging-every-shell-command/

krader1961:
Based on feedback in some other issues it appears that at least a couple of distros have been building ksh with this feature enabled. The question is whether they knew they were doing so (i.e., enabled it deliberately) and is there anyone actually using this feature? I suspect the answers are "no" and "no". If you have reason to believe otherwise please speak up and comment on this issue.

fpmurphy:
@krader1961.  Auditing of shell usage is a required feature in most high security facilities such as DoD, NSA, NRO, etc.  You are not getting feedback from the appropriate people because you are not engaging the appropriate people..  If anything, auditing in ksh93 should be enhanced.

krader1961:
[...]

PR #1022, opened 2018-11-26
Use correct filename for audit file

siteshwar:
`AUDIT_FILE` contains name of the file that should be used as audit
file. Use it's value instead of stringifying it. Also, remove unused
macro `stringify`.

Related: #240


krader1961:
Good catch. LGTM. Is there a trivial way to unit test this?

PR #1024, opened 2018-11-27
Add tests for unrecognized command line options in builtins

siteshwar:


krader1961:
LGTM. Technically a couple of these changes aren't strictly backward compatible but I would argue anyone relying on that broken behavior deserves to have their script broken by this change.

PR #1027, opened 2018-11-27
Improve tests for `echo` builtin

siteshwar:
Running `echo` builtin inside subshell eats last newline character.
Instead use expect based tests to check for newline characters.

### DIFF
--- a/src/cmd/ksh93/tests/echo.exp
+++ b/src/cmd/ksh93/tests/echo.exp
@@ -1,0 +1,44 @@
+# Test for echo builtin
+set pid [spawn $ksh]
+expect_prompt
+
+# ==========
+log_test_entry
+send "echo -n hello world; echo EOT\r"
+expect "\r\nhello worldEOT\r\n" {
+    puts "echo -n does not print new line"
+}
+expect_prompt
+
+# ==========
+log_test_entry
+send "echo -e 'hello\nworld'\r"
+expect "\r\nhello\r\nworld\r\n" {
+    puts "echo -e interprets escape sequences"
+}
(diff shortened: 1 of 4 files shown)


src/cmd/ksh93/tests/echo.exp
```diff
@@ -0,0 +1,36 @@
+# Test for echo builtin
+set pid [spawn $ksh]
+expect_prompt
+
+# ==========
+log_test_entry
+send "echo -n hello world\r"
+expect "\r\nhello world" {
```
krader1961: This doesn't do what you think it does. Hint: Does this pass or fail if you change the command to "echo hello world\r"?

There are a couple of ways to fix this. I would fix it by replacing these two statements with this:

```
send "echo -n hello world; echo EOT\r"
expect "\r\nhello worldEOT\r\n" {
```

FWIW, I frequently make this mistake. Writing robust expect based tests is hard. Which is why I will often deliberately change the test to violate the condition being tested to confirm the expect pattern does what I'm expecting.

src/cmd/ksh93/tests/echo.exp
```diff
@@ -0,0 +1,36 @@
...
+send "echo -e 'hello\nworld'\r"
+expect "\r\nhello\r\nworld\r\n" {
+    puts "echo -e interprets escape sequences"
+}
+expect_prompt
+
+# ==========
+log_test_entry
+send "echo -ne 'hello\nworld'\r"
+expect "\r\nhello\r\nworld" {
```
krader1961: Same as above you need to add a sentinal value.

PR #1029, opened 2018-11-28
Add test for `--man` option

siteshwar:
Running builtins with `--man` option should show their respective
documentation.

### DIFF
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -25,6 +25,13 @@ for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|\[|
     [[ "$actual" =~ "$expect" ]] || log_error "$name should show usage info on unrecognized options" "$expect" "$actual"
 done
 
+# ==========
+for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|hash|type|source|\[|:)'); do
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -25,6 +25,13 @@ for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|\[|
     [[ "$actual" =~ "$expect" ]] || log_error "$name should show usage info on unrecognized options" "$expect" "$actual"
 done
 
+# ==========
+for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|hash|type|source|\[|:)'); do
+    actual=$($name --man 2>&1)
```
krader1961: This is good enough. But if you want to be extra cautious I'd recommend stripping the ANSI escape sequences and collapsing newlines and whitespace and capturing just the first few lines:

```
actual=$(umask --man 2>&1 | head -5 | sed $'s,\x1B\[[0-9;]*[a-zA-Z],,g'  | tr -s '\n ' ' ')
```

That makes it easy to tighten up the pattern match:

```
expect="NAME $name - "
[[ "$actual" =~ "^$expect" ]]
```

krader1961:
LGTM

src/cmd/ksh93/tests/builtins.sh
siteshwar: Point taken!

PR #1031, opened 2018-11-29
Add few more `printf` related tests

siteshwar:
- Add test for `print -u` without any arguments
- Add test for `printf "%Q"`
- Add test for `printf "%p"`


siteshwar:
Also, `%p` and `%Q` arguments of `printf` are not documented. That's why I missed writing test for them earlier. We should document them.

krader1961:
LGTM

Issue #507, opened 2018-05-06
Embedded command documentation needs to be moved outside the C source files


krader1961:
See issue #503 for an example where the information in `man ksh` differs from `some_command --help`. In that issue it is with respect to the `hist` command but there are plenty of others. It is bad enough that having two sources for command documentation inevitably leads to such discrepancies. Embedding the command help text in the C source also makes it hard for a human to read and edit.

Contrast this state of affairs with how the `fish` shell does it. It supports both `builtin_command --help` __and__ `man builtin_command`. And due to how the help text is managed it is guaranteed they will always be in agreement. This project should be able to do the same thing.

P.S., If and when this is resolved it would be a good idea to switch the markup language from the ancient troff style to Markdown.

krader1961:
For what it's worth the fish shell has decided to switch from Doxygen to [Sphinx](http://www.sphinx-doc.org/en/master/). Something I strongly recommend for this project. The ancient Troff markup language is all but unknown to anyone younger than myself (50+ years old). Not to mention that Troff markup is awful by modern sensibilities and there are few tools to translate it to other formats.

krader1961:
Another example is the `whence` builtin. Typing `whence --help` says it supports a `-q` flag which works as expected. But the man page makes no mention of that flag.

I'm tempted to mark this as a blocking bug for the next ksh release. It should be impossible for the man page and "internal" documentation (e.g., "a_builtin --help") to disagree. Not just because the people modifying the code keep two distinct sources of truth in sync. But because there is only one source of truth.

dannyweldon:
The fact that the ast `getopts` builtin embeds documentation means that the `getopts` version should be the most correct version, as it has to be consistent with the options themselves.

The `getopts` builtin has an option to dump the raw `nroff` code:

    thecmd --nroff

which means one possibility is to generate sections of the `ksh` man page from the builtins themselves using the above, which is currently dumping a full man page, but it could be modified to dump just the sections needed similarly to:

    thecmd --help

Or the output of `thecmd --nroff` could be parsed to extract just the relevant information.  This would enforce the DRY principle, as opposed to currently, where there are two sources of truth.

The `getopts` builtin also has an api version specification which could be increased to allow a newer format that allows the use of embedded markdown syntax.

krader1961:
@siteshwar and I discussed this issue in the ksh93/users Gitter room today. What follows are the comments I made.

[...]

PR #1032, opened 2018-11-29
Document `%Q` and `%p` options of `printf`

siteshwar:


krader1961:
LGTM although this, once again, points out how important it is to address issue #507 since this leaves the format specifiers undocumented in the "man ksh" output.

Issue #820, opened 2018-08-30
Shall `read` builtin support json format ?


siteshwar:
Support for `json` format was added with [2014-12-24-beta](https://github.com/att/ast/commit/0bc6abf9e) commit. There is currently [one bug](#39) filed against it and I suspect support for it may be very fragile. Is it something that we should target for next release ?

siteshwar:
I am not against adding support for `json`, but I believe this is something that we should not target for next release.

siteshwar:
I see this feature would be useful, but support for `json`  could not be considered even experimental at this stage. For e.g. executing this code in interactive mode:

```
read -m json foo
{ "bar" : "baz" }
```

causes shell to exit. I would like to see this feature disabled for next release and it should be reintroduced in future releases with better testing. There is no point in shipping features that are known to be broken. I will prefer not to spend time on fixing such features before next release.

krader1961:
I have no objection to disabling the feature if that can be done by introducing no more than a couple of `#if` directives at strategic points; e.g., in `b_read()` by commenting out the json entry from the `methods[]` array. If it requires a lot of `#if` directives scattered throughout the code we probably need to talk about completely removing the feature.

krader1961:
Note that in addition to `read -m json` not working the code also does now produce valid JSON output for a trivial array with an int and a string:

```
KSH PROMPT:1: x=(1, 'x')
KSH PROMPT:2: typeset -p x
typeset -a x=(1, x)
KSH PROMPT:5: print -j x
(
        1,
        x
)
KSH PROMPT:8: printf '%(json)B\n' x
(
        1,
        x
)
```

I've changed my mind about keeping this code. If JSON handling were confined to *bltins/read.c* and *bltins/print.c* I'd be willing to just disable its inclusion in a default build. But it is very invasive to the *sh/nvtree.c* module and the "namval" code has so many problems I don't want to have to also worry about the json support causing additional problems.

While I've never used it and it does appear somewhat awkward to use it looks like the [jq](https://stackoverflow.com/questions/43373176/store-json-directly-in-bash-script-with-variables) utility is the generally accepted solution to the problem of working with JSON in bash and similar shells.

dannyweldon:
While `jq` is a nice utility, I believe the purpose of the ability for `ksh` to be able to read and write `json` is more about being able to parse `json`-formatted config files and convert them to `ksh` data structures, something that `jq` cannot do, nor can the other shells handle.

[...]

PR #1033, opened 2018-11-29
Disable json support in `read` builtin

siteshwar:
Related: #820

### DIFF
--- a/src/cmd/ksh93/bltins/read.c
+++ b/src/cmd/ksh93/bltins/read.c
@@ -79,6 +79,9 @@ struct Method {
     void *fun;
 };
 
+// Support for json in `read` builtin seems very unstable, so this code is kept disabled.
+// See https://github.com/att/ast/issues/820
+#if SUPPORT_JSON
 static_fn int json2sh(Shell_t *shp, Sfio_t *in, Sfio_t *out) {
     int c, state = 0, lastc = 0, level = 0, line = 1;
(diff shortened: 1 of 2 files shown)


krader1961:
If we're going to keep the code I'd prefer to introduce a build time flag like we did with `audit-file` in *meson_options.txt* and predicate the json stuff on that symbol. On the other hand I've changed my mind and decided that given that this feature is badly broken for reading and printing JSON and adds a lot of complexity to the "namval" core code I'd just as soon rip it out.

siteshwar:
@krader1961 I don't want to add another `meson` configuration option for it. If people want to enable `json` support, they can do it by defining `SUPPORT_JSON` macro while building.

src/cmd/ksh93/bltins/read.c
```diff
@@ -196,8 +199,13 @@ static_fn int json2sh(Shell_t *shp, Sfio_t *in, Sfio_t *out) {
     }
     return 0;
 }
+#endif
 
-static struct Method methods[] = {{"json", json2sh}, {"ksh", NULL}, {NULL, NULL}};
+static struct Method methods[] = {
+#if JSON_SUPPORT
```
krader1961: s/JSON_SUPPORT/SUPPORT_JSON/

krader1961:
LGTM modulo the one misspelling.

PR #1034, opened 2018-11-30
Extend tests for test (bracket) builtin

siteshwar:


src/cmd/ksh93/tests/test.exp
```diff
@@ -0,0 +1,29 @@
+# Test for echo builtin
```
krader1961: s/echo/`test`/

src/cmd/ksh93/tests/bracket.sh
```diff
@@ -454,6 +472,15 @@ x=10
...
 test 123 -eq 123x 2>/dev/null
 [[ $? -ge 2 ]] || log_error 'test builtin should return value greater than 1 on error'
+
+# ==========
+# Running test without any arguments should return 1
+test
+[[ $? -eq 1 ]] || log_error "test builtin should return 1 if expression is missing"
+
+# ==========
+test 1 -eq 2-1 || log_error "Arithmetic expressions should work inside test builtin"
```
krader1961: This is not POSIX compatible AFAICT. I can't see anything in the ksh man page which suggests this should evaluate to true (exit status zero). Yes, the `[[ 1 -eq 2-1 ]] || ...` formulation is documented to work this way. But not the `test` or `[` builtins. If we're validating this works to ensure backward compatibility with an undocumented feature there should be a comment to that effect. And since we're committing to support this behavior it would be nice if the documentation was updated to reflect the validity of this statement. Because both the BSD and GNU external `test` commands complain that `2-1` is not a valid number.

@siteshwar Please don't feel like I'm dumping on you. My frustration is with the old ksh team which implemented such extensions, without documenting them, that are not POSIX compatible, while simultaneously preaching how important it is to maintain POSIX compatibility.

Finally, if we're going to support this undocumented feature I would make this even more subtle and write it so both operands involved an arithmetic expression:

```
test 6/2 -eq 4-1 || log_error ....
```

PR #1035, opened 2018-11-30
Extend tests for `trap` builtin

siteshwar:


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -110,6 +110,37 @@ fi
 mkdir -p $TEST_DIR/a/b/c 2>/dev/null || log_error  "mkdir -p failed"
 $SHELL -c "cd $TEST_DIR/a/b; cd c" || log_error "initial script relative cd fails"
 
+#   trap - trap signals and conditions
+
+# ==========
+# -a append the current trap setting to the specified action.
+trap : SIGHUP
+trap -a : SIGHUP
```
krader1961: I would recommend using something like `true` rather than `:` to help distinguish this action from the previous trap action.  And perhaps using `false` for the first trap. That's because the order of the actions matters. Although the man page isn't entirely clear what the "append" order should be.

The man page says this: `The  -a  option  causes  the  current trap setting to be appended to action.` Which I would normally think of as prepending the action to the current trap setting and thus would have implemented as a `-p` flag. But apparently when this feature was added the `-p` flag to print a trap already existed.

PR #1042, opened 2018-12-03
Widen nvflag

krader1961:
This is another sequence of changes intended to make the namval code easier to understand and less likely to be a source of bugs.


siteshwar:
LGTM

Issue #1038, opened 2018-12-01
Changing size of `nvflag` in `struct Namval` breaks things


krader1961:
The discussion about support for JSON in issue #820 notes that symbols `NV_JSON` and `NV_JSON_LAST` are aliases for other symbols. Which is dangerous and shouldn't be necessary. So I opened PR #1037 to start work on eliminating such aliases. The next step in that cleanup is to change the definition of `struct Namval` member `nvflag` to be wider than an `unsigned short`. However turning it into a 32 bit unsigned int breaks unit tests. Even with no changes to the bit field symbols.

**That should be impossible.** That is, the size of that bit field should have zero effect on the behavior of the code as long as the size is wide enough to accommodate all bit values stored in that structure member. And converting from `unsigned short` (16 bits) to `uint32_t` (32 bits) is guaranteed to satisfy that constraint. This diff breaks 18 unit tests on macOS that pass without this change:

```
diff --git a/src/cmd/ksh93/include/nval.h b/src/cmd/ksh93/include/nval.h
index 66462dd9..ee293a5b 100644
--- a/src/cmd/ksh93/include/nval.h
+++ b/src/cmd/ksh93/include/nval.h
@@ -104,7 +104,7 @@ struct Namdecl {
 struct Namval {
     Dtlink_t nvlink;        // space for cdt links
     char *nvname;           // pointer to name of the node
-    unsigned short nvflag;  // attributes
+    unsigned int nvflag;  // attributes
 #if _ast_sizeof_pointer >= 8
     uint32_t nvsize;  // size or base
 #else
```

Something is seriously wrong with the code that allocates or uses a `struct Namval`. The aforementioned change should have zero effect on the behavior of the code other than the amount of memory used.

krader1961:
Running this on on macOS results in the "double-free" ASAN failure shown below:

```
meson -DASAN=true
ninja >x 2>&1
meson test --setup=asan -t4 exit
```

The `ast_realloc()` call and subsequent attempt to free the same storage a second time is highly suggestive of a miscalculation of the size of a namval structure. Which is a problem we've seen in other contexts.  And is why I opened issue #1002.

```
==26382==ERROR: AddressSanitizer: attempting double-free on 0x607000003350 in thread T0:
    #0 0x10451c1bd in wrap_free (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x551bd)
    #1 0x1042bd32b in ast_free vmbusy.c:50
[...]

PR #1045, opened 2018-12-04
Remove NV_PUBLIC

krader1961:
The definition of `NV_PUBLIC` is wrong. It includes two symbols,
`NV_NOSCOPE` and `NV_NOADD`, that can never appear in the `nvflag` struct
member. So it doesn't make any sense to include them given how `NV_PUBLIC`
is used.

It is a no-op for two of the three places it is used. For the third use
all it does is filter out `NV_NOFREE` from the flag so do that explicitly.

Looking at the ksh93u+ branch its use appears to be broken there as
well. So the code has been this way for quite a while.

Related #1038


siteshwar:
LGTM

Issue #954, opened 2018-10-14
`whence -a` is broken if builtin for a command is enabled


siteshwar:
**Description of problem:**
`whence -a` is broken if builtin for a command is enabled

**Ksh version:**
2017.0.0-devel-1945-gc7752ec6-dirty

**How reproducible:**
Always

**Steps to reproduce:**
```
whence -a cat
builtin cat
whence -a cat
```
**Actual results:**
```
cat is a shell builtin
cat is /usr/bin/cat
cat is a tracked alias for /home/situ/upstream/ast/build/cat
```

**Expected results:**
```
cat is a shell builtin
cat is /usr/bin/cat
cat is a tracked alias for /usr/bin/cat
```

**Additional info:**
This bug is present in `ksh-20120801`.

DavidMorano:
I reported this very bug a bazillion years ago, to the AST team, and they never did clean it up. I think that they had some other problem with |whence| (some worse problem) and they somehow forgot about this present bug reported above. I have learned to live with this particular bug (just ignoring it) for almost 20 years now.

krader1961:
Thank you, @siteshwar and @DavidMorano, for commenting on this. This is but one of many similar issues where the old AST team seemed to be more focused on implementing new behavior (which introduced new bugs) than fixing moderately serious bugs in the existing code.

Few projects are bug free. A lot of code I've written has bugs. But I've always prioritized fixing bugs likely to be seen over implementing new features. That the old AST team felt that ignoring significant bugs (as opposed to bugs unlikely to ever be noticed) was acceptable makes me a sad puppy.j

DavidMorano:
> ...  than fixing moderately serious bugs in the existing code.

I do not know exactly why this fell through the cracks (it did), but I
think initially it was felt that it was not a high priority as there was
some other |whence| related bug that they felt that had to be fixed first.
I forget what that other bug was now (it was so long ago). Then somehow,
this bug above was never handled.

brdjns:
> Thank you, @siteshwar and @DavidMorano, for commenting on this. This is but one of many similar issues where the old AST team seemed to be more focused on implementing new behavior (which introduced new bugs) than fixing moderately serious bugs in the existing code.
> 
> Few projects are bug free. A lot of code I've written has bugs. But I've always prioritized fixing bugs likely to be seen over implementing new features. That the old AST team felt that ignoring significant bugs (as opposed to bugs unlikely to ever be noticed) was acceptable makes me a sad puppy.j

[...]

PR #1049, opened 2018-12-05
`whence -a` should print correct path for tracked aliases

siteshwar:
`whence -a` should print path that a tracked alias refers to instead of
associating it with current directory.

This bug was reported to previous maintainers around 2 decades ago, but
it was never fixed. Original credits to @DavidMorano for finding and
reporting this bug.

Better late than never!

Resolves: #954

### DIFF
--- a/src/cmd/ksh93/bltins/whence.c
+++ b/src/cmd/ksh93/bltins/whence.c
@@ -283,8 +283,9 @@ static_fn int whence(Shell_t *shp, char **argv, int flags) {
                 if (flags & V_FLAG) {
                     if (*cp != '/') {
                         if (!np && (np = nv_search(name, shp->track_tree, 0))) {
+                            const char *command_path = np->nvalue.pathcomp->name;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/bltins/whence.c
```diff
@@ -283,8 +283,9 @@ static_fn int whence(Shell_t *shp, char **argv, int flags) {
                 if (flags & V_FLAG) {
                     if (*cp != '/') {
                         if (!np && (np = nv_search(name, shp->track_tree, 0))) {
+                            const char *command_path = ((struct pathcomp *)np->nvalue.cp)->name;
```
krader1961: Rather than the explicit cast just do `... = np->nvalue.pathcomp->name`.

P.S., I really, really, want to get rid of all these explicit casts. Especially in the case of `np->nvalue`. They're dangerous and hide bugs. And in the case of this struct member I want to add a corresponding enum that indicates the type of the last value assigned so that we can add asserts when using the value to verify the type that was stored is what is expected where it is used.

krader1961:
LGTM

Issue #1051, opened 2018-12-06
`cat` builtin tries to open non-existent files in a loop


siteshwar:
**Description of problem:**
`cat` builtin tries to open non-existent files in a loop

**Ksh version:**
2017.0.0-devel-2097-ga10c3a13

**How reproducible:**
Always

**Steps to reproduce:**
```
builtin cat
cat this_file_does_not_exit
```

**Actual results:**
`cat` tries to open non-existent file in a loop

**Expected results:**
`cat` should exit after the error

**Additional info:**
This is a regression introduced by e0b7783976ded76ca282e345f4cd11af657d402b

siteshwar:
Closed through #1054.

PR #1052, opened 2018-12-06
Fix a loop condition in `cat` builtin

siteshwar:
`cat` builtin goes in an infinite loop while trying to open a
non-existent file. This is a regression introduced by e0b778397.

Resolves: #1051

### DIFF
--- a/src/lib/libcmd/cat.c
+++ b/src/lib/libcmd/cat.c
@@ -489,8 +489,8 @@ int b_cat(int argc, char **argv, Shbltin_t *context) {
             }
         }
         if (sferror(sfstdout)) break;
-        cp = *argv++;
-    } while (cp);
+        ;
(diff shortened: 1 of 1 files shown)


src/lib/libcmd/cat.c
```diff
@@ -489,8 +489,8 @@ int b_cat(int argc, char **argv, Shbltin_t *context) {
             }
         }
         if (sferror(sfstdout)) break;
-        cp = *argv++;
-    } while (cp);
+        ;
+    } while ((cp = *argv++));
```
krader1961: The problem is that this block contains a `continue` that escaped my notice and how it is structured is weird. I've just created a PR that I think is cleaner. Also, the empty statement on the preceding line should be dropped even if this is merged.

siteshwar:
Closing in favor of #1054.

PR #1053, opened 2018-12-06
Extend tests for builtins for standard commands

siteshwar:
Enable tests for `--man` and unrecognized options for builtins for
standard commands.

### DIFF
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -19,14 +19,26 @@
 ########################################################################
 
 # ==========
-for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|\[|:)'); do
+for name in $(builtin -l | grep -Ev '(echo|test|true|false|login|newgrp|uname|getconf|\[|:)'); do
+    # Extract builtin name from /opt path
+    if [[ "$name" =~ "/opt" ]];
+    then
+        name="${name##*/}"
+    fi
+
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -19,14 +19,26 @@
 ########################################################################
 
 # ==========
-for name in $(builtin -l | grep -Ev '(echo|/opt|test|true|false|login|newgrp|\[|:)'); do
+for name in $(builtin -l | grep -Ev '(echo|test|true|false|login|newgrp|uname|getconf|\[|:)'); do
+    # Extract builtin name from /opt path
+    if [[ "$name" =~ "/opt" ]];
```
krader1961: I don't understand why you're doing this. If the builtin has been enabled it will appear in the `builtin -l` output under its bare name (e.g., `cat`) and its magic path (e.g., `/opt/bin/path`). But the magic path is never legal. For example, try running `/opt/ast/bin/dirname` before and after running `builtin dirname`.

krader1961: Ah yes, I had forgotten that I had enabled all builtins by default using the `PATH` mechanism precisely to ensure they get exercised.

krader1961:
LGTM

src/cmd/ksh93/tests/builtins.sh
krader1961: Although I'm not sure how I feel about them not showing up in the `builtin -l` output under their bare name. That seems wrong.

PR #1054, opened 2018-12-07
Fix builtin cat iteration over argv

krader1961:
My commit e0b7783 to resolve some lint errors introduced a bug due to
the odd structure of the code and because I overlooked the `continue`
within the `do...while()`. This fixes that mistake and transforms the
code to use a more typical, easier to understand, control structure.

### DIFF
--- a/src/lib/libcmd/cat.c
+++ b/src/lib/libcmd/cat.c
@@ -80,6 +80,8 @@ static const char usage[] =
 
     "[+SEE ALSO?\bcp\b(1), \bgetconf\b(1), \bpr\b(1)]";
 
+static char *stdin_argv[] = {"-", NULL};
+
 #define RUBOUT 0177
 
 /* control flags */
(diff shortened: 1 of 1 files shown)


src/lib/libcmd/cat.c
```diff
@@ -461,10 +463,9 @@ int b_cat(int argc, char **argv, Shbltin_t *context) {
         dovcat = 1;
     }
     if (flags & d_FLAG) sfopen(sfstdout, NULL, "wt");
-    cp = *argv;
-    if (cp) argv++;
-    do {
-        if (!cp || !strcmp(cp, "-")) {
+    if (!*argv) argv = stdin_argv;
```
siteshwar: Why do we need this assignment ?

siteshwar: Ok, I presume it's because we want to run this loop at least once.

krader1961: Precisely. I wouldn't necessarily write it this way in a different context. But given that `cat` by definition treats a single `-` argument as meaning "read from stdin" it's the simplest approach in this case. Although to be honest I'm inclined to do an even bigger refactoring because the body of this loop should be a separate function that is called for each `argv` element and is called once if argv is empty. But I felt it better to leave that refactoring for a subsequent change.

siteshwar:
LGTM

siteshwar:
Merging to continue my work on `cat` builtin tests.

PR #1058, opened 2018-12-08
Add tests for `basename` builtin

siteshwar:


src/cmd/ksh93/tests/basename.sh
```diff
@@ -0,0 +1,30 @@
...
+actual=$(basename "$TEST_DIR/foo.bar")
+expect="foo.bar"
+[[ "$actual" = "$expect" ]] || log_error "basename failed to print filename" "$expect" "$actual"
+
+# ==========
+# Last argument is treated as suffix and removed from output
+actual=$(basename "$TEST_DIR/foo.bar" .bar)
+expect="foo"
+[[ "$actual" = "$expect" ]] || log_error "basename failed to strip suffix" "$expect" "$actual"
+
```
krader1961: I would add another test to verify that removing a suffix is a no-op if it would elide the filename. That is, `basename /a/.c .c` should result in `.c` not the empty string.

krader1961:
LGTM

src/cmd/ksh93/tests/basename.sh
siteshwar: Added.

PR #1064, opened 2018-12-09
Enable auditing in travis builds

siteshwar:


krader1961:
LGTM but this begs the question of how to verify that the expected entries are written to the audit file.

siteshwar:
Yes, at some point we will have to write a separate test to verify that entries are not corrupted.

PR #1065, opened 2018-12-09
Add tests for `dirname` builtin

siteshwar:


krader1961:
LGTM as a starting point but these tests only do the most cursory validation of the options supported by the command. And from the help text I can't actually figure out what these options are supposed to do.

Issue #1066, opened 2018-12-10
lib/package directory needs cleanup


krader1961:
I was greping for a symbol and noticed it matched a file in the *lib/package/tgz* directory. The contents of the *lib/package* directory is now badly out of date with the current state of the project. It's not clear if the entire directory should be eliminated or its files updated to reflect the current state of the project. But at a minimum the files in *lib/package/tgz* should be removed since they are no longer valid.

krader1961:
Commit #1067 removed the prepackaged source archives that have not been relevant for a long time. AFAICT the rest of the *lib/package* directory is no longer relevant and should be removed. With the possible exception of the *lib/package/LICENCES* directory in which case those files should be moved to a different directory and/or the licenses of this code clarified.

PR #1067, opened 2018-12-10
Remove out of date source archives

krader1961:
The files being removed are no longer relevant. They don't reflect the
current state of the project (e.g., we no longer use the legacy Nmake
build system) so remove them.

Related #1066


siteshwar:
LGTM

Issue #11, opened 2016-04-27
ksh regression: with FIGNORE, . and .. are no longer automatically excluded from glob expansions


stephane-chazelas:
From the manual:

> If FIGNORE  is set, then each file name component that matches the pattern defined by the value of FIGNORE is ignored when generating the matching filenames.   **The  names .  and ..  are also ignored.**

That used to be true. As in, that works as documented in ksh93k+ for instance, but in modern versions (and also in ksh93m, I couldn't find ksh93l to test) that doesn't:

```
$ FIGNORE=x ksh93v- -c 'echo *'
. .. a .a
$ FIGNORE=x ksh93k+ -c 'echo *'
a .a
```

siteshwar:
I would like to see how it worked in `ksh93k+`. Are the older sources for ksh available to download from somewhere ? The oldest I could find from our repositories is `ksh93q` in RHEL-3.

dannyweldon:
I found some old archives here for ksh93o+ (ast-ksh.2003-07-24.tgz):

http://ftp.vim.org/ftp/ftp/pub/ibiblio/distributions/slack390/slack390-9.1/source/ap/ksh93/

I'm not sure when the source was originally available publicly.

krader1961:
This "feature" is an example of the sort of enhancement to zsh that caused me to abandon it as my primary shell. That is, a feature that was added even though it isn't a good fit for the language and is all but guaranteed to cause problems out of proportion to any benefit it provides solely because someone thought it was a "cool idea". But if we're going to retain it I would prefer that we not revert to the documented behavior that hasn't existed for 16 years. It would be better to document the feature more clearly. In particular showing how the user can exclude the `.` and `..` directory entries via patterns like `FIGNORE='+(.|..|x*)'`.

Please note that if you don't define `FIGNORE` then something like `echo *` does not include file names that begin with a `.`. So why in the world should defining that variable only exclude `.` and `..` but not other names that begin with a `.`?

siteshwar:
Ignoring `.` and `..` seems to  be for sanity purpose. For e.g. consider the case when you have set `FIGNORE="*.swp"` and execute `rm -rf *` in a directory. That would end up deleting parent directory too.

krader1961:
> Ignoring . and .. seems to be for sanity purpose....

Without a `FIGNORE` var all "dot" file names are ignored by the expansion of `*`. The problem as I see it is that this only special cases `.` and `..` while otherwise passing through other paths that begin with a `.`.

[...]

PR #1068, opened 2018-12-11
 Ignore `.` and `..` in globbing if `FIGNORE` is set

siteshwar:
Resolves: #11


krader1961:
This has been broken since at least 2002-03-26 (the ksh93m+ release date).  That's 16 years. Before we "fix" this (scare quotes are intentional) I think we need a better understanding of the intent of this "feature" and whether we should change its behavior after 16 years.

Issue #1069, opened 2018-12-11
`echo` builtin does not print to stdout


siteshwar:
**Description of problem:**
`echo` builtin does not print to stdout

**Ksh version:**
2017.0.0-devel-2109-g615d0b8d

**How reproducible:**
Always

**Steps to reproduce:**
Run this script:

```
#!/bin/ksh

#set -x

max=0;
max_x=;
max_y=;
max_z=;


s=1718

for k in {1..300}; do 
        for l in {1..300}; do
                sum=0
                size=0
                [ "$k" -gt "$size" ] && size="$k"
                [ "$l" -gt "$size" ] && size="$l"
                size=$((300-size))
                for ((z=size;z>=0;z--)); do
                        /bin/echo $k $l $z        
                        for ((i=0;i<=z;i++)); do 
                                for ((j=0;j<=z;j++)); do
                                        nx=$((k+i))
                                        ny=$((l+j))
                                        ((sum+=((((nx+10)*ny+s)*(nx+10))%1000)/100-5))
                                done
                        done
                        if [ "$sum" -gt "$max" ]; then
                                max=$sum
                                max_x=$k
                                max_y=$l
                                max_z=$z
                        fi
                done
        done
done 

echo $max $max_x $max_y $max_z
```

**Actual results:**
`echo` does not print anything to stdout

**Expected results:**
`echo` should print:

```
1 1 299
1 1 298
....
```

**Additional info:**
This issue is present in `ksh93u+` release. Thanks @lnykryn for the reproducer script.

PR #1070, opened 2018-12-11
Sync all streams after executing a builtin

siteshwar:
After executing a builtin, all streams should be synced to avoid any
delay in output.

Resolves: #1069


krader1961:
LGTM

Issue #9, opened 2016-04-26
<>; redirection operator doesn't work for the last command of a `ksh -c inline-script`


stephane-chazelas:
$ echo test > a; ksh -c 'echo x 1<>; a'; cat a
    x
    st

It's OK, if we insert a command after that:

    $ echo test > a; ksh -c 'echo x 1<>; a; exit'; cat a
    x

siteshwar:
@stephane-chazelas Sorry for taking so long to respond to this issue. 

This is what other shells do:

bash (devel branch):

```
echo test > a; ./bash -c 'echo x 1<> a'; cat a
x
st
```


zsh (zsh 5.5.1):

```
> echo test > a; zsh -c 'echo x 1<> a'; cat a
x
st
```

So this behavior is compatible with other shells. `>` operator causes a file to be overwritten, but `<>` does not overwrite a file. So this behavior seems expected. 


What's more interesting is `ksh` behavior changes if I use `1<>;` instead of `1<>` with `exit` command:

```
echo test > a; ksh -c 'echo x 1<> a; exit'; cat a
x
st
```


```
echo test > a; ksh -c 'echo x 1<>; a; exit'; cat a
x
```

siteshwar:
Ok, I see how `<>` is different from `<>;` in man page now:


```
 <>;word       The same as <>word except that if the command completes successfully, word is truncated to the offset at command completion.  <>;word cannot be used with the exec(2).  built-in.
```

So this is a bug.

krader1961:
Yes, this is technically a bug. But it has probably been broken since it was introduced on 2009-01-23 for the ksh93t+ release. From the *lib/package/ast-open.README* file:

> 09-01-23 +The redirection operator <>; was added.  It is similar to <> except that if the command it is applied to succeeds, the file is truncated to the offset at the command completion.

It's hard to know if it was broken in the ksh93t+ release that introduced this feature. Or by a subsequent change in the ksh93u+ release. I can't readily test the former version but it is broken in the most recent public release. The fact this idiom is used in the *src/cmd/ksh93/tests/io.sh* unit test, and that test passes, suggests this feature has never been unit tested. Too, I don't understand how this would be meaningfully used in a ksh script. Yes, I have written lots of code that opens a file, updates its content, and truncates it to a specific length. But I have never expected to be able to do this in a shell script.

What is the probability anyone is using this feature? It seems to me to be zero or so close to zero that we don't care.

[...]

PR #1073, opened 2018-12-12
Avoid using non-forking subshells with `-c` option

siteshwar:
Resolves: #9


krader1961:
LGTM and I agree that disabling the problematic no-fork behavior for one-liner `ksh -c` invocations is a reasonable thing to do.

krader1961:
For the record I don't understand what the code being removed was intended to do. It seems to me the fact the statements were specified via the `-c` flag rather than in a file (or stdin) should be completely irrelevant as to whether non-forking behavior was in effect. The usual logic for executing a "pipeline" of statements should handle whether non-forking behavior is used regardless of how the pipeline was given to the shell. Especially since the `-c` arg may be arbitrarily complex.

In other words, to state the blindingly obvious, whether the `echo x 1<>; a` is in a shell script or run via `ksh -c` should not matter.

PR #1078, opened 2018-12-15
Hash enum

krader1961:
This PR eliminates the legacy `HASH_` symbols that no longer has any obvious meaning since the legacy hash subsystem has been replaced by the CDT subsystem. What I don't understand is why this wasn't done by the AST team long ago.


krader1961:
Perhaps the most significant aspect of this PR is to eliminate the need for several explicit `char*` casts. Explicit typecasts of that sort are problematic and should always be avoided if at all possible. When that isn't possible the use of `void*` as the function parameter type is preferable to an explicit cast by the caller.

siteshwar:
LGTM

Issue #951, opened 2018-10-13
builtin `cat` broken


krader1961:
When I enabled all builtins by default for unit tests, see #944, it broke the *src/cmd/ksh93/test/basic.sh* and *subshell.sh* unit test scripts. Specifically statements like this that involve reading */dev/fd/3* causes the tests to timeout:

```
if cat /dev/fd/3 >/dev/null 2>&1
```

Those have to be modified to run the external `cat` command to succeed. Why that is the case is unknown at this time.

siteshwar:
If `cat` is executed as an external command `if cat /dev/fd/3 >/dev/null 2>&1` is never true. So the tests under this condition are never executed. If I enable `cat` builtin and commet out [this call](https://github.com/att/ast/blob/f0ed7a93f75028c5237f7d5356dc6f4262b58241/src/cmd/ksh93/tests/basic.sh#L455) to `empty_fifos` all tests pass.

Also, the tests that fail rely on time differences:


```
s=SECONDS
set -o pipefail
for ((i=0; i < 30; i++))
do
    print hello
    sleep .1
done | $bin_sleep 1
(( (SECONDS-s) < 2 )) || log_error 'early termination not causing broken pipe'
```

```
set -o pipefail
float start=$SECONDS end
for ((i=0; i < 2; i++))
do
    print foo
    sleep 1.5
done | { read; $bin_true; end=$SECONDS ;}
(( (SECONDS-start) < 1 )) && log_error "pipefail not waiting for pipe to finish"
set +o pipefail
(( (SECONDS-end) > 2 )) &&  log_error "pipefail causing $bin_true to wait for other end of pipe"
```

They should be rewritten to test exit status instead.

krader1961:
> If cat is executed as an external command if cat /dev/fd/3 >/dev/null 2>&1 is never true. 

Ah, yes. I was thinking the `exec` would create a file descriptor that would remain open across an exec of an external command. But I see from looking at the code these are marked close-on-exec.

> Also, failing tests rely on time differences

Sadly, not true. I took a closer look and the pipefail isn't happening. It's not just a timing race. The difference in behavior is due to the producer/consumer test beginning near line 433. Note that the test in question does not use the `cat` command and therefore has no business being predicated on whether the builtin `cat` is enabled.

PR #1083, opened 2018-12-18
Subshell isatty

krader1961:


krader1961:
Note that I make no claims that this fixes problems involving stdin, stderr, or any other file descriptor inside a subshell.

siteshwar:
I see these tests failing under FreeBSD in CI:

```
    default: 100/231 basic                                   FAIL    44.77 s (exit status 4)
    default: 101/231 basic/shcomp                            FAIL    44.77 s (exit status 4)
```

@krader1961 Is it passing for you on FreeBSD ?

krader1961:
The aforementioned core dump is because `sh_open()` in the block of code only used on systems without */dev/fd* is failing and returning -1. TBD is why that is.

krader1961:
The FreeBSD failure is not due to this PR. It's due to my commits for issue #951 two days ago (specifically commit 857b2f92). It turns out that if your system doesn't support /dev/fd/_N_ for _N_ > 2 then you get a single magic fifo used for process substitution. The tests that are failing need more than one. So I'll fix this by again guarding the problematic tests on whether full /dev/fd support is available (albeit in a cleaner manner).

krader1961:
Argh! It looks like I broke systems without a fully functional /dev/fd with commit 4c65882b. But if that is true it begs the question why those particular tests are guarded by a check for whether the system fully supports /dev/fd paths. It looks like that is because version ksh93u+ fails those tests. In other words the reason for those tests to be predicated on full /dev/fd support is because ksh93u+ fails those tests on systems without full /dev/fd support.

Issue #70, opened 2017-09-30
ksh93: ${var+set} does not work within loops


McDutchie:
In the current release version of ksh93 as well as the beta, expansions like `${var+set}` (yielding 'set' if the variable is set) do not work correctly when used within a `for`, `while` or `until` loop.

Test script 1:

    unset -v var
    for i in 1 2 3 4 5; do
            case ${var+s} in
            ( s )   echo set; unset -v var;;
            ( '' )  echo unset; var=;;
            esac
    done

Actual output:

    unset
    unset
    unset
    unset
    unset

Expected output:

    unset
    set
    unset
    set
    unset

Test script 2:

    var=
    for i in 1 2 3 4 5; do
            case ${var+s} in
            ( s )   echo set; unset -v var;;
            ( '' )  echo unset; var=;;
            esac
    done

Actual output:

    set
    set
    set
    set
    set

Expected output:

    set
    unset
    set
    unset
    set

It looks like the expansion remains static within the loop, though it should change with the variable's set/unset status.

Bizarrely, this bug only applies to ${var+set}... something like ${var-unset}, ${var:-empty} or ${var:+nonempty} works fine:

    unset -v var
    for i in 1 2 3 4 5; do
            case ${var-u} in
            ( x )   echo set; unset -v var;;
            ( u )   echo unset; var=x;;
            esac
    done

Output (OK):

    unset
    set
    unset
    set
    unset

McDutchie:
FYI, it is an effective workaround to use a shell function that tests if the variable is set, even if that function uses the same `${foo+s}` sort of parameter expansion.

PR #1087, opened 2018-12-21
Fix `${+}` parameter expansion inside loops

siteshwar:
Resolves: #70

### DIFF
--- a/src/cmd/ksh93/tests/modifiers.sh
+++ b/src/cmd/ksh93/tests/modifiers.sh
@@ -94,3 +94,22 @@ do
         fi
     done
 done
+
+# ==========
+# https://github.com/att/ast/issues/70
+cat > $TEST_DIR/modifier-in-loop.sh <<EOF
+unset -v var
+for i in 1 2 3 4 5; do
+        case \${var+s} in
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/tests/modifiers.sh
```diff
@@ -94,3 +94,22 @@ do
         fi
     done
 done
+
+# ==========
+# https://github.com/att/ast/issues/70
+cat > $TEST_DIR/modifier-in-loop.sh <<EOF
```
krader1961: FWIW, This scenario is why I introduced the concept of unit test sub directories. For example, *src/cmd/ksh93/tests/signal/*. Whether or not we need a different mechanism for organizing tests within a unit test module is TBD. Don't refactor this as it is okay and consistent with 99.9% of the test code. I'm just commenting to get people to think about what we should do to improve the ksh unit test framework.

krader1961:
Drop the last commit, 6b6a386, from the PR as I've already merged a similar change but with more comments to remind myself (or someone) to take a closer look at why that test, and at least one other in the `types.sh` unit test, is so sensitive to the value of `STK_FSIZE`. Other than that...

LGTM

Issue #911, opened 2018-09-24
`set -b`does not show message on job completion


siteshwar:
**Description of problem:**
According to man page entry of `set -b`:

```
-b      Prints job completion messages as soon as a background job changes state rather than waiting for the next prompt.
```

**Ksh version:**
2017.0.0-devel-1846-g0296704d

**How reproducible:**
Always

**Steps to reproduce:**
```
set -b
sleep 3&
```

and wait on the prompt for completion message to appear.

**Actual results:**
Completion message is printed only at next prompt.

**Expected results:**
Completion message should be printed as soon as the job exits.

**Additional info:**
This bug is present in `93u+ 2012-08-01` release.

krader1961:
For the record this is broken in ksh93u+ and ksh93v- as well as the current development branch. So it's been broken at least six years.

PR #1089, opened 2018-12-22
`set -b` should immediately notify job state changes

siteshwar:
Restore last `SH_TTYWAIT` state instead of turning it off while handling
job state changes. Keeping it off without check breaks `set -b` because
job state changes are only shown if `SH_TTYWAIT` is on.

Resolves: #911

### DIFF
--- a/src/cmd/ksh93/sh/jobs.c
+++ b/src/cmd/ksh93/sh/jobs.c
@@ -408,9 +408,13 @@ bool job_reap(int sig) {
         flags = WUNTRACED | wcontinued;
     }
     oerrno = errno;
+
+    // Save tty wait state
+    int was_ttywait_on = sh_isstate(shp, SH_TTYWAIT);
+
     while (1) {
         if (!(flags & WNOHANG) && !shp->intrap && job.pwlist) {
-            sh_onstate(shp, SH_TTYWAIT);
+            if (!was_ttywait_on) sh_onstate(shp, SH_TTYWAIT);
         }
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/jobs.c
```diff
@@ -408,9 +408,13 @@ bool job_reap(int sig) {
         flags = WUNTRACED | wcontinued;
     }
     oerrno = errno;
+
     while (1) {
+        // Save tty wait state
+        int was_ttywait_on = sh_isstate(shp, SH_TTYWAIT);
```
krader1961: This should be hoisted outside the loop.

src/cmd/ksh93/sh/jobs.c
```diff
@@ -434,7 +438,11 @@ bool job_reap(int sig) {
         }
 #endif  // SHOPT_COSHELL
         pid = waitpid((pid_t)-1, &wstat, flags);
-        sh_offstate(shp, SH_TTYWAIT);
+        if (was_ttywait_on) {
```
krader1961: If it was already on there isn't any need to enable it again, is there? Shouldn't this just be

```
if (!was_ttywait_on) sh_offstate(shp, SH_TTYWAIT);
```

krader1961: Also, there is a `break` inside the loop so it seems like the same statement is needed around line 580 when the loop exits.

krader1961:
LGTM

Issue #1096, opened 2019-01-02
Compilation fails on NetBSD -current aarch64 (evbarm) 


DoktorCranium:
Trying to build the ksh93 on NetBSD aarch64 since the pkgsrc/shells/ast-ksh is broken. Getting the following error during ninja build 

[root@arm64 ast]# ninja -C build
ninja: Entering directory `build'
[2/3] Linking target src/cmd/ksh93/shcomp.
FAILED: src/cmd/ksh93/shcomp 
clang  -o src/cmd/ksh93/shcomp 'src/cmd/ksh93/src@cmd@ksh93@@shcomp@exe/sh_shcomp.c.o' -Wl,--as-needed -Wl,--export-dynamic -Wl,--start-group src/cmd/ksh93/libksh.a src/lib/libast/libast.a src/lib/libcmd/libcmd.a src/lib/libdll/libdll.a -lm -lexecinfo -lm -lexecinfo -lm -lexecinfo -Wl,--end-group  
src/cmd/ksh93/libksh.a(bltins_math.c.o): In function `local_nexttoward':
/mnt/usb/ast/build/../src/cmd/ksh93/bltins/math.c:109: undefined reference to `nexttowardl'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x190): undefined reference to `expm1l'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x478): undefined reference to `lgammal'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x4c0): undefined reference to `log1pl'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x4d8): undefined reference to `log2l'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x568): undefined reference to `remainderl'
src/cmd/ksh93/libksh.a(bltins_math.c.o):(.data.rel.ro+0x658): undefined reference to `tgammal'
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)
[3/3] Linking target src/cmd/ksh93/ksh.
FAILED: src/cmd/ksh93/ksh 
[...]

PR #1097, opened 2019-01-02
Add feature test for expm1l, log1pl and remainderl

siteshwar:
These functions are not available on NetBSD. Add a feature test for them
and fallback to their respective counterparts that don't accept long
arguments if these functions are not available.

Resolves: #1096


krader1961:
This is incomplete. For example, `nexttowardl` is also not available on NetBSD. It is also incorrect because the functions have to accept and return `Sfdouble_t` (i.e., `long double`) unless you change the signature encoded in the first byte of the string containing the function name.

PR #1098, opened 2019-01-04
Add feature test for several math functions

siteshwar:
These functions are not available on NetBSD. Add a feature test for them
and fallback to their respective counterparts that don't accept long
arguments if these functions are not available.

### DIFF
--- a/src/cmd/ksh93/bltins/math.c
+++ b/src/cmd/ksh93/bltins/math.c
@@ -100,13 +100,17 @@ static Sfdouble_t local_nexttoward(int type_1, Sfdouble_t arg_1, int type_2, Sfd
 
     switch (type_1) {
         case 1: {
-            return nexttowardf((float)arg_1, arg_2);
+            return nexttowardf(arg_1, arg_2);
         }
         case 2: {
-            return nexttoward((double)arg_1, arg_2);
+            return nexttoward(arg_1, arg_2);
         }
         case 3: {
+#if _lib_nexttowardl
             return nexttowardl(arg_1, arg_2);
+#else
+            return nexttoward(arg_1, arg_2);
+#endif
         }
         default: { return 0; }
     }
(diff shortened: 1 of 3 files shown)


siteshwar:
@krader1961 You have put this comment in my previous pull request:

```
It is also incorrect because the functions have to accept and return Sfdouble_t (i.e., long double) unless you change the signature encoded in the first byte of the string containing the function name.
```

Which signature you are referring to ? If user passes a number that's out of bound for `double` type, it's anyways going to create issues, no matter what the function signature is. Did I miss something here ?

siteshwar:
@krader1961 fwiw if we can't get these math functions to work reliably on NetBSD, we can disable it there.

siteshwar:
I realized math tests are failing on NetBSD with these changes. If changes required to make it work on NetBSD can introduce regressions, I will just disable these functions.

src/cmd/ksh93/bltins/math.c
```diff
@@ -147,7 +151,11 @@ const struct mathtab shtab_math[] = {{"\001acos", (Math_f)acosl},
                                      {"\001erfc", (Math_f)erfcl},
                                      {"\001exp", (Math_f)expl},
                                      {"\001exp2", (Math_f)exp2l},
+#if _lib_expm1l
                                      {"\001expm1", (Math_f)expm1l},
+#else
+                                     {"\001expm1", (Math_f)expm1},
```
krader1961: This isn't legal. The code in *arith.c* will pass a single argument of type `long double` to `expm1`. The only way this would work is if `long double` is the same as `double` on NetBSD. Which the C standard allows but I doubt it is true given the NetBSD has long double variants for the other functions. You have to create a wrapper function; e.g., as was done for `j0` via `local_j0` since `j0` takes a `double` not a `long double`.

siteshwar: Fixed.

src/cmd/ksh93/bltins/math.c
```diff
@@ -100,13 +100,17 @@ static Sfdouble_t local_nexttoward(int type_1, Sfdouble_t arg_1, int type_2, Sfd
 
     switch (type_1) {
         case 1: {
-            return nexttowardf((float)arg_1, arg_2);
+            return nexttowardf((float)arg_1, (float)arg_2);
```
krader1961: None of these explicit casts are required. And while you can argue they provide useful information to a reader of the code they can also cause bugs. What if `nexttowardf()` actually took a different type such as a `double`. This will compile without warning but will cause two conversions and lost of information.

src/cmd/ksh93/bltins/math.c
```diff
@@ -178,17 +212,33 @@ const struct mathtab shtab_math[] = {{"\001acos", (Math_f)acosl},
                                      {"\001j1", (Math_f)local_j1},
                                      {"\002jn", (Math_f)local_jn},
                                      {"\042ldexp", (Math_f)ldexpl},
+#if _lib_lgammal
```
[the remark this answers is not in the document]
siteshwar: Instead of this. I did:

```
#define expm1l local_expm1
```

Any objection to this style ?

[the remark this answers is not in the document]
krader1961: It's a coin toss. It's not clear if your approach or mine is more consistent with how the other non-standard functions (e.g., `j0`) are handled. Your version is as justifiable as mine.

krader1961:
LGTM

PR #1101, opened 2019-01-06
Coverity: Logically dead code

krader1961:
Eliminate some dead code; i.e., code which can never be executed. Also
restructure the enclosing loop to make it easier to read by eliminating
assignments inside a control structure.

Coverity CID#253646


siteshwar:
LGTM

PR #1102, opened 2019-01-06
Eliminate sfpoll()

krader1961:
Coverity warned about logically dead code in the `sfpoll(()` function.
The use of var `c` in that function is definitely FUBAR. Rather than try
to fix it just remove the function since it isn't used by ksh.

Coverity CID#253729


siteshwar:
LGTM

Issue #817, opened 2018-08-29
`read -u` prints input line on prompt


siteshwar:
**Description of problem:**
`read -u` prints input line on next prompt

**Ksh version:**
2017.0.0-devel-1560-g92d2d6d1

**How reproducible:**
Always

**Steps to reproduce:**

Execute these lines:
```
echo "Sample Text" > /tmp/foo
exec 3< /tmp/foo
read -u3 foo
```
**Actual results:**
`Sample Text` is printed on next prompt.

**Expected results:**
Prompt should be empty.

**Additional info:**
This bug is present in `93v- 2014-12-24` release.

PR #1103, opened 2019-01-07
`read -u` should not set default value for terminal input

siteshwar:
Remove an invalid fall through in `read` builtin.

Resolves: #817


src/cmd/ksh93/bltins/read.c
```diff
@@ -318,8 +318,8 @@ int b_read(int argc, char *argv[], Shbltin_t *context) {
                             (shp->gd->hist_ptr && fd == sffileno(shp->gd->hist_ptr->histfp)))) {
                     break;
                 }
+                break;
             }
-            // FALLTHRU
```
krader1961: And here we have a good example why fall thru by default is a bad idea :smile:

krader1961:
LGTM

PR #1104, opened 2019-01-07
Cleanup fifo in poll feature test at exit

siteshwar:
This feature test creates fifo file in current working directory.
Cleanup it up at exit.

### DIFF
--- a/features/poll.c
+++ b/features/poll.c
@@ -16,8 +16,15 @@ void sig_handler(int sig) {
     exit(1);
 }
 
+void cleanup_fifo() {
+    unlink(FIFO_FNAME);
+}
+
 int main() {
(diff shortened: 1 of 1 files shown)


siteshwar:
Ideally meson should run programs compiled through `compiler.run()` with `MESON_BUILD_ROOT` as current working directory. I will open an issue with meson upstream to follow up on it.

krader1961:
> Ideally meson should run programs compiled through compiler.run() with MESON_BUILD_ROOT as current working directory.

Agreed. I didn't notice this because I always do my builds with *build* as the CWD so I can just do `meson && ninja` without having to specify the build dir name.

krader1961:
LGTM

PR #1106, opened 2019-01-07
Disable support for `json` in `print` builtin

siteshwar:
Resolves: #820


src/cmd/ksh93/tests/comvar.sh
```diff
@@ -689,10 +689,10 @@ c.c=()
...
 [[ $(print -v c) == "$exp" ]] || log_error 'setting compound array c.c=() does not preserve -C attribute'
 
-compound xx=(this=that integer x=5)
-exp=$'{\n\t"this": "that",\n\t"x": 5\n}'
-[[ $(print -j xx) == "$exp" ]] || log_error "got $(print -j xx)" expected "$exp"
-
-compound y=(foo=bar;compound x=(lef=one right=2);integer z=5)
-exp=$'{\n\t"foo": "bar",\n\t"x": {\n\t\t"lef": "one",\n\t\t"right": "2"\n\t},\n\t"z": 5\n}'
-[[ $(print -j y) == "$exp" ]] || log_error "got $(print -j y)" expected "$exp"
+# compound xx=(this=that integer x=5)
```
krader1961: Probably should have a "TODO" comment explaining why this is disabled and linking to a relevant issue.

krader1961:
LGTM

kshji:
I'm not happy about disable support for json.... 
I documented some bugs for David 2014. I also told that there is working set of  C json parser to make working version: https://github.com/krz8/jsoncvt. 

Of course you can use ex. jq with eval set to build same set, but it's so much nicer to use compound variables : read and print using builtin properties.

Same way as builtin csv support is very usable: read -S to read csv and printf "%(csv)"
```sh
typeset -a cols
while IFS=";" read -A -S cols
do ...
done < x.csv
```

krader1961:
@kshji We'd love to see people like you who care about such things contribute fixes to the code. The problem is that there are a lot of bugs even with JSON support disabled. And at the moment there are only two people (@siteshwar and myself) contributing fixes. We simply cannot afford at this time to enable a new feature that itself has bugs independent of all the other bugs that are present. We have to fix the more egregious (e.g., use-after-free) bugs and get distros to adopt a new release based on this project before we can add new features like support for JSON.

Issue #1107, opened 2019-01-07
Legacy feature test script does not recognize `CC` environment variable


ryandesign:
**Description of problem:**

Hello, the last time I built ksh from master on 20181014 it built fine, but trying to build the latest code as of 20190106 fails.

**Ksh version:**

20190106

**How reproducible:**

Always

**Steps to reproduce:**
1. Compile ksh on macOS High Sierra 10.13.6

**Actual results:**

```
[3/476] Compiling C object 'src/lib/libast/e416224@@ast@sta/conftab.c.o'.
FAILED: src/lib/libast/e416224@@ast@sta/conftab.c.o 
[...]

PR #1109, opened 2019-01-08
Use compiler set in `CC` environment variable

siteshwar:
Use `CC` environment variable while doing legacy feature tests.

Resolves: #1107


siteshwar:
@ryandesign Let us know if it works for you.

PR #1113, opened 2019-01-10
Lint: iousepipe() return value is a constant

krader1961:
Both clang static analysis and Coverity Scan reported that the value
assigned to `unpipe` for the `if (shp->subshell) {` case is unused when
building when `SHOPT_SPAWN` is false. There are only two callers of
`iousepipe()` and one of them ignores the return value which makes sense
since the return value is a constant -- once you realize that `sh_rpipe()`
can't fail. For the call site that does use the return value just set the
`unpipe` to true only if `SHOPT_SPAWN` is true to mirror where the value of
`unpipe` is used.

Note that while `sh_rpipe()` can fail it cannot return a failure value.
For good or bad it always returns success unless there is a problem in
which case it longjmp's via the call to `errormsg()`.

This still leaves some cruft since `unpipe` is only needed at all if
`SHOPT_SPAWN` is true. But removing that cruft requires a much larger
refactoring.

### DIFF
--- a/src/cmd/ksh93/include/io.h
+++ b/src/cmd/ksh93/include/io.h
@@ -56,7 +56,7 @@ extern void sh_ioinit(Shell_t *);
 extern int sh_iomovefd(Shell_t *, int);
 extern int sh_iorenumber(Shell_t *, int, int);
 extern void sh_pclose(int[]);
-extern int sh_rpipe(int[]);
+extern void sh_rpipe(int[]);
 extern void sh_iorestore(Shell_t *, int, int);
 #if !_AST_no_spawnveg
 extern void sh_vexrestore(Shell_t *, int);
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -740,7 +740,7 @@ int sh_pipe(int pv[]) {
 //
 // Create a real pipe when pipe() is socketpair.
 //
-int sh_rpipe(int pv[]) {
+void sh_rpipe(int pv[]) {
     Shell_t *shp = sh_getinterp();
     int fd[2];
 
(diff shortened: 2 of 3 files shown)


siteshwar:
Doesn't this change break this code ?https://github.com/att/ast/blob/bd3d8688b08ec2facf1c88d1ae080db72e0194c0/src/cmd/ksh93/sh/xec.c#L1596 ?

krader1961:
> Doesn't this change break this code ?

No, because that code is already broken. It is impossible for `unpipe` to be non-zero when that statement is executed. Which is why the clang static analysis pointed out that the assignment in this block is unused if `SHOPT_SPAWN` is false:

https://github.com/att/ast/blob/master/src/cmd/ksh93/sh/xec.c#L1318-L1325

I've confirmed this both by code inspection and by instrumenting the code rather than just trusting the compiler. I'll replace that statement with an `assert()` but this begs the question of why that statement is even present. Was it possible for the condition to be true in the past or has it always been a no-op? Does the presence of that line imply there is a bug rather than someone neglecting to remove a dead block of code when some prior change was made?

krader1961:
Face palm /s

There are only five places that can result in `comsub` being anything other than zero or one in the blocks of code being discussed with respect to this pull-request. And they're all in *src/cmd/ksh93/sh/macro.c* via the `comsubst(...)` calls. This is a poster child example for why magic constants without meaningful names are a really bad idea. Not to mention using a data type that makes aliasing errors likely. I would fire the person who wrote this code if they wrote such code after being told why it is unacceptable.

krader1961:
There is no way to know now why the use of `unpipe` in the `case TSETIO:` branch is present. Both static analysis and empirical testing shows that it cannot be, and is never, true in that code path. It looks like the definition of `unpipe` should be limited to the scope of the `case TFORK:` block of code so I've moved it there and eliminated the reference in the `case TSETIO:` branch.

It also looks like at least one of the `// FALLTHRU` cases is actually impossible (I'm going to investigate those cases). Which illustrates why the `sh_exec()` code is a white hot mess that needs to be refactored.

P.S., There are a huge number of vars local to a function that should be defined in a more specific block of code. Too much of this code still assumes the semantics and restrictions of the original K&R C specification from four decades ago.

PR #1114, opened 2019-01-13
Refactor sh_exec()

krader1961:
While investigating a different issue I realized that `case TFORK:` cannot
fall through to `case TSETIO:`. So refactor the code to make that clear.


krader1961:
The diff as shown by github obscures the nature of the change. Which was to convert this:

```
if (job.parent) {
    // parent branch
} else {
   // child branch
}
```

Into this:

```
if (!job.parent) {
    forked_child(...);
    __builtin_unreachable();
}
// parent branch
```

siteshwar:
LGTM

PR #1119, opened 2019-01-16
`jobs` builtin should print correct working directory

siteshwar:


krader1961:
LGTM although I don't know that replacing `wd:` with the more verbose `Working directory:` is necessary. How about `pwd:`?

siteshwar:
Fixed.

Issue #1116, opened 2019-01-14
Analyze changes made after last stable release (ksh93u+)


siteshwar:
`ksh93u+` was last stable release that was embraced by various distributions. We should analyze all the changes made after this release. Any backward incompatible changes should be reverted. If there is a new feature that's worth mentioning, it should be mentioned in `CHANGELOG.md` file. Changes can be found [here](https://github.com/att/ast/blob/5cedc989a424ddcb54ba5c2c19c9c36cbeffbefd/src/cmd/ksh93/RELEASE#L1-L655).

krader1961:
FWIW, Any introduction of a new builtin is strictly speaking not backward compatible since the new builtin may have different behavior than the platform's external command of the same name. Which is why @siteshwar removed a few new builtins such as `date`. See issue #990 where I raised the question whether the `tmxdate()` function should be removed or the builtin `date` reinstated in order to justify the cost of maintaining the *src/lib/libast/tm/tmxdate.c* code. Of course, removing `tmxdate()` introduces it's own backward incompatible change. Specifically, `sleep '7 day'` won't be supported. But since that has never been documented it should be acceptable.

siteshwar:
We should revert this change:

```
306 13-05-22 +Added -c option to typeset which is similar to typeset -m except
307       that the variable is copied rather than moved.  This is experimental
308       and there are known bugs.  If it can't be made reliable it will
309       not be added.
```

siteshwar:
Could this change cause issues with compatibility ?

```
13-05-29  ksh93 now intercepts the `LC_TIME` variable.
```

siteshwar:
Another experimental change:

```
226 13-07-30 +An experimental change to each enumeration variable have sub-variables
227       for each enumeration constant.  ${enum.name} will expand to the
228       numerical value of the enumeration name associated with enum.
```

siteshwar:
```
182 - 13-05-29  namespace commands are no longer allowed inside function definitions
183       and now generate a syntax error.
```

This is a backward incompatible change too. But it looks like a sane change that can be easily worked around in scripts by moving namespace definitions outside functions. I will keep it.

PR #1126, opened 2019-01-21
`${foo.__}` should not expand to parent node

siteshwar:
This was an experimental feature that was added after last stable
release. Disable it.

Related: #1116

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -661,15 +661,6 @@ static_fn char *stack_extend(Shell_t *shp, const char *cname, char *cp, int n) {
     return name;
 }
 
-static_fn Namval_t *nv_parentnode(Namval_t *np) {
-    Namval_t *mp = np;
-    if (nv_istable(np)) return nv_parent(np);
-    mp = nv_typeparent(np);
-    if (mp) return mp;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/name.c
```diff
@@ -1047,22 +1047,6 @@ Namval_t *nv_create(const char *name, Dt_t *root, int flags, Namfun_t *dp) {
                 cp++;
                 break;
             }
-            case '_': {
-                if (cp[1] == '_' && (cp[2] == 0 || cp[2] == '.') && shp->oldnp) {
-                    cp += 2;
-                    dp->last = cp;
-                    nvcache.ok = 0;
-                    shp->oldnp = np = nv_parentnode(shp->oldnp);
```
krader1961: This is the only use of `nv_parentnode()`. Remove it's definition too?

krader1961:
LGTM modulo also removing the now unused `nv_parentnode()` implementation.

dannyweldon:
While this _was_ initially experimental, it was subsequently documented in the man page after some bugs were fixed and regression tests were added, which would indicate that it was considered ready for release.

    13-08-23  Documented __ in sh.1 man page.
    13-08-22 Fixed __ to work in nested types.

So if you still want it disabled, you might want to remove the man page entry and the regression tests.  It is an important but minor feature btw for user-defined types and it would be a shame to delay it's release, but whatever.

Issue #1129, opened 2019-01-22
closing stdin in a subshell leaks into the parent when posix_spawn() is enabled


krader1961:
I recently made it possible to build again with ksh using `posix_spawn()` (see commit be962c7). That results in the *basic.sh* unit test to fail. I've boiled it down to a minimal test case and can reproduce on macOS 10.14 and Fedora 28. Replace the content of *src/cmd/ksh93/tests/basic.sh* with this:

```
set -x
x=$($SHELL -c 'echo yes' <&-)

lsof -p $$
print xxx > $TEST_DIR/file
for x in 1 2
do
    tee >(sleep 1; cat > $TEST_DIR/file) <<< "hello" > /dev/null
    lsof -p $$
done
```

Then build and test with `meson -Duse-spawn=true; ninja; meson test basic`. If you inspect *build/meson-logs/testlog.txt* you should see that the subshell dies from a SIGSEGV when executing the builtin `cat`. Why this requires a `for` loop is unknown but removing that loop does not results in failures. The `lsof -p $$` aren't required to reproduce the problem. They're only to show that the shells stdin is closed.

Also, the SIGSEGV is because after executing `fp = sfstdin;` the subsequent use of `fp` is fatal because `sfstdin` is NULL due to the problem described above.

krader1961:
And, of course, this is another example for why I wrote in issue #1117 that stdin, stdout, and stderr should never be left closed. The problem here appears to involve the non-forking optimization described in issue #480 but I don't really know at this stage.

PR #1130, opened 2019-01-22
Fix mistaking sh_state() for sh_isstate()

### LINKED ISSUE #1074: No-fork bug in subshell.c
This statement

https://github.com/att/ast/blob/e756d5b57410fff4db657cdf9f4d94ff2a8958f3/src/cmd/ksh93/sh/subshell.c#L582

is broken as the condition can never be true since `sh_state(SH_NOFORK)` always returns a non-zero value. Thus the `sh_onstate(shp, SH_NOFORK)` will never be executed. So the question is what was intended? This is one of only two places where `sh_onstate(shp, SH_NOFORK)` appears (the other is in *xec.c*).

P.S., This same bug exists in the ksh93u+ release.

krader1961:
I took another look at this in light of issue #1129. Which caused
me to realize it is clear the author of that assignment meant to use
`sh_isstate()` to find out if the flag is set. Sadly this doesn't fix
issue #1129.

Fixes #1074

### DIFF
--- a/src/cmd/ksh93/sh/subshell.c
+++ b/src/cmd/ksh93/sh/subshell.c
@@ -53,6 +53,10 @@
 #include "shnodes.h"
 #include "variables.h"
 
+#if USE_SPAWN
+#include "ast_sys.h"
+#endif
+
 #ifndef PIPE_BUF
 #define PIPE_BUF 512
 #endif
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #1141, opened 2019-01-29
Fix cppcheck null pointer dereference warning

### LINKED ISSUE #1140: cppcheck warns about a bogus non-NULL test
The `cppcheck` tool is issuing this warning:

```
[src/lib/libast/include/sfio.h:343] warning (nullPointerRedundantCheck):
 Either the condition 'if(sp&&flag==3)' is redundant or there is possible null pointer dereference: f.
```

The warning is because `sffileno()` unconditionally dereferences its argument. But this block of code implies `sp` might be NULL:

https://github.com/att/ast/blob/ce9c44c9668fd6da6c93d964b88e93fe2b746dcc/src/cmd/ksh93/sh/io.c#L1439-L1441

The warning is because `io_patseek()` requires `sp` to be non-NULL but we then proceed after that function returns to test that it is non-NULL. That extra test should be eliminated or replaced by an `assert()`.

krader1961:
Fixes #1140

### DIFF
--- a/src/cmd/ksh93/sh/io.c
+++ b/src/cmd/ksh93/sh/io.c
@@ -1437,8 +1437,10 @@ int sh_redirect(Shell_t *shp, struct ionod *iop, int flag) {
                         goto fail;
                     }
                     if (!sp) sp = sh_iostream(shp, fn, fn);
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

Issue #1137, opened 2019-01-28
Upgrade to gcc-9.0 breaks build on Fedora rawhide


siteshwar:
Fedora rawhide has recently rebased to gcc-9.0 and `ksh` copr builds have been broken since then. Build log can be found [here](https://copr-be.cloud.fedoraproject.org/results/@ksh/latest/fedora-rawhide-x86_64/00849449-ksh/build.log.gz).

siteshwar:
Also, see related discussion on [fedora-devel](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/TH4JKSOEVOZU6Y6LAY2IGUUW53CLOVTW/).

siteshwar:
Build failures are a side effect of  c4a7e3e38f9bca9584e190abf2d652348a649f13.

Issue #1118, opened 2019-01-15
What should be the fate of the getconf builtin?


krader1961:
The man page implies that `getconf` is always available but that isn't the case -- in either ksh93u+ or the current master branch version.

```
KSH PROMPT:1: type getconf
getconf is a tracked alias for /usr/bin/getconf
KSH PROMPT:2: builtin getconf
KSH PROMPT:3: type getconf
getconf is a shell builtin
```

On my primary macOS server the `meson` configure step takes 16.4 seconds to complete. Of that 9.8 seconds are spent running the *src/lib/libast/comp/conf.sh*  script that generates the data structures used by the `getconf` builtin.

The behavior of the builtin has some questionable behaviors. For example, it recognizes both `_POSIX2_PBS_ACCOUNTING` (the correct spelling) and `_POSIX_PBS_ACCOUNTING` (which is not correct). See http://pubs.opengroup.org/onlinepubs/009604499/functions/sysconf.html. And a lot of the variables are of dubious relevancy today; e.g., `_POSIX_PII_OSI` and `_POSIX2_FORT_DEV`. The former is related to obsolete ISO/OSI networking protocols and the latter tells whether the system has a FORTRAN 77 compiler. Others, such as `LIBSUFFIX` have the wrong value (`.so` when it should be `.dylib` on macOS).

This is also a command that would not be needed in a busybox type application.

krader1961:
FWIW, the underlying AST code for this feature is in the */port/* directory which implies it was originally meant to provide support for platforms without a `getconf` command or APIs like `sysconf()` such as UWIN. That is not the case today since the project now requires at least a modicum of native support for the POSIX, and related, standards.

krader1961:
While I whined in my original problem statement about the cost of the *src/lib/libast/comp/conf.sh* script when building ksh I did not mean to imply I want to remove this command just because it makes the build slow. My point is that it performs a huge number of feature tests in an extremely inefficient manner. And that cost is difficult to justify since a) the command isn't enabled by default and b) any script using `getconf` is almost certain to be satisfied by the external command of the same name provided by the platform.

siteshwar:
[...]

PR #1142, opened 2019-01-29
Fix build failures caused by update to gcc 9.0

siteshwar:
gcc 9.0 prints line numbers with error messages which affects parsing of
error messages in `conf.sh` script. Keep diagnostic error messages
disabled.

Resolves: #1137

### DIFF
--- a/scripts/libast_prereq.sh
+++ b/scripts/libast_prereq.sh
@@ -25,9 +25,18 @@ fi
 
 cd "$MESON_BUILD_ROOT"
 
+if cc --version | grep -q "GCC"
+then
+    gcc_major_version=$(cc -dumpversion | cut -d. -f1)
+    if [ "$gcc_major_version" -ge 9 ]
+    then
+        extra_flags="-fno-diagnostics-show-line-numbers"
+    fi
+fi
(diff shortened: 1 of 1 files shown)


scripts/libast_prereq.sh
```diff
@@ -25,9 +25,15 @@ fi
 
 cd "$MESON_BUILD_ROOT"
 
+gcc_major_version=$(cc -dumpversion)
+if [ "$gcc_major_version" -ge 9 ]
```
krader1961: This will fail if `cc` is an alias for `clang` as `clang -dumpversion` outputs an string like `4.2.1` (the current value on macOS).

krader1961: In fact, it's kind of weird that it only outputs the major version on gcc.

krader1961: Also, FWIW, if we just got rid of the builtin `getconf` command this awful script could be eliminated. See issue #1118.

siteshwar: Fixed.

krader1961:
LGTM

PR #1144, opened 2019-01-30
Update readme to document new mailing list

### LINKED ISSUE #1013: Shall we set up a new mailing list ?
The lists under lists.research.att.com are not working anymore. Also, it's unlikely they will be made functional again and will not be kept well maintained in future. I would like to hear comments about if we should set up a new mailing list. If yes, what mailing list providers we can use.

krader1961:
Resolves #1013

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,11 @@ AT&T KornShell users.
 
 ### Mailing lists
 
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #1145, opened 2019-01-30
Add tests for `chmod` builtin

siteshwar:
Add new tests for different flags of `chmod` builtin. Tests for `-H` and
`-P` flags are disabled as they are broken.


src/cmd/ksh93/tests/chmod.sh
```diff
@@ -0,0 +1,203 @@
...
+# At the moment this is primarily to avoid a regression of issue #949.
+
+# chmod - change the access permissions of files
+
+#  chmod changes the permission of each file according to mode, which can be
+#  either a symbolic representation of changes to make, or an octal number
+#  representing the bit pattern for the new permissions.
+#
+
+# OPTIONS
```
krader1961: Seems like a weird comment.

src/cmd/ksh93/tests/chmod.sh
```diff
@@ -0,0 +1,203 @@
...
+    }
+elif [[ $(stat --format '%A' a 2>/dev/null) == '-rw-------' ]]
+then
+    # This appears to be a GNU `stat` command that supports the `-A` flag.
+    function stat_perms {
+        stat -c '%A' "$1"
+    }
+else
+    # We have no idea how to report pathname permissions in a human readable form.
+    log_error "no usable 'stat' command found"
```
krader1961: The way I've handled this type of situation in other unit tests is to use this pattern since there isn't any point running any of the test and this doesn't really represent a failure of the `chmod` builtin:

```
log_warning "no usable 'stat' command found"
exit 0
```

krader1961:
LGTM modulo my recommendation to exit the test early without error if we can't portably stat the files.

PR #1154, opened 2019-02-07
Add tests for `cmp` builtin

siteshwar:


src/cmd/ksh93/tests/cmp.sh
```diff
@@ -0,0 +1,97 @@
+# Tests for `cmp` builtin
+
+cat  > "$TEST_DIR/file1" <<EOF
```
krader1961: Just a reminder that `$TEST_DIR` is guaranteed to be the PWD when the test begins. I'm not opposed to being explicit here but it's not required.

PR #1155, opened 2019-02-08
The =~ operator in the `test` command is an error

### LINKED ISSUE #1152: `test foo =~ bar` causes crash
**Description of problem:**

Using the `=~` operator with the `test`/`[` builtin reliably causes ksh to crash.

**Ksh version:**

Version A 2017.0.0-devel-2327-gb56dd2a

**Steps to reproduce:**

    $ ../ksh -c 'test foo =~ foo'
    Abort trap: 6
    $ ../ksh -c '[ foo =~ foo ]'
    Abort trap: 6

krader1961:
Fixes #1152

### DIFF
--- a/src/cmd/ksh93/bltins/test.c
+++ b/src/cmd/ksh93/bltins/test.c
@@ -545,7 +545,21 @@ int test_binop(Shell_t *shp, int op, const char *left, const char *right) {
         case TEST_LE: {
             return lnum <= rnum;
         }
-        default: { abort(); }
+        case TEST_REP: {
+            // The ksh93u release treated this as a silent failure. That is, `test foo =~ foo` would
+            // simply return zero (false) regardless of the operands and whether the condition is
+            // true. We now alert the user that the `=~` binop is not supported by the POSIX test
+            // command. See https://github.com/att/ast/issues/1152.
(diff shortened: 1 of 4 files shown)


siteshwar:
LGTM, but please mention it in the changelog.

src/cmd/ksh93/bltins/test.c
```diff
@@ -545,7 +545,21 @@ int test_binop(Shell_t *shp, int op, const char *left, const char *right) {
         case TEST_LE: {
             return lnum <= rnum;
         }
-        default: { abort(); }
+        case TEST_REP: {
+            // The ksh93u release treated this as a silent failure. That is, `test foo =~ foo` would
+            // simply return zero (false) regardless of the operands and whether the condition is
+            // true. We now alert the user that the `=~` binop is not supported by the POSIX test
+            // command. See https://github.com/att/ast/issues/1152.
+            errormsg(SH_DICT, ERROR_exit(2), e_badop, "=~");
```
siteshwar: May be we can provide a more helpful error message here. It should suggest to use `[[` operator.

krader1961: Good idea. Done.

PR #1156, opened 2019-02-08
Document the limitations of `fmtbuf()`

### LINKED ISSUE #962: Investigate use of memory allocated by `fmtbuf()` function
`fmtbuf()` function is used to allocate small chunks of memory that are short lived. We have seen one bug report (#959) due to prolonged use of memory allocated by this function. This issue tracks if there are other places in code that abuse memory allocated by `fmtbuf()`.

krader1961:
Previous changes have eliminated questionable, unnecessary, or unwanted
uses of `fmtbuf()`. Eliminating the remaining uses is more trouble than
it's worth. Instead, remove support for a "big" buffer since that should
never occur. And make it crystal clear that only small buffers can be
acquired; specifically, no larger than 1 KiB.

Resolves #962

### DIFF
--- a/src/lib/libast/string/fmtbuf.c
+++ b/src/lib/libast/string/fmtbuf.c
@@ -23,35 +23,42 @@
 
 #include <stdlib.h>
 
-#include "ast.h"
 #include "ast_aso.h"
+#include "ast_assert.h"
 
-/*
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

PR #1159, opened 2019-02-09
Change `union Value` to `struct Value` to aid in detecting bugs

krader1961:
**This will not be merged** in its current form. There are still two tests failing as a consequence of changing `union Value` to `struct Value` with metadata. But now that almost all tests are passing with this in place it is time to get feedback on the general idea.


krader1961:
This is now passing all tests. With ASAN enabled it introduces no new failures. I believe this is now robust enough to be merged. This will allow identifying mismatches between the stored and fetched data types. Which might identify and correct other bugs.

It is unfortunate that this change is so large but that is largely a consequence of the original code playing fast and loose with the C language latitude regarding pointers and concrete types.

krader1961:
Merged. It disables detection of store/fetch type mismatches although a few changes in this PR fix such problems based on this feature. It does increase the size of `union Value` deliberately, despite not using the extra meta-data by default, in order to help detect bugs. A future change will eliminate that meta-data from the structure for non-debug builds.

Issue #1157, opened 2019-02-08
`cut -b` and `cut -c` causes shell to abort


siteshwar:
**Description of problem:**
`cut -b` and `cut -c` causes abort

**Ksh version:**
2017.0.0-devel-2338-g917cec4c

**How reproducible:**
Always

**Steps to reproduce:**
```
KSH > cat x
a:b:c
x:y:z
xyz
KSH > builtin cut
KSH > cut -b1 x
```
**Actual results:**
../src/lib/libcmd/cut.c:322: failed assertion 'lp == cut->list'

**Expected results:**
```
a
x
x
```

**Additional info:**
It works as expected with `ksh93v-` release.

siteshwar:
It's a regression introduced by e9220475fef45650471b4e7e2af79d51119ff35c

krader1961:
Comment out the assert for the moment. The problem is that, like so much of the code in this project, it is too clever by half. It is broken in so far as it does access memory past the end of the `Cut_t` structure. Or it would except `struct Cut_s` is defined in a way that accidentally ensures that the code doesn't access unallocated memory. It does access an uninitialized slot in the `cut->list` array but apparently doesn't actually use the invalid data.

saper:
@krader1961 At least here https://github.com/att/ast/blob/e9220475fef45650471b4e7e2af79d51119ff35c/src/lib/libcmd/cut.c#L140 it tries to allocate "enough" space. Can coverity be told that the last item of a struct can be of undetermined length? It is quite a common C idiom I think...

krader1961:
I actually instrumented the code and confirmed it does allocate "enough" space but doesn't correctly initialize it and thus does read a garbage value. Basically, the person who wrote this code was being clever rather than writing clear, obviously correct, code.

Also, the idiom you're referring to should never define the final element to be anything other than a zero length array (or pointer). You'll often see it written as an array of length one which is wrong but in this instance it's defined as length two which is bollocks. Lastly, the allocation is a pessimistic allocation that in general will allocate more space than is required. Which is fine but there should be a comment explaining why it is using `strlen(cp)`.

siteshwar:
There was another regression introduced by 8be93c88cf494c1a4676b47f415b4e79557df023 and the test case in description still does not work.

PR #1161, opened 2019-02-11
Extend cut tests

siteshwar:
Resolves: #1157

### DIFF
--- a/src/lib/libcmd/cut.c
+++ b/src/lib/libcmd/cut.c
@@ -245,7 +245,7 @@ static void cutcols(Cut_t *cut, Sfio_t *fdin, Sfio_t *fdout) {
     int skip;  // non-zero for don't copy
     int must;
     const char *xx;
-    const int *lp = cut->list;
+    const int *lp;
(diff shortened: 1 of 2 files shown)


src/lib/libcmd/cut.c
```diff
@@ -258,6 +258,7 @@ static void cutcols(Cut_t *cut, Sfio_t *fdin, Sfio_t *fdout) {
         if (!bp && !(bp = sfgetr(fdin, 0, SF_LASTR))) break;
         len = sfvalue(fdin);
         xx = 0;
+        lp = cut->list;
```
krader1961: OMFG! Do you see why this initialization has to be here rather than at the top of the function? It's because of the two increments to `lp` within this module. The second is unconditional and not needed as the statement could be rewritten to avoid the otherwise gratuitous increment. The first, conditional, increment is because the first element of `cut->list` has a special meaning; i.e., it's not really part of the list of columns.

krader1961:
LGTM

PR #1170, opened 2019-02-19
Remove `-f` option of `getconf` builtin

siteshwar:
This option was used to support virtual paths that were added after
`ksh93u+` release. We have already removed support for these virtual
paths, so this option should be removed.

Related: #510


krader1961:
Before making any changes to the `getconf` code I'd like to see issue #1118 resolved. I fee pretty strongly that we should just remove this feature. The implementation is incomprehensible and, as you've recently documented, has numerous bugs.

krader1961:
LGTM in as much as this does nothing more than revert the only change to this module after the ksh93u release.

PR #1175, opened 2019-02-19
Add tests for `getconf` builtin

siteshwar:


krader1961:
Rather than add these tests, which implies we care about maintaining this command, I'd prefer to update the docs to make it clear the command is deprecated and to be used at your own risk. Note that ksh93u fails the `getconf -a` test as well. I vote that rather than fix it we just freeze the implementation. Bugs in `getconf` that are now a minimum of 7 years old aren't worth fixing. There are hundreds, if not thousands, of bugs to be fixed or enhancements we'd like to make that are higher priority than any change to the `getconf` builtin.

siteshwar:
@krader1961 At least these tests will ensure that things won't get worse as we make changes, so I would still prefer to merge this. Regarding the fate of `getconf` I would still like to hear more comments from others.

krader1961:
Okay, I wouldn't have bothered but since you've already written these go ahead and merge. I think we should also add a bold comment at the top of the *getconf.c* module reminding people not to touch it in order to avoid introducing any change in behavior.

src/cmd/ksh93/tests/getconf.sh
```diff
@@ -0,0 +1,101 @@
+# Tests for `getconf` builtin
+
+# ==========
+#   -a, --all       Call the native getconf(1) with option -a.
+actual=$(getconf -a)
```
siteshwar: @krader1961 This test fails on macOS, so I presume `-a` option is not supported on macOS. I am going to disable it and merge this pull request.

PR #1177, opened 2019-02-21
Remove `getconf` from the ksh.1 man page

krader1961:
The `getconf` command is the only optional (i.e., not enabled by default)
builtin mentioned in the main ksh man page. Remove that text.  We do not
want people to

a) think that just executing `getconf ...` will run the ksh builtin
(since it doesn't without first doing `builtin getconf`), and

b) we want to deprecate this command and ultimately remove it in favor
of the platform version.

Related #1118

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ None at this time.
 
 ## Notable fixes and improvements
 
+- Mention of the `getconf` builtin has been removed from the main ksh man
+  page. That command has never been enabled by default and is now deprecated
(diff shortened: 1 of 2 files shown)


krader1961:
If you are inclined to reject this change please justify why this is the only builtin documented in the primary ksh man page that is not enabled by default. And that documentation is woefully incomplete.

siteshwar:
I am fine with removing this from man page, but I don't think there is a concrete plan to deprecate and remove it from ksh.

Issue #1150, opened 2019-02-04
`union Value` causes bugs


krader1961:
Cppcheck has been issuing the following warnings from day one:

```
[src/cmd/ksh93/sh/streval.c:348] portability (invalidPointerCast):
 Casting between integer* and long double* which have an incompatible binary data representation.
[src/cmd/ksh93/sh/streval.c:944] portability (invalidPointerCast):
 Casting between integer* and long double* which have an incompatible binary data representation.
```

But that is only the most obvious problem with how `union Value`  is used. Simply changing it from a union to struct breaks ksh so badly you can't even start an interactive shell with no dot files. So clearly there are multiple places that are retrieving a value using a member different than was used to store the value.

I'm to to remediate this state of affairs by turning it into a struct with a hidden union. Then introduce a setter function that will record which union member was used for the store and a getter function that validates the same union member is used for the fetch.

krader1961:
Here's an example of something that is more or less harmless but obviously wrong from a logic perspective:

```
if (np && np->nvalue.cp) np->nvalue.rp->help = tp->help;
```

That is from *src/cmd/ksh93/bltins/typeset.c*. Obviously if we're using the `struct Ufunction *rp` union member we can't possibly have stored a char pointer in the union.

krader1961:
Another example from the same *typeset.c* module:

```
if (np->nvalue.ip && np->nvalue.rp->hoffset >= 0) {
```

Again, we can't have both a `int *ip` and a `struct Ufunction *rp` stored in the same union. This is truly awful.

krader1961:
The majority of the value type mismatches are between `const char*` and `char*`. Which is itself a problem. Albeit a minor, probably inconsequential, problem. But there are lots of other, more serious, value type mismatches. Such as confusing a `struct pathcomp*` with a `const char*`.

krader1961:
And another bug found as a result of addressing this issue. Function `array_grow()` contains these lines:

```
struct index_array *ap;
size = (newsize - 1) * sizeof(struct Value *) + newsize;
ap = calloc(1, sizeof(*ap) + size);
```

Note that `struct index_array` is defined thusly:

```
struct index_array {
    Namarr_t namarr;
    void *xp;             // if set, subscripts will be converted
    int cur;              // index of current element
    int last;             // index of highest assigned element
    int maxi;             // maximum index for array
    unsigned char *bits;  // bit array for child subscripts
    struct Value val[1];  // array of value holders
};
```

[...]

PR #1179, opened 2019-02-22
Replace struct Value Nambfp_f bfp

krader1961:
Struct Value member `bfp` and its associated type `Nambfp_f` is not
only unnecessary it obscures the intent due to the explicit casts it
requires. This change is probably not optimal. It represents a baby step
to increasing the sanity involving using `struct Value` (formerly `union
Value`). Replace structure member `bfp` with `shbltinp` which eliminates
the casts between the two types.

Related #1150


krader1961:
I've merged this because it is a mechanical change that does not change the logic of the code. It simply replaces a pointless, misleading, typedef with the correct typedef.

PR #1181, opened 2019-02-25
Feature test: Detect num cpus

krader1961:
Several unit tests depend on knowing the number of online CPUs to
work correctly. If that isn't defined the test may default to a value
inappropriate for the system resulting in timeouts.

This also removes three CDT tests that depend on Vmalloc features. Those
tests will never to updated to work with the system malloc.


features/detect_nproc.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+# Exit with a status reflecting the CPU count if we can determine it. If not
+# assume two CPUs since at least one test, API/trehash.c, deadlocks if the CPU
+# count is one.
+if getconf _NPROCESSORS_ONLN >/dev/null 2>&1
+then
+	# Some systems have a getconf that supports this var. Use it in preference
+	# to the options below because those other mechanisms include offline CPUs.
+	exit $(getconf _NPROCESSORS_ONLN)
```
siteshwar: This should work unless we are running on supercomputers with greater than 255 cores. And historically `ksh` has run on supercomputers.

src/cmd/tests/cdt/tvsafehash.c
```diff
@@ -1,293 +0,0 @@
-/***********************************************************************
```
siteshwar: Removing these tests should have been done through a separate commit.

features/detect_nproc.sh
krader1961: Good point. Fixed to ensure we never exit with less than 2 (since at least one test fails to run if nproc is 1) or more than 64 (which is the default for most tests if N_PROC isn't defined).

PR #1183, opened 2019-02-27
Struct namarray

krader1961:
This is a far larger PR, in terms of the number of commits, than I would normally do. The only substantive change is in the first commit whose comment explains the non-trivial aspect of that change.  All of the other changes are trivial, mechanical, changes. I'm including them in this PR to show that code cleanups add a lot of value with respect to making bugs less likely.


siteshwar:
LGTM

PR #1185, opened 2019-02-27
Add tests for `uname` builtin

siteshwar:


krader1961:
I don't understand the rationale for the second commit. Those options and associated code exist in the ksh93u release.

siteshwar:
> I don't understand the rationale for the second commit. Those options and associated code exist in the ksh93u release.

They are neither documented nor recognized by `uname` builtin. Check the output of `uname --man` or try to use these options.

krader1961:
I don't trust mechanisms like `uname --man` output given the poor quality of the code, and documentation, in this project. However, I verified that even the ksh93u release does not recognize those flags. Despite that code implementing support for those flags.

So... LGTM

Issue #1184, opened 2019-02-27
`uname -d` should print domain name returned by `getdomainname`


siteshwar:
**Description of problem:**
`uname -d` causes a crash.

**Ksh version:**
2017.0.0-devel-2407-gdd1f44d3

**How reproducible:**
Always

**Steps to reproduce:**
```
builtin uname
uname -d
```

**Actual results:**
It causes `ksh` to crash.

**Expected results:**
It should print domain name returned by `getdomainname`.

**Additional info:**
It does not cause a crash in `ksh93v-` release.

krader1961:
Works for me on macOS and other BSD. Albeit it is not useful since it outputs `[domain]`. But then I don't run NIS/YP on any of my systems. Having said that the implementation is obviously broken:

```
        if (flags & OPT_domain) {
            s = astconf("SRPC_DOMAIN", NULL, NULL);
            if (!(*s)) getdomainname(s, sizeof(buf));

            output(OPT_domain, s, "domain");
        }
```

Notice that it is passing the pointer returned by `astconf()` to `getdomainname()` while telling that function the pointer is to a buffer of size `sizeof(buf)`. Bzzt! Wrong!

This feature isn't in the ksh93u release so I vote we simply remove it rather than fix it.

PR #1187, opened 2019-03-01
`uname -d` should not crash

siteshwar:
Pass correct buffer to `getdomainname()` while executing `uname -d`.

Resolves: #1184


src/lib/libcmd/uname.c
```diff
@@ -298,8 +298,10 @@ int b_uname(int argc, char **argv, Shbltin_t *context) {
 #endif
         if (flags & OPT_domain) {
             s = astconf("SRPC_DOMAIN", NULL, NULL);
-            if (!(*s)) getdomainname(s, sizeof(buf));
-
+            if (!(*s)) {
```
krader1961: I would drop the pointless inner parens but it's fine either way.

krader1961:
LGTM

PR #1189, opened 2019-03-04
Change type of struct Namval member nvenv

krader1961:
98% of the uses of struct Namval member nvenv involve a Namval_t. So change
the type of that member so all of those casts can be eliminated.  There are
a couple of places that actually do treat it as a char*.  For the moment
simply cast the handful of places which overload `nvenv` to store a char*.


siteshwar:
LGTM

PR #1192, opened 2019-03-04
Remove stack related unused functions

siteshwar:


krader1961:
LGTM. Don't forget to also remove the unused *src/lib/libast/include/stack.h* header.

PR #1197, opened 2019-03-05
Refactor and add tests for `pathcat` function

siteshwar:


src/lib/libast/tests/path/pathcat.c
```diff
@@ -0,0 +1,59 @@
...
+    int failed = 0;
+    const char *root, *path1, *path2;
+    char concat_path[BUFSIZ];
+    int separator;
+    char *result;
+
+    root = "/foo;";
+    path1 = "bar";
+    path2 = "baz";
+    separator = (int)';';
```
krader1961: A char literal is an int so the cast isn't needed.

krader1961:
LGTM

PR #1198, opened 2019-03-06
Fix bug in uname builtin

krader1961:
The program incorrectly uses var `s` in the `gethostname()` call. Also,
the test of the length of `ut.nodename` is always true on all systems
I've checked (Linux, *BSD, macOS) so `gethostname()` won't even be called.
And, as far as I can determine, the days when the fields of `struct uname`
might be just 8 or 9 chars in length are long past.


siteshwar:
LGTM

PR #1200, opened 2019-03-07
Refactor and test `pathbin` function

siteshwar:


src/lib/libast/path/pathbin.c
```diff
@@ -19,28 +19,30 @@
...
+    char *path;
+    // Used to restore value of saved astconf path from earlier function call
+    static char *astconf_path;
 
-    static char *val;
-
-    if ((!(bin = getenv("PATH")) || !*bin) && !(bin = val)) {
-        if (!*(bin = astconf("PATH", NULL, NULL)) || !(bin = strdup(bin))) {
-            bin = "/bin:/usr/bin:/usr/local/bin";
+    path = getenv("PATH");
```
krader1961: Yes, please continue to move assignments inside `if()` statements that can be easily hoisted.

src/lib/libast/path/pathbin.c
```diff
@@ -19,28 +19,30 @@
...
+    if ((!path || !*path) && !(path = astconf_path)) {
```
krader1961: And here we have an example why such assignments are dangerous. Note that `astconf_path` is uninitialized at this juncture which means the condition will be true or false depending on whatever random garbage is on the stack when this function is called. It looks like you intended to drop this assignment and test from the `if()` since you moved it into the body.

src/lib/libast/path/pathbin.c
```diff
@@ -19,28 +19,30 @@
...
 
 #include "ast.h"
 
+//  Return current PATH
+//
 char *pathbin(void) {
-    char *bin;
+    char *path;
+    // Used to restore value of saved astconf path from earlier function call
+    static char *astconf_path;
```
krader1961: I know it's implicitly initialized to zero but it's better to add an explicit `= NULL;` initializer.

src/lib/libast/path/pathbin.c
```diff
@@ -19,28 +19,30 @@
...
 
-    static char *val;
-
-    if ((!(bin = getenv("PATH")) || !*bin) && !(bin = val)) {
-        if (!*(bin = astconf("PATH", NULL, NULL)) || !(bin = strdup(bin))) {
-            bin = "/bin:/usr/bin:/usr/local/bin";
+    path = getenv("PATH");
+    if ((!path || !*path) && !(path = astconf_path)) {
+        path = astconf("PATH", NULL, NULL);
+        if (!*path || !(path = strdup(path))) {
```
krader1961: I'd restructure this to hoist the assignment.

```
if (path && *path) path = strdup(path);
if (!path) path = "/bin:/usr/bin:/usr/local/bin";
```

Although I'm not entirely convinced testing if the strdup() succeeded is a good idea or really helpful.

krader1961:
LGTM

Issue #1153, opened 2019-02-07
The pattern for using `optget()` should be standardized


krader1961:
Once in awhile I spend time resolving lint warnings. While doing so for the *src/lib/libcmd/mkdir.c* module the addition of a `default: { abort(); }` statement to the `optget(()` loop caused the abort to be unexpectedly executed.  That's because the code used this anti-pattern:

```
for (;;) {
    switch(optget(argv, usage)) {
        case 'm':
            // do stuff
            continue;
        // more cases
    }
    break;
}
```

Basically, process options until we see a flag we don't recognize. I put "basically" in scare-quotes because the switch block explicitly handles `case ':':` and `case '?':` for invalid flags. **The switch block just doesn't handle the zero return value that means there are no more options!** Instead it relies on the anti-pattern above to break out of the `for(;;)` flag parsing loop.

There are other modules that use the same or similar anti-pattern for parsing flags. These should all be normalized to a consistent, sane, pattern. In fact, most of the code uses this pattern:

```
while ((n = optget(argv, sh_optexec))) {
    switch (n) {
    }
}
```

In other words, the flag parsing clearly stops as soon as the last flag is processed.

**Every single place in this project that calls `optget()` should use the same pattern.** Which is not to say the above example is ideal. But the code should use a consistent pattern that doesn't result in lint.

krader1961:
Note that it seems like all the code in *src/lib/libcmd/* uses this anti-pattern. While the other callers of `opget()` employ a saner pattern. Which pattern is optimal from a human readability viewpoint isn't the point. The point is that all the code in the project should employ the same pattern.

krader1961:
FWIW, besides the lint warnings this pattern is confusing:

```
for(;;) {
    switch(optget(...)) {
        case 'x':
            // do something with arg 'x'
            continue;
        case ':':
            // issue error message
            break;
    }
    break;
}
```

Note that a `continue` inside a `case` block continues at the top of the enclosing loop. But a `break` inside a `case` block exits the `switch` block rather than breaking out of the enclosing loop. While the `break` at the bottom of that `switch` block terminates the enclosing loop. Yes, I would expect anyone who claims to understand the C language to know that since it is a legal construct in the language. **But it is still an anti-pattern that will confuse people!** Worse, if someone changes the structure such as by introducing a nested loop it is likely to introduce a bug. Which is why tools like oclint warn about that idiom.

krader1961:
[...]

PR #1201, opened 2019-03-08
Normalize optget() usage in enum.c

krader1961:
This makes the b_enum() use of optget() match that seen in other
modules such as cd_pwd.c. It also eliminates this oclint warning about
an anti-pattern:

    src/cmd/ksh93/bltins/enum.c:286:9: avoid branching statement as last
    in loop [convention|P2]

It also eliminates two of the four uses of the `error(ERROR_USAGE | x, ...)`
pattern. Bringing those error message invocations in line with all the
other code.

Related #1153


siteshwar:
LGTM

PR #1203, opened 2019-03-08
Refactor and add tests for `pathaccess()` function

siteshwar:


src/lib/libast/path/pathaccess.c
```diff
@@ -38,19 +39,21 @@
 
 #include "ast.h"
 
-char *pathaccess(const char *dirs, const char *a, const char *b, int mode, char *path,
+char *pathaccess(const char *dirs, const char *path1, const char *path2, int mode, char *path,
                  size_t size) {
-    int sib = a && a[0] == '.' && a[1] == '.' && a[2] == 0;
-    int sep = ':';
+    int is_path1_parentdir = path1 && path1[0] == '.' && path1[1] == '.' && path1[2] == 0;
```
krader1961: FYI, I am not a fan of the micro optimizations such a this that exist in the code. They often result in the code accessing bytes past the end of a buffer/string. They also are harder to read. So while doing these types of cleanup I recommend changing them to a more idiomatic formulation. For example:

```
bool is_path1_parentdir = path1 && strcmp(path1, "..") == 0;
```

src/lib/libast/path/pathaccess.c
```diff
@@ -19,17 +19,18 @@
...
- * if (mode&PATH_REGULAR) then path must not be a directory
- * if (mode&PATH_ABSOLUTE) then path must be rooted
- * path returned in path buffer
- */
+//
+// Glenn Fowler
+// AT&T Bell Laboratories
+//
+// Return path to file path1/path2 with access mode using `:` separated dirs
+// both path1 and path2 may be 0
```
krader1961: s/0/NULL/

krader1961:
LGTM modulo a couple of nits.

PR #1204, opened 2019-03-09
Refactor and add tests for `pathgetlink()` function

siteshwar:


src/lib/libast/tests/path/pathgetlink.c
```diff
@@ -0,0 +1,42 @@
...
+#include "ast.h"
+#include "terror.h"
+
+tmain() {
+    UNUSED(argc);
+    UNUSED(argv);
+    const char *linkname = "foo";
+
+    char buff[10];
+    char filename[] = "foobarbaz";
```
krader1961: Why isn't this `const char *filename = "foobarbaz";`?

src/lib/libast/path/pathgetlink.c
```diff
@@ -31,19 +31,19 @@
...
-/*
- * return external representation for symbolic link text of name in buf
- * the link text string length is returned
- */
-
-int pathgetlink(const char *name, char *buf, int siz) {
+//
+// Return external representation for symbolic link text of name in buff
+// the link text string length is returned
+///
```
krader1961: Just two forward slashes will do :smile:

krader1961:
LGTM modulo two minor nits.

src/lib/libast/tests/path/pathgetlink.c
siteshwar: I was playing around with some other functions that required this variable to be writable, but will fix it.

krader1961:
FWIW, There is far too little use of the `const` qualifier in this project where using that qualifier would help to identify bugs or APIs that are ill-defined.

PR #1205, opened 2019-03-09
Refactor and test `pathstat()` function

siteshwar:


src/lib/libast/tests/path/pathstat.c
```diff
@@ -0,0 +1,37 @@
...
+
+    struct stat statbuf;
+
+    if (open(temp_file, O_CREAT, 0666) < 0) terror("Failed to create temp file");
+
+    if (pathstat(temp_file,  &statbuf) && statbuf.st_mode == 0666) terror("Failed to stat() %s", temp_file);
+
+    if (symlink(temp_file, link_to_temp_file) < 0) terror("Failed to create symbolic link to %s", temp_file);
+
+    atexit(cleanup_handler);
```
krader1961: We should arrange to run these tests in a unique temp dir like we do for the ksh tests. Thus obviating the need for these kind of cleanup functions.

src/lib/libast/path/pathstat.c
```diff
@@ -24,15 +24,15 @@
...
-int pathstat(const char *path, struct stat *st) {
-    int oerrno;
-
-    oerrno = errno;
-    if (!stat(path, st)) return 0;
-    errno = oerrno;
-    return lstat(path, st);
+    old_errno = errno;
+    if (!stat(path, statbuf)) return 0;
+    errno = old_errno;
```
krader1961: I'm not suggesting you remove this but for the life of me I can't see the point of restoring the original errno value. Any code using this function (and there is only one in the project as I type this) it should only look at errno if this function returns -1 to indicate an error. Thus there should be no need to save/restore errno.

krader1961:
LGTM

PR #1206, opened 2019-03-10
Add test for relative paths in `pathaccess()` function

siteshwar:


src/lib/libast/tests/path/pathaccess.c
```diff
@@ -1,20 +1,41 @@
...
     if (pathaccess(paths, 0, 0, PATH_ABSOLUTE, path, PATH_MAX)) terror("pathaccess() returns path to a file that does not exist");
 
+    // Test for relative paths
+    if (open(relative_path, O_CREAT, 0666) < 0) {
+        terror("Failed to create test file");
+    }
+
+    atexit(cleanup_handler);
+
+    if (!getcwd(cwd, PATH_MAX)) terror("Failed to get current working directory");
```
krader1961: FWIW, it's safer to use `sizeof(cwd)` than `PATH_MAX` (same thing below). That way if someone at a later date decides to change the size of the array these statements won't need to be updated.

krader1961:
LGTM

PR #1207, opened 2019-03-10
Add an option to build API tests only

siteshwar:
This option would be helpful while checking coverage of API tests.


krader1961:
Not sure I see the point but feel free to merge. It seems to me we only care about the coverage provided by all tests combined. The API tests alone will never cover more than a tiny percentage of the code.

siteshwar:
> Not sure I see the point but feel free to merge. It seems to me we only care about the coverage provided by all tests combined. The API tests alone will never cover more than a tiny percentage of the code.

Sure. At the end we only care about coverage provided by both type of tests. I am currently working on API tests and it's time consuming to run shell tests along with API tests, so I added an option to disable shell tests for my convenience.

PR #1210, opened 2019-03-12
Remove definition of `pathnative()` function

siteshwar:
This function does nothing more than copying existing path to a new
buffer. It's not needed anymore, so remove it.


krader1961:
LGTM. Looking at the old beta branch it appears this was meant primarily, if not solely, to support UWIN. And since we have made an explicit decision to drop support for UWIN and other non-UNIX like environments it's clearly not needed.

PR #1212, opened 2019-03-12
Refactor and add tests for `pathcanon()` function

siteshwar:


src/lib/libast/tests/path/pathcanon.c
```diff
@@ -0,0 +1,99 @@
...
+    if (strcmp(path, filename)) terror("Failed to resolve symbolic link");
+
+    // Try to resolve physical path from broken symbolic link
+    unlink(linkname);
+    strcpy(path, linkname);
+    pathcanon(path, PATH_MAX, PATH_PHYSICAL | PATH_EXISTS);
+    fprintf(stderr, "%s\n", path);
+    if (strcmp(path, "")) terror("Resolving broken link should fail");
+
+    // TODO: Three dots. It is special cased in `pathcanon()` function. Why ?
```
krader1961: It also special-cases four dots. No effing idea why. Possibly related to the named extended attr (XATTR) support the code used to have. I think this specific test should be removed since we don't really know what the semantics are supposed to be and there shouldn't be any ksh script relying on this "feature". Instead add a TODO to the `pathcanon()` code reminding someone to remove that logic if they can do so without risking breaking other behavior.

src/lib/libast/path/pathcanon.c
```diff
@@ -19,27 +19,27 @@
...
+// path name canonicalization -- preserves the logical view
+//
+//	Remove redundant `.`'s and `/`'s
+//	Move `..`'s to the front
+//	/.. preserved (for pdu and newcastle hacks)
+//	if (flags&PATH_ABSOLUTE) then pwd prepended to relative paths
+//	if (flags&PATH_PHYSICAL) then symlinks resolved at each component
+//	if (flags&(PATH_DOTDOT|PATH_PHYSICAL)) then each .. checked for access
+//	if (flags&PATH_EXISTS) then path must exist at each component
+//	if (flags&PATH_VERIFIED(n)) then first n chars of path exist
```
krader1961: Drop this line. Not even the old beta branch used `PATH_VERIFIED`.

krader1961:
LGTM

krader1961:
LGTM

siteshwar:
> LGTM

These tests are actually broken because it seems `/tmp` is a symbolic link on macOS. That's why I have not merged it.

krader1961:
Yes, on macOS */tmp* is a symlink to *private/tmp*. You might want to use the value of `$TMPDIR` if it's set.

PR #1213, opened 2019-03-13
Lint: avoid branching statement as last in loop

krader1961:
What's fascinating about the code being changed is that the "while"
loop is never executed more than once since there is no `continue`
statement in its body. It should have been written as an `if()...` but
apparently the author wanted to avoid a `goto`. Instead they preferred
to use a confusing loop that doesn't actually loop. Which is weird given
how many goto statements already exist in this code.

P.S., I attempted to refactor this code by hoisting the entire `while()`
loop into a separate function. But doing so isn't practical given the
embedded `goto` statements that reference a label outside of the loop body.


siteshwar:
LGTM

PR #1217, opened 2019-03-14
Add `--allowerasing` parameter to dnf in travis

siteshwar:
Builds for fedora on Travis CI broke because `expect` package now
conflicts with `whois-mkpasswd` package that is installed by default.
This commit fixes it.

https://bugzilla.redhat.com/show_bug.cgi?id=1687870


krader1961:
LGTM

PR #1223, opened 2019-03-15
Coverity 253677 Dereference before null check

### LINKED ISSUE #1222: Logic bug in p_comlist()
Coverity Scan 253677 warns about a "Dereference before null check" in `p_comlist()`:

https://github.com/att/ast/blob/0c145d1089dd3af9bf734d06664ceba1528fcfce/src/cmd/ksh93/sh/deparse.c#L472-L488

It should be fairly obvious that `cp` can never be NULL at the top of that `do ... while()` loop.

krader1961:
Coverity Scan issue 253677 points out that a non-NULL test is always
true.

P.S., This code is never exercised by unit tests.

Fixes #1222

### DIFF
--- a/src/cmd/ksh93/sh/deparse.c
+++ b/src/cmd/ksh93/sh/deparse.c
@@ -469,14 +469,9 @@ static_fn void p_comlist(const struct dolnod *dol, int endchar) {
     int flag = ' ', special;
 
     argv = dol->dolval + ARG_SPARE;
-    cp = *argv;
-    special = (cp[0] == '[' && cp[1] == 0);
+    cp = *argv++;
+    special = strcmp(cp, "[") == 0;
     do {
-        if (cp) {
-            argv++;
-        } else {
-            cp = "";
-        }
         if (*argv == 0) {
             flag = endchar;
             if (flag == '\n') begin_line = 1;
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/sh/deparse.c
```diff
@@ -469,22 +469,17 @@ static_fn void p_comlist(const struct dolnod *dol, int endchar) {
...
     argv = dol->dolval + ARG_SPARE;
-    cp = *argv;
-    special = (cp[0] == '[' && cp[1] == 0);
+    cp = *argv++;
+    special = strcmp(cp, "[") == 0;
     do {
-        if (cp) {
-            argv++;
-        } else {
-            cp = "";
```
siteshwar: What would happen if `cp` points to `NULL` character ?

krader1961: If you're asking what happens if `cp` is NULL that can't happen. At this juncture we've already dereferenced cp before entering the loop or we've exited the loop because `cp` is NULL (see the `while(...)` condition). Now it may be the case that statement `special = strcmp(cp, "[") == 0;` also needs to also check that `cp` is non-NULL. 

If you really do mean what happens if `cp` points to an empty string (first bye is `'\0'`) then the `cp = "";` doesn't change that situation.

PR #1224, opened 2019-03-15
Lint: avoid branching statement as last in loop

krader1961:
Eliminate another reason oclint warns about a confusing anti-pattern in
this module.


siteshwar:
LGTM

PR #1227, opened 2019-03-17
Correct PS1 info in man page

nyanster:
PS1's default value is a dollar sign followed by a space, not just a dollar sign on its own.
Confirmed this both by checking in ksh93 itself, and by checking posix.  PS2, PS3, and PS4 already show the space included in their default values, in the man page.


nyanster:
Confirmation in code:
https://github.com/att/ast/blob/master/src/cmd/ksh93/data/msg.c#L173

siteshwar:
LGTM. Thanks for the fix!

PR #1228, opened 2019-03-18
Refactor and add tests for `pathpath()` function

siteshwar:


src/lib/libast/tests/path/pathpath.c
```diff
@@ -0,0 +1,46 @@
...
+
+    // Search this path in current directory and expand it
+    pathpath("foo", NULL, PATH_ABSOLUTE, path, PATH_MAX);
+    if (strcmp(path, absolute_test_path))
+        terror("Failed to expand path of file in current directory");
+
+    if (!pathpath("cat", NULL, PATH_EXECUTE, path, PATH_MAX))
+        terror("Failed to find `cat` in current $PATH");
+
+    if (!pathpath(path, NULL, PATH_EXECUTE, path, PATH_MAX))
```
krader1961: Passing `path` twice like this causes this statement at line 65 to fail outright on macOS due to overlapping buffers and causes undefined behavior on every platform:

```
strcpy(path, path_to_search);
```

PR #1231, opened 2019-03-18
Add tests for `ast_temp_file()` function

siteshwar:


src/lib/libast/tests/path/pathtemp.c
```diff
@@ -0,0 +1,28 @@
...
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "ast.h"
+#include "terror.h"
+
+const char *temp_file;
+
+void cleanup_handler(void) { unlink(temp_file); }
```
krader1961: You're missing the `atexit()` call to install this handler. I'd also recommend initializing `temp_file` above to NULL and predicating the unlink on `if (temp_file)`.

PR #1232, opened 2019-03-18
[kshbugbot] Fix code styling

kshbugbot:


siteshwar:
It would be good to have a style check test in our CI. I will see how to do that.

PR #1234, opened 2019-03-20
Add tests for base64 encoding and decoding

siteshwar:


src/lib/libast/tests/string/base64.c
```diff
@@ -0,0 +1,61 @@
...
+    { "iVFjwRZTPrXzQoEbnisP", "aVZGandSWlRQclh6UW9FYm5pc1A=" },
+    { "JDkAUes8uamPqIPwQKGV", "SkRrQVVlczh1YW1QcUlQd1FLR1Y=" },
+    { "bf0bCcTT1LySrR9eHi53", "YmYwYkNjVFQxTHlTclI5ZUhpNTM=" },
+    { "wwsbeqm8xYinra82fDlx", "d3dzYmVxbTh4WWlucmE4MmZEbHg=" },
+    { "pFkDWTcwbGATCiPJpkhy", "cEZrRFdUY3diR0FUQ2lQSnBraHk=" },
+    { "ENwnLtgW5VLvNicthw0N", "RU53bkx0Z1c1Vkx2TmljdGh3ME4=" },
+    { NULL, NULL }
+};
+
+// TODO: What shall we do with the test in 'src/lib/libast/tests/misc/base64.c' ?
```
krader1961: Might as well eliminate it. Looking at it I see it only writes a count of the errors to stdout. The Meson test framework won't actually register a failure of that test even if it does fail.

krader1961:
LGTM

PR #1237, opened 2019-03-21
Coverity CID 253726 Logically dead code

krader1961:
The `if (table)` test is pointless because the immediately preceding
test of `table` guarantees it is true at this juncture. While fixing this
logic error I refactored the code to make the logic clearer for a human.


siteshwar:
LGTM

PR #1241, opened 2019-03-21
Replace calls to `fmterror()` with `strerror()`

siteshwar:
`fmterror()` function internally calls `strerror()` and returns it's
return value. There is no need to have this wrapper function and
`strerror()` can be called directly.


krader1961:
Good catch. LGTM

PR #1244, opened 2019-03-22
Lint: missing default in switch statements

krader1961:
It isn't completely clear whether the original `switch() { ... }` block
enumerates all possible cases. And thus a `default: { abort(); }` is
appropriate. At first glance that appears to be the case; especially since
it does explicitly test for `SF_ECONT`. But that is not at all certain
let alone obvious. So rather than add a "default" case restructure the
code. This eliminates the need for a goto label.

FWIW, I think the existing code relies too heavily on `switch` statements
where other constructs would be simpler and clearer.

### DIFF
--- a/src/lib/libast/sfio/sfrd.c
+++ b/src/lib/libast/sfio/sfrd.c
@@ -93,7 +93,8 @@ ssize_t sfrd(Sfio_t *f, void *buf, size_t n, Sfdisc_t *disc) {
         }
     }
 
-    for (dosync = 0;;) { /* stream locked by sfsetfd() */
+    dosync = 0;  // stream locked by sfsetfd()
+    while (1) {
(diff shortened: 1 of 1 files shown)


src/lib/libast/sfio/sfrd.c
```diff
@@ -275,20 +276,11 @@ ssize_t sfrd(Sfio_t *f, void *buf, size_t n, Sfdisc_t *disc) {
 
     do_except:
         if (local) SETLOCAL(f);
-        switch (_sfexcept(f, SF_READ, (ssize_t)r, dc)) {
-            case SF_ECONT:
```
siteshwar: This line is not covered by tests. While this change looks correct on a quick look, it would be great to have a test for it.

PR #1248, opened 2019-03-25
Add tests for `fmtint()` function

siteshwar:


src/lib/libast/tests/string/fmtint.c
```diff
@@ -0,0 +1,29 @@
...
+
+tmain() {
+    UNUSED(argc);
+    UNUSED(argv);
+
+    const char *actual_result;
+
+    struct IntegerToStringTest tests[] = {
+        {9, "9"},           {123, "123"},   {24321, "24321"}, {10000000, "10000000"},
+        {-10000, "-10000"}, {-151, "-151"}, {0, NULL}};
```
krader1961: This should include a test for 0 => "0" (modify the for loop below to terminate when `tests[i].expected_result` is NULL rather than `tests[i].input`). Ideally also test INT_MIN and INT_MAX but that is a bit more challenging given that ints may be 32 or 64 bits.

krader1961:
LGTM modulo the need to verify zero is correctly converted.

PR #1249, opened 2019-03-25
Add tests for `fmtmode()` function

siteshwar:


src/lib/libast/string/fmtmode.c
```diff
@@ -19,18 +19,23 @@
...
+// if external!=0 then mode is modex canonical
+//
 #include "config_ast.h"  // IWYU pragma: keep
 
 #include "modelib.h"
 #include "modex.h"
 
+// This function always returns same static buffer.
+// It should be used with care.
+// TODO: Fix this function to not use static buffer.
```
krader1961: I'm not convinced the added complexity is needed. Especially not if the solution is to use `fmtbuf()` which itself is a dangerous API. Plenty of libc functions return a static buffer. It is perfectly acceptable in situations where the string has a small bound on its max length and it is unlikely callers will need to call it a second time before the result of the first call is consumed.

krader1961:
LGTM modulo my concern about adding a comment suggesting the use of `fmtbuf()`.

PR #1250, opened 2019-03-25
 Add tests for `fmtperm()` function

siteshwar:


src/lib/libast/string/fmtperm.c
```diff
@@ -19,24 +19,25 @@
...
+// AT&T Bell Laboratories
+//
+// Return strperm() expression for perm
+//
 #include "config_ast.h"  // IWYU pragma: keep
 
 #include <sys/stat.h>
 
+// This function always returns same static buffer
+// TODO: Fix this function to not return same static buffer
```
krader1961: Same concern I had for `fmtmode()`. Having it return a static buffer is a reasonable choice. The caller can always `strdup()` the result if they need to call this a second time before consuming the result of the first call. Using `fmtbuf()` is dangerous -- we don't want more uses of that API. And having this return a `malloc()`'d buffer just means the caller has to always free it.

krader1961:
LGTM

Issue #1243, opened 2019-03-22
API tests aren't hermetic and thus can fail randomly


krader1961:
I've noticed unexpected, transitory, failures from recently introduced API tests on both Travis CI and my local systems. For example,

```
 26/255 API/path/pathgetlink.c                  FAIL     0.03 s (exit status 1)
--- command ---
MALLOC_PERTURB_='165' MallocGuardEdges='1' MALLOC_CHECK_='3' MALLOC_CONF='junk:true' MallocErrorAbort='1' MallocScribble='1' MallocLogFile='/dev/null' /source/build/src/lib/libast/tests/path/path_pathgetlink.c
--- stderr ---
	Line=24: FAILED Failed to create symbolic link [errno=17]
```

Where line 24 is effectively doing `symlink("foobarbaz", "foo");`. So in this case the failure is because filename `foo` already exists.

We need to enhance the API test running framework ensure each API test runs with PWD set to a unique temp dir. Same as we do for the ksh non-API tests.  This also allows us to simplify these tests since the test no longer has to worry about removing file system artifacts they create. The test runner is responsible for that cleanup.

PR #1256, opened 2019-03-28
API tests depend on POSIX shell script (and not ksh)

siteshwar:
Related: #1243

### DIFF
--- a/src/lib/libast/tests/run_test.sh
+++ b/src/lib/libast/tests/run_test.sh
@@ -1,3 +1,4 @@
+#!/bin/sh
 # This is loosely based on the src/cmd/ksh93/tests/util/run_test.sh script.
 #
 # This helps ensure each API test runs in a unique temp dir that is automatically cleaned up when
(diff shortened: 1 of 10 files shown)


siteshwar:
@krader1961 The script that you wrote for running API tests is not POSIX compliant. Tests are failing on FreeBSD with this error:

```
 default: /home/vagrant/ast/src/lib/libast/tests/run_test.sh: function: not found
 default: /home/vagrant/ast/src/lib/libast/tests/run_test.sh: typeset: not found
 default: /home/vagrant/ast/src/lib/libast/tests/run_test.sh: ${@:2}: Bad substitution
```

src/lib/libast/tests/run_test.sh
```diff
@@ -6,22 +6,20 @@
 #
 # Make sure when called via `meson test` we've got the expected args.
 #
-function log_info {
-    typeset lineno=$1
-    print -r "<I> run_test[$lineno]: ${@:2}"
+log_info() {
+    echo "<I> run_test[$1]: $2"
```
krader1961: FWIW this isn't equivalent to the original version which allows the message to be passed as multiple words. This version requires the message to be quoted in order to be passed as a single arg. In practice this probably doesn't matter because we should probably always be quoting the message so it is a single word. My thought was that if someone did write

```
log_info this is the message
```

rather than

```
log_info "this is the message"
```

it should still display the entire message.

krader1961:
LGTM

krader1961:
Also, you'll probably want to remove the `ksh` package from the various docker configs in *.travis.yml*.

siteshwar:
> Also, you'll probably want to remove the `ksh` package from the various docker configs in _.travis.yml_.

I already did.

PR #1257, opened 2019-03-29
Fix a heap-buffer-overflow bug

krader1961:
If the length of string `xp` is greater than the length of string `cp`
the expression will access beyond the buffer holding `cp`.

Also, it is clear the condition is meant to only be true if the strings
compare equal in the first `xlen` chars but it is true if they are not
equal! Too, as best I can determine the body of that `if()` is never
executed; either with the original or corrected condition.

### DIFF
--- a/src/cmd/ksh93/sh/name.c
+++ b/src/cmd/ksh93/sh/name.c
@@ -889,7 +889,7 @@ Namval_t *nv_create(const char *name, Dt_t *root, int flags, Namfun_t *dp) {
                         flags &= ~(noscope ? 0 : NV_NOSCOPE);
                     } else if (c) {
                         char *xp = NULL;
-                        ssize_t xlen;
+                        size_t xlen;
                         c = (cp - sp);
                         // Eliminate namespace name.
                         if (shp->last_table && !nv_type(shp->last_table)) {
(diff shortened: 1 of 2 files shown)


src/cmd/ksh93/sh/name.c
```diff
@@ -889,16 +889,18 @@ Namval_t *nv_create(const char *name, Dt_t *root, int flags, Namfun_t *dp) {
...
                         c = (cp - sp);
                         // Eliminate namespace name.
                         if (shp->last_table && !nv_type(shp->last_table)) {
                             xp = nv_name(shp->last_table);
                             xlen = strlen(xp);
                         }
                         cp = nv_name(np);
-                        if (xp && strncmp(cp, xp, xlen) && cp[xlen] == '.') cp += xlen + 1;
                         copy = strlen(cp);
+                        if (xp && copy > xlen && !strncmp(cp, xp, xlen) && cp[xlen] == '.') {
```
siteshwar: You changed `strncmp(cp, xp, xlen)` to `!strncmp(cp, xp, xlen)`. Why ?

krader1961: Because the intent is clearly to test if `xp` is a prefix, that is equal to the first `xlen` chars, of `cp`. The original code tested if they were not equal.

P.S., I instrumented the code and there isn't a single test that causes the body of that `if()` to be executed regardless of whether it is a string equality or inequality test.

PR #1258, opened 2019-03-30
slowread() fails to handle errors from readf()

krader1961:
Something has changed on macOS server and/or this project. The `set.exp`
unit test is now failing consistently on line 52 with a SIGBUS. I
instrumented the code and it's because ksh is running slowread() at that
point and it fails to handle a -1 return value from the readf function.
When it fails while running that unit test errno == ENOTSOCK. The
problem is because it uses the -1 (assigned to rsize which was unsigned)
as an index into `buff[]`. Program goes boom due to invalid address.

So handle failures of the underlying read function.


siteshwar:
LGTM

PR #1262, opened 2019-04-01
Fix ASAN heap-buffer-overflow bug in lex.c

### LINKED ISSUE #1261: ASAN heap-buffer-overflow in lex.c when running variables.sh unit test
The `poplevel()` in this block of code is called when `lp->lexd.lex_max` is zero:

https://github.com/att/ast/blob/bd94eb56c30dbeb1402d2a8642050f7fe9188175/src/cmd/ksh93/sh/lex.c#L921-L925

Since `poplevel()` first decrements `lp->lexd.lex_max` then uses it as an index into `lp->lexd.lex_match` this causes the word before the start of that buffer to be accessed. The buffer is allocated here:

https://github.com/att/ast/blob/bd94eb56c30dbeb1402d2a8642050f7fe9188175/src/cmd/ksh93/sh/lex.c#L2210-L2218

This is yet another good example for why I feel pretty strongly that more `assert()` statements are needed in this code.

TBD whether this is a logic bug (e.g., a missing `pushlevel()`) or whether `poplevel()` should simply be a no-op if there are no pushed levels.

krader1961:
There are some syntax errors that cause `poplevel()` to be called twice.
Which is obviously incorrect since it pops one too many levels and can
cause a `lp->lexd.lex_match[-1]` dereference to occur.

Fixes #1261

### DIFF
--- a/src/cmd/ksh93/sh/lex.c
+++ b/src/cmd/ksh93/sh/lex.c
@@ -57,8 +57,7 @@
 
 #define oldmode(lp) (lp->lexd.lastc >> CHAR_BIT)
 #define endchar(lp) (lp->lexd.lastc & 0xff)
-#define setchar(lp, c) (lp->lexd.lastc = ((lp->lexd.lastc & ~0xff) | (c)))
-#define poplevel(lp) (lp->lexd.lastc = lp->lexd.lex_match[--lp->lexd.level])
+#define setchar(lp, c) lp->lexd.lastc = ((lp->lexd.lastc & ~0xff) | (c))
 
 static_fn char *fmttoken(Lex_t *, int, char *);
 #ifdef SF_BUFCONST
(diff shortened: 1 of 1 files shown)


siteshwar:
LGTM

src/cmd/ksh93/sh/lex.c
```diff
@@ -57,8 +57,7 @@
 
 #define oldmode(lp) (lp->lexd.lastc >> CHAR_BIT)
 #define endchar(lp) (lp->lexd.lastc & 0xff)
-#define setchar(lp, c) (lp->lexd.lastc = ((lp->lexd.lastc & ~0xff) | (c)))
-#define poplevel(lp) (lp->lexd.lastc = lp->lexd.lex_match[--lp->lexd.level])
+#define setchar(lp, c) lp->lexd.lastc = ((lp->lexd.lastc & ~0xff) | (c))
```
dannyweldon: Not sure why the brackets have been removed from this macro (setchar).  I thought that it was a good idea to surround macros with brackets to prevent side-effects?

kdudka: Yes, it was.

krader1961: @dannyweldon Broadly speaking there are two types of macros: a) those that return a rvalue and can be used in an expression and b) those meant to be used like a void returning function call. The `setchar()`, `poplevel()` and `pushlevel()` are type b macros. It is a mistake to use them in an expression such as `if (x && setchar())`. The parens around a type b macro are irrelevant. In fact, traditionally such macros are written using this pattern to ensure that if they are used in an expression it will fail to compile:

```
#define setchar(lp, c) do { something; } while (0)
```

Today I would write that as a `static inline void` function. I already changed `poplevel()` and `pushlevel()` to functions and intended to do the same to `setchar()` but got sidetracked. Thanks for the reminder to finish that cleanup.

P.S., Yes, macros are used in other ways as well. For example, it's less common but sometimes a macro is meant to be used as an lvalue. There are other highly stylized patterns where a macro can't be substituted with a static inline function but you also don't want it to be usable in an expression. See, for example, `SFMTXRETURN`.

dannyweldon: Wow.  Thanks for the detailed explanation, @krader1961.

kdudka: > The parens around a type b macro are irrelevant.

They are not irrelevant in general.  When such macros are used in code, they look like function calls.  If you remove the brackets, such seeming function calls may have completely unexpected side effects in certain cases.  By removing the protection that was already there, you can shoot yourself in the foot later on.

PR #1280, opened 2019-04-16
Update fallback version number to `2020.0.0-alpha1`

siteshwar:
When git tags are not available, fallback to `2020.0.0-alpha1` version.


krader1961:
LGTM but I'm curious why the build can't simply do `meson -Dfallback-version-number=2020.0.0-alpha1`?

siteshwar:
> LGTM but I'm curious why the build can't simply do meson -Dfallback-version-number=2020.0.0-alpha1?

Sure, that's an option. But it's better if it works by default.

PR #1290, opened 2019-04-25
Simplify struct Enum initialization

krader1961:
Stop initializing a struct Enum using a single compound allocation.
Instead do it in a more straightforward manner. This not only makes the
logic easier to understand it makes it more likely that tools like ASAN
will detect invalid buffer accesses (e.g., buffer overruns).

Related #1002


siteshwar:
LGTM

PR #1305, opened 2019-05-06
Add script that is used for releasing new versions of ksh

siteshwar:
This script creates tarballs and signs them with user's gpg key.

### DIFF
--- a/bin/upload-release.sh
+++ b/bin/upload-release.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+set -x
+SELF="$0"
+TAG="$1"
+TOKEN="$2"
+
+NV="ksh-${TAG}"
+
+usage() {
+    printf "Usage: %s TAG TOKEN\n" "$SELF" >&2
+    exit 1
+}
+
+die() {
+    printf "%s: error: %s\n" "$SELF" "$*" >&2
+    exit 1
+}
+
+# check arguments
+test "$TAG" = "$(git describe --tags "$TAG")" || usage
+test -n "$TOKEN" || usage
+
+# dump a tarball
+SRC_TAR="${NV}.tar"
+git archive --prefix="$NV/" --format="tar" HEAD -- . > "$SRC_TAR" \
+                                        || die "failed to export sources"
+
+# produce .tar.gz
+TAR_GZ="${NV}.tar.gz"
+gzip -c "$SRC_TAR" > "$TAR_GZ"          || die "failed to create $TAR_GZ"
+
+# produce .tar.xz
+TAR_XZ="${NV}.tar.xz"
+xz -c "$SRC_TAR" > "$TAR_XZ"            || die "failed to create $TAR_XZ"
+
+# sign the tarballs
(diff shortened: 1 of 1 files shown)


scripts/upload-release.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
krader1961: This should probably go in the *bin* directory since it's meant to be run by a human rather than indirectly via `meson`.

scripts/upload-release.sh
```diff
@@ -0,0 +1,82 @@
...
+set -x
+SELF="$0"
+TAG="$1"
+TOKEN="$2"
+
+NAME="ksh"
```
krader1961: This var isn't used so either there is someplace it was meant to be used or it should be removed.

bin/upload-release.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+set -x
+SELF="$0"
+TAG="$1"
+TOKEN="$2"
```
krader1961: Consider renaming this GITHUB_AUTH_TOKEN (also in the usage message below).

bin/upload-release.sh
```diff
@@ -0,0 +1,82 @@
...
+# sign the tarballs
+for file in "$TAR_GZ" "$TAR_XZ"; do
+    gpg --armor --detach-sign "$file" || die "tarball signing failed"
+    test -f "${file}.asc" || die "tarball signature was not created"
+done
+
+# I did these steps manually during ksh-2020.0.0 alpha release.
+# TODO: Find out why these API calls fail 
+# file to store response from GitHub API
+# JSON="./${NV}-github-relase.js"
```
krader1961: s/relase/release/

bin/upload-release.sh
```diff
@@ -0,0 +1,82 @@
...
+NAME="ksh"
+NV="ksh-${TAG}"
+
+usage() {
+    printf "Usage: %s TAG TOKEN\n" "$SELF" >&2
+    exit 1
+}
+
+die() {
+    printf "%s: error: %s\n" "$SELF" >&2
```
krader1961: s/"$SELF"/"$SELF" "$*"/

bin/upload-release.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+set -x
+SELF="$0"
+TAG="$1"
+TOKEN="$2"
+
+NAME="ksh"
```
krader1961: This var isn't used.

bin/upload-release.sh
```diff
@@ -0,0 +1,82 @@
...
+# curl "https://api.github.com/repos/att/ast/releases" \
+#     -o "$JSON" --fail --verbose \
+#     --header "Authorization: token $TOKEN" \
+#     --data '{
+#     "tag_name": "'"$TAG"'",
+#     "target_commitish": "master",
+#     "name": "'"$NV"'",
+#     "draft": false,
+#     "prerelease": true
+# }' || exit $?
```
krader1961: Why not call `die` here?

PR #1310, opened 2019-05-14
Add description of a hook that fixes code styling

siteshwar:
in contribution guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -375,6 +375,14 @@ circumstances it might be advisable to circumvent it with `git push
 
 To install the hook, put it in .git/hooks/pre-push and make it executable.
 
+To fix code styling issues before making a commit, add this script as a pre-commit hoook :
+
+```sh
+#!/bin/sh
+
+SKIP_UNCOMMITTED_CHECK=1 bin/style all
+```
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -375,6 +375,14 @@ circumstances it might be advisable to circumvent it with `git push
 
 To install the hook, put it in .git/hooks/pre-push and make it executable.
 
+To fix code styling issues before making a commit, add this script as a pre-commit hoook :
+
+```sh
+#!/bin/sh
+
+SKIP_UNCOMMITTED_CHECK=1 bin/style all
```
krader1961: The `bin/style` command deliberately refuses to let you restyle the entire code base when uncommitted changes are present. That is to avoid mixing unrelated style cleanups with a commit. We really don't want to change that behavior.

A simple `bin/style` is all that is needed. That will fix any style mistakes in the files being committed. However, that is not sufficient by itself. That's because the current implementation deliberately doesn't do a `git add` to stage any modified files. So the commit will end up not including the restyled files. It will only include the original, incorrectly styled, change. So what we want here is not an option to bypass the check for uncommitted changes but an option to force any style fixes to be added to the files staged to be committed.

PR #1311, opened 2019-05-14
git hook for fixing style problems at commit time

krader1961:
Document how to create a git hook for fixing style problems at commit time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -373,7 +373,17 @@ will run `meson test` and only allow the push if that succeeds. In some
 circumstances it might be advisable to circumvent it with `git push
 --no-verify`, but usually that should not be necessary.
 
-To install the hook, put it in .git/hooks/pre-push and make it executable.
+To install the hook, put it in *.git/hooks/pre-push* and make it
+executable.
+
+To fix code styling issues before making a commit, add this
+script as a pre-commit hoook; that is in an executable file named
+*.git/hooks/pre-commit*:
+
+```sh
+#!/bin/sh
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -373,7 +373,17 @@ will run `meson test` and only allow the push if that succeeds. In some
...
+To install the hook, put it in *.git/hooks/pre-push* and make it
+executable.
+
+To fix code styling issues before making a commit, add this
+script as a pre-commit hoook; that is in an executable file named
+*.git/hooks/pre-commit*:
+
+```sh
+#!/bin/sh
+STAGE_STYLE_FIXUPS=1 bin/style
```
siteshwar: `STAGE_STYLE_FIXUPS=1 bin/style all` ?

siteshwar: May be we can add a git command here to find all files that are staged to commit ?

krader1961: > STAGE_STYLE_FIXUPS=1 bin/style all 

As I said in my comment on your PR we do not want to "style all" on each commit. Not because it's expensive (though it is) but because we don't want style fix-ups unrelated to a commit to be bundled with that commit.

> May be we can add a git command here to find all files that are staged to commit ?

That's what `bin/style` already does. If you edit a C module, introduce a style problem, and run `bin/style` is it not fixing the error in that file? Also, it should do the right thing whether or not you've already done `git add` to put the change in the staging area.

As an added bonus if you've already committed the change running `bin/style` will fix the files in the most recent. After which you can do `git commit --amend -a` if there were any style fix-ups.

PR #1314, opened 2019-05-20
Chmod

krader1961:
FWIW, I'm still not happy with the chmod builtin code. And I'm not 100% certain all the corner cases are handled correctly. But this fixes the major bogosities and, when compared to the platform `chmod` and the ksh93u+ builtin, this works correctly. At least with respect to all uses in the unit tests.


siteshwar:
LGTM

PR #1316, opened 2019-05-23
Stop dynamically generating conftab.c and conftab.h

### LINKED ISSUE #1315: Build failure with gcc-9.1.0
**Description of problem:**
Building ksh with gcc-9.1.0 on Gentoo Linux fails.

**Ksh version:**
5d692eaf53f8e50380d2c6961371e72f61950a4b

Build log attached.
[build.log](https://github.com/att/ast/files/3208681/build.log)

krader1961:
GCC 9.1 changes the output of `gcc -E` which breaks the conf.sh script.
We want to eliminate the `getconf` builtin and the `astconf()` function.
No unit tests depend on either of those. And it is highly unlikely any
real world scripts do so. So as the first step in removing those APIs and
to make building this possible with gcc 9.1 just use a static version of
conftab.c and conftab.h.

Fixes #1315
Related #1118


krader1961:
This builds cleanly (without warnings) and passes all unit tests on all platforms. Including OpenIndiana. It isn't ideal in so far as anyone using the `getconf` builtin might, but probably won't, get a result that incorrect in a manner that causes problems. However, since the probability anyone is using the ksh `getconf` builtin, let alone for something that is likely to be broken by this change, is so close to zero I think this is acceptable as a transition step to removing the command completely.

PR #1325, opened 2019-06-07
Remove uses of astconf()

krader1961:
This removes the remaining `astconf()` calls outside of the *getconf.c* module. It also introduces some new tests to verify that the `echo` command behaves correctly whether the universe, based on `$PATH`, should result in bsd or att behavior.


krader1961:
The forced push was because I realized I had not committed my fix to the `echo \c` tests. Before that update the bsd test for `echo \\c` behavior was trivial and did not actually test much of anything.

krader1961:
The forced push makes reviewing this too difficult. So closing it in favor of an easier to review PR.

PR #1327, opened 2019-06-09
Fix Broken time And times Built-ins

etscrivner:
Currently, the `time` and `times` built-ins result in an assertion failure
causing a crash. This is due to an erroneous assertion from commit
595f499709319d28ebbdbedcd69b2bff223e6f3a which attempted to fix a linter
warning about the usage of an uninitialized `timeval` variable `tb`.

This change removes the assertion and instead runs `memset` on `tb` in the case
where it would have been used without having been initialized. This fixes the
`cppcheck` linter warning and restores the `time` and `times` built-ins to
working order.

A simple set of tests is added to ensure that these built-ins do not break in
the future.

### DIFF
--- a/src/cmd/ksh93/sh/xec.c
+++ b/src/cmd/ksh93/sh/xec.c
@@ -2153,7 +2153,7 @@ int sh_exec(Shell_t *shp, const Shnode_t *t, int flags) {
                 job.waitall = 0;
             } else {
 #ifdef timeofday
-                tb.tv_sec = tb.tv_usec = 0;
+                memset(&tb, 0, sizeof(tb));
 #else
                 bt = 0;
 #endif  // timeofday
@@ -2163,7 +2163,6 @@ int sh_exec(Shell_t *shp, const Shnode_t *t, int flags) {
 #ifdef timeofday
             times(&after);
             timeofday(&ta);
-            assert(tb.tv_sec);
             at = shp->gd->lim.clk_tck * (ta.tv_sec - tb.tv_sec);
             at += ((shp->gd->lim.clk_tck *
                     (((1000000L / 2) / shp->gd->lim.clk_tck) + (ta.tv_usec - tb.tv_usec))) /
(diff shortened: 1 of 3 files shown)


krader1961:
@etscrivner Thanks for taking a stab at fixing this but the fix isn't correct. The `memset()` is equivalent to `tb.tv_sec = tb.tv_usec = 0;`. All you've done to "fix" this is remove the assert. The assert is meant to ensure that `tb` has been initialized to a non-zero value so that the calculation of `at` in the statements following the assert produces a meaningful result. The question is whether `tb` should be initialized to zero (and the assert removed) or should it be initialized via `timeofday(&tb)`.

Simply removing the assert and adding `DPRINTF("at %lld", (long long)at);` after the assignment to `at` shows that it has an absurdly large value when `tb` is initialized to zero. If instead of initializing it to zero we do `timeofday(&tb);` then the calculation of `at` produces a much more reasonable answer; i.e., zero. So that appears to be the correct fix.

Also, simply running `time` to ensure it doesn't blow up isn't a particularly useful test. Because you can't capture the output of `time` using simple redirection the only viable way to test it is to write an "expect" based unit test.

krader1961:
Also, we now always `#define timeofday` so all those `#ifdef timeofday` can be eliminated. If we ever find ourselves needing `timeofday()` to work on a platform without `gettimeofday()` we'll probably want to create a fallback implementation that does something reasonable using whatever mechanism is available for getting the current wall clock time with high precision.

krader1961:
Sigh! And, of course, the fact that `tb` wasn't initialized in one branch prior to commit 595f499 appears to be harmless because the value of `at` based on it isn't used in that case. Which is exhibit 666 for how badly this code is structured.

etscrivner:
@krader1961 Thanks for the review, if you don't mind I'd like to take another stab at fixing this since the functionality is currently completely broken. It sounds like the fix here is twofold:

1. Remove `#ifdef timeofday` checks.
3. Initialize `tb` with `timeofday(&tb)`.

Definitely understand how having a test that checks functionality isn't completely broken seems minimally useful, but in this particular case the lack of such a test definitely allowed this functionality to remain broken for over a year. Happy to add more "correct" seeming tests. Do we currently do any `expect` based on testing that I could base this on?

krader1961:
@etscrivner Please review #1329

Issue #1328, opened 2019-06-09
Replaces uses of timeofday() with 


krader1961:
While reviewing PR #1327 I noticed that we now assume `gettimeofday()` is available. The code that currently uses the `timeofday()` macro should be using `tvgettime()` from libast; i.e., *src/lib/libast/tm/tvgettime.c*.

krader1961:
There is only one use of `tvgettime()`: in `tvtouch()`. And `tvtouch()` is only used one place: `hist_touch()`. We're probably better off eliminating that (indirect) use of `tvgettime()` rather than adding additional uses.

PR #1329, opened 2019-06-09
Fix the `time` keyword

krader1961:
Commit 595f499 by me introduced an assert and initialized var `tb`
in a path where it was used without being initialized. However, the
initialization to zero and the subsequent assert it was non-zero was
incorrect. This fixes that and adds a unit test to verify `time` and
`times` produce reasonable output.

Related #1328


etscrivner:
LGTM, thanks for fixing and adding more comprehensive tests :pray:

PR #1332, opened 2019-06-12
Replace `times` alias with builtin

### LINKED ISSUE #16: 'times' should be a special builtin
In ksh93, times is implemented as an alias to "{ { time;} 2>&1;}" and "command" as an alias to "command " (which means aliases are expanded after it).

That means that things like:

```
$ ksh93 -c 'LC_ALL=C times'
ksh93: syntax error at line 1: `{' unexpected
$ ksh93 -c 'command times'
ksh93: syntax error at line 1: `{' unexpected
```

Don't work, so the times utility is not POSIX compliant.

"times" should be implemented as a special builtin.

More generally, even though allowed by POSIX, implementing builtin utilities as aliases is not a good idea IMO if only for the reasons detailed at:
http://thread.gmane.org/gmane.comp.standards.posix.austin.general/12485/focus=12568

krader1961:
Fixes #16

### DIFF
--- a/features/meson.build
+++ b/features/meson.build
@@ -83,6 +83,10 @@ if feature_data.get('_lib_lchmod', 0) == 0
     endif
 endif
 
+feature_data.set10('_lib_getrusage',
+    cc.has_function('getrusage',
+        prefix: '\n'.join(['#include <sys/resource.h>', '#include <sys/time.h>']),
+        args: feature_test_args))
 feature_data.set10('_lib_sigqueue',
     cc.has_function('sigqueue', prefix: '#include <signal.h>', args: feature_test_args))
 # TODO: Enable iswprint() detection when we understand why doing so causes the
--- a/src/cmd/ksh93/bltins/times.c
+++ b/src/cmd/ksh93/bltins/times.c
@@ -1,0 +1,74 @@
+//
+// `times` builtin command
+//
+#include "config_ast.h"  // IWYU pragma: keep
+
+#include <errno.h>
+#include <math.h>
+#include <string.h>
+#include <sys/times.h>
+
+#include "builtins.h"
+#include "defs.h"
+#include "error.h"
+#include "option.h"
+#include "sfio.h"
+#include "shcmd.h"
+
+int b_times(int argc, char *argv[], Shbltin_t *context) {
+    Shell_t *shp = context->shp;
+    const char *cmd = argv[0];
+
+    while ((argc = optget(argv, sh_opttimes))) {
+        switch (argc) {
+            case ':': {
+                errormsg(SH_DICT, 2, "%s", opt_info.arg);
+                break;
+            }
+            case '?': {
+                errormsg(SH_DICT, ERROR_usage(2), "%s", opt_info.arg);
+                __builtin_unreachable();
+            }
+            default: { break; }
+        }
+    }
+    if (error_info.errors) {
+        errormsg(SH_DICT, ERROR_usage(2), "%s", optusage(NULL));
+        __builtin_unreachable();
+    }
+
+    argv += opt_info.index;
+    if (*argv) {
+        errormsg(SH_DICT, 2, "%s: unexpected arguments", cmd);
+        errormsg(SH_DICT, ERROR_usage(2), "%s", optusage(NULL));
+        __builtin_unreachable();
+    }
+
+    struct tms cpu_times;
+    clock_t rv = times(&cpu_times);
+    if (rv == (clock_t)-1) {
+        errormsg(SH_DICT, ERROR_usage(2), "times() function unexpectedly failed: errno %d %s",
+                 errno, strerror(errno));
+        __builtin_unreachable();
+    }
+
+    double utime, stime, utime_min, utime_sec, stime_min, stime_sec;
+
+    utime = (double)cpu_times.tms_utime / shp->gd->lim.clk_tck;
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
+    stime = (double)cpu_times.tms_stime / shp->gd->lim.clk_tck;
+    stime_min = floor(stime / 60);
+    stime_sec = stime - stime_min;
+    sfprintf(sfstdout, "%dm%.2fs %dm%.2fs\n", (int)utime_min, utime_sec, (int)stime_min, stime_sec);
+
+    utime = (double)cpu_times.tms_cutime / shp->gd->lim.clk_tck;
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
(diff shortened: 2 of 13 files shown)


siteshwar:
@krader1961 Please hold until I review it.

krader1961:
FWIW, This took me just a smidge over two hours to throw together. I don't claim the code for calculating the CPU deltas is optimal from a performance perspective. But it is straightforward; that is, easy to understand. This code is not a "hot" path. That is, this code is not performance critical. It should be executed rarely. So wringing every last clock tick from its execution is not justified.

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+    clock_t rv = times(&cpu_times);
+    if (rv == (clock_t)-1) {
+        errormsg(SH_DICT, ERROR_usage(2), "times() function unexpectedly failed: errno %d %s",
+                 errno, strerror(errno));
+        __builtin_unreachable();
+    }
+
+    double utime, stime, utime_min, utime_sec, stime_min, stime_sec;
+
+    utime = (double)cpu_times.tms_utime / shp->gd->lim.clk_tck;
```
siteshwar: It may be a good idea to put a comment about user time and system time here.

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
+    stime = (double)cpu_times.tms_stime / shp->gd->lim.clk_tck;
+    stime_min = floor(stime / 60);
+    stime_sec = stime - stime_min;
+    sfprintf(sfstdout, "%dm%.2fs %dm%.2fs\n", (int)utime_min, utime_sec, (int)stime_min, stime_sec);
+
+    utime = (double)cpu_times.tms_cutime / shp->gd->lim.clk_tck;
```
siteshwar: It may be a good idea to put a comment about children user time and system time here.

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
```
siteshwar: Shall it be `utime % 60` ?

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+        errormsg(SH_DICT, ERROR_usage(2), "times() function unexpectedly failed: errno %d %s",
+                 errno, strerror(errno));
+        __builtin_unreachable();
+    }
+
+    double utime, stime, utime_min, utime_sec, stime_min, stime_sec;
+
+    utime = (double)cpu_times.tms_utime / shp->gd->lim.clk_tck;
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
```
siteshwar: Shall it be `utime % 60` ?

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+    stime = (double)cpu_times.tms_stime / shp->gd->lim.clk_tck;
+    stime_min = floor(stime / 60);
+    stime_sec = stime - stime_min;
```
siteshwar: Same as above ?

src/cmd/ksh93/bltins/times.c
```diff
@@ -0,0 +1,74 @@
...
+    sfprintf(sfstdout, "%dm%.2fs %dm%.2fs\n", (int)utime_min, utime_sec, (int)stime_min, stime_sec);
+
+    utime = (double)cpu_times.tms_cutime / shp->gd->lim.clk_tck;
+    utime_min = floor(utime / 60);
+    utime_sec = utime - utime_min;
+    stime = (double)cpu_times.tms_cstime / shp->gd->lim.clk_tck;
+    stime_min = floor(stime / 60);
+    stime_sec = stime - stime_min;
```
siteshwar: Same as above ?

src/cmd/ksh93/tests/time.exp
```diff
@@ -34,7 +34,7 @@ expect_prompt
 # The times command with no args produces reasonable output.
 log_test_entry
 send "times\r"
-expect -re "\r\nuser$zero_time_re\r\nsys$zero_time_re\r\n" {
+expect -re "\r\n$zero_time_re $zero_time_re\r\n$zero_time_re $zero_time_re\r\n" {
```
siteshwar: Can we add a test for non-zero times ?

src/cmd/ksh93/bltins/times.c
krader1961: okay

src/cmd/ksh93/bltins/times.c
krader1961: Can't use `%` because that only works for ints. While I could do all the arithmetic using int types that is more complicated.

src/cmd/ksh93/tests/time.exp
krader1961: I'd love to. The question is how to do so in a portable fashion that doesn't result in flaky test failures. Note that `zero_time_re` actually allows for a small, less than one second, non-zero value precisely because the shell might have managed to accumulate 10 ms or more of user or system mode time.

features/meson.build
```diff
@@ -83,6 +83,10 @@ if feature_data.get('_lib_lchmod', 0) == 0
     endif
 endif
 
+feature_data.set10('_lib_getrusage',
+    cc.has_function('getrusage',
```
siteshwar: On which systems this test will fail ?

src/cmd/ksh93/tests/time.exp
siteshwar: We don't need to check for specific times, just checking for non-zero value should be sufficient.

features/meson.build
krader1961: No idea if there are any systems we care about which don't provide `getrusage()`. This is just a bit of extra caution since the code did not use it prior to this change; and instead used the inferior `times()` API.

src/cmd/ksh93/tests/time.exp
krader1961: The problem is that for it to be meaningful we need to ensure the CPU time is always in a predictable range; e.g., [0.5, 0.9] or [1.0, 9.0]. Which, it just occurred to me, isn't hard to do. I can create a CPU bound loop that checks the `SECONDS` var to terminate the loop. That allows us to check that the user mode time for the shell and child processes is correctly reported. I can't think of a simple way to deterministically test the system mode time.

siteshwar:
LGTM

Issue #1337, opened 2019-06-24
Memory leak when autoloading functions are used in a loop


siteshwar:
**Description of problem:**
Memory leak when autoloading functions are used in a loop

**Ksh version:**
2020.0.0-alpha1-191-gd0e6263f

**How reproducible:**
Always

**Steps to reproduce:**
```
#!/bin/ksh

###case 3###
#function call via autoload
FPATH=.
autoload func1

for (( j=0; j<10000; j++)); do
    a=`func1`
    pmap -x $$ |tail -1
done
```

**Actual results:**
Memory usage increases in this example.

**Expected results:**
Memory usage should not increase.

**Additional info:**

Related bug: https://bugzilla.redhat.com/show_bug.cgi?id=921138 (internal)

krader1961:
I'm curious if this happens if `a=$(func1)` is used. I wouldn't expect that formulation to also leak memory but if it doesn't that fact may be a helpful. Also, the recreation steps are slightly ambiguous. Does it matter whether there is a autoloaded function named `func` in the CWD? If that file has to exist does it matter what the contents are?

How much memory is leaked each time through the loop? Is it four, eight, sixteen, more bytes? A four byte leak can probably be ignored (albeit still a bug). But a non-trivial amount of memory is a problem unless we're talking about the error case where the `func1` autoloaded file doesn't exist.

krader1961:
I put a file named *func1* in the CWD with this content and made it executable:

```
function func1 {
    :
}
```

That leaks ~256 bytes each time through the "for" loop in the problem reproducer script. It does not appear to leak any memory using the ksh93u+ binary in Fedora 28.

@siteshwar, Was this problem reported against ksh93u+ or the current git master branch?

krader1961:
I went to the trouble of building ksh93v- on my Fedora 28 virtual machine. It also leaks memory when running that test but at half the rate: ~128 bytes each time through the "for" loop. That factor of two difference in the size of the leak is almost certainly due to changes after the ksh93v- code was open sourced and which changed the size of a specific structure. Or possibly due to switching from AST vmalloc to the system malloc subsystem. Which would mean the leak was introduced by the AST team after the ksh93u+ release and before changes to the Github code by the OSS community. Sorry, but I'm a bit sensitive about this type of issue given the god like status of David Korn and the rest of the AST team  :smile:

[...]

PR #1338, opened 2019-06-24
Fix a memory leak in `regcomp_bra()` function

siteshwar:
Memory allocated for `cc` variable was never freed. This was causing
one of the Red Hat internal tests to fail.

Related: #1337

### DIFF
--- a/src/lib/libast/regex/regcomp.c
+++ b/src/lib/libast/regex/regcomp.c
@@ -1086,6 +1086,7 @@ static_fn Rex_t *regcomp_bra(Cenv_t *env) {
     unsigned char buf[4 * (COLL_KEY_MAX + 1)];
     int ic;
     char mbc[COLL_KEY_MAX + 1];
+    Cchr_t *cc_keys = 0;
 
     if (!(e = regcomp_node(env, REX_CLASS, 1, 1, sizeof(Set_t)))) return 0;
     collate = complicated = elements = 0;
@@ -1314,7 +1315,7 @@ static_fn Rex_t *regcomp_bra(Cenv_t *env) {
         dt = lc_collate_data;
         if (ast.locale.serial != lc_collate_serial || !dt) {
             disc.key = offsetof(Cchr_t, key);
-            cc = calloc(1, elementsof(primary) * sizeof(Cchr_t));
+            cc_keys = cc = calloc(1, elementsof(primary) * sizeof(Cchr_t));
             if (cc && (dt = dtopen(&disc, Dtoset))) {
                 for (i = 0; i < elementsof(primary) - 1; i++, cc++) {
                     cc->nam[0] = primary[i];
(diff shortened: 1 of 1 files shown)


src/lib/libast/regex/regcomp.c
```diff
@@ -1294,7 +1294,7 @@ static_fn Rex_t *regcomp_bra(Cenv_t *env) {
 
     if (complicated) {
         Dt_t *dt;
-        Cchr_t *cc;
+        Cchr_t *cc, *cc_begin = 0;
```
krader1961: I instrumented the code and it appears you dealt with all the leaks of this particular pointer. However, I'd like to propose a slightly different approach. The only reason you needed to introduce `cc_begin` is because `cc` is used for two (or more) different purposes. Which is a dangerous anti-pattern.  Don't introduce an alias. Instead define a new var at the top of the function that is initialized to NULL; perhaps named `cc_keys` rather than `cc_begin`. Then assign to that var and use it in the

```
    if (ast.locale.serial != lc_collate_serial || !dt) {
    }
```

block of code. Then simply call `free(cc_keys);` before each `return` in this function.

Also, notice that it appears the `if (dt) {` that immediately follows the preceding block is always true. Both based on analysis of the code and empirical testing by adding a `assert(dt);` just before the `if (dt) {`.

siteshwar: > The only reason you needed to introduce cc_begin is because cc is used for two (or more) different purposes. Which is a dangerous anti-pattern. Don't introduce an alias. Instead define a new var at the top of the function that is initialized to NULL; perhaps named cc_keys rather than cc_begin.

How's this approach different from my current approach ? I am freeing this variable at each return from this code block.

krader1961: My proposal simplifies the code on a conceptual level by eliminating the aliasing of the `cc` variable to mean different things. Also, the addition of a pointer on the stack does not matter from a performance standpoint in this specific use case. My approach means you only need to call `free(cc_keys);` before each `return ...;` rather than adding a lot of `free()` calls before `goto...;` statements. Etcetera.

Your solution fixes the problem. It also further enshrines an anti-pattern. When fixing a bug in the code we should also eliminate bad practices (i.e., anti patterns) when possible.

src/lib/libast/regex/regcomp.c
```diff
@@ -1410,7 +1415,9 @@ static_fn Rex_t *regcomp_bra(Cenv_t *env) {
                         case 0:
                             goto error;
                         case ':':
-                            if (env->flags & REG_REGEXP) goto complicated_normal;
+                            if (env->flags & REG_REGEXP) {
```
krader1961: It's okay if it remains this way but you could revert this to its original single line form. Also, several places below.

src/lib/libast/regex/regcomp.c
```diff
@@ -1314,7 +1315,7 @@ static_fn Rex_t *regcomp_bra(Cenv_t *env) {
         dt = lc_collate_data;
         if (ast.locale.serial != lc_collate_serial || !dt) {
             disc.key = offsetof(Cchr_t, key);
-            cc = calloc(1, elementsof(primary) * sizeof(Cchr_t));
+            cc_keys = cc = calloc(1, elementsof(primary) * sizeof(Cchr_t));
```
krader1961: Out of curiosity I inserted a `free(cc_keys);` after the `dtinsert(dt, cc);` a few lines below. I then ran all the unit tests under ASAN and everything works. ASAN did not detect any "use after free" errors. **Do not make that change** because I don't understand why everything works if the buffer is immediately freed after being inserted. According to the documentation a copy of the buffer is supposedly only made if the discipline defines a `makef` function which this code does not do. But I find it interesting that the buffer can be immediately freed without repercussion. There is something weird going on.

PR #1342, opened 2019-06-26
Refactor `fmtre()`

krader1961:
The current structure of `fmtre()` is hard to understand because it uses
a very unusual pattern. It also causes lint warnings about the final
`break;` at the bottom of the loop. Refactor the function to make the
code easier to understand.

### DIFF
--- a/src/lib/libast/string/fmtre.c
+++ b/src/lib/libast/string/fmtre.c
@@ -28,6 +28,7 @@
  */
 #include "config_ast.h"  // IWYU pragma: keep
 
+#include <stdbool.h>
 #include <string.h>
 
 #include "ast.h"
@@ -45,7 +46,7 @@ char *fmtre(const char *as) {
     Stack_t *p;
     char *x;
     int n;
-    int end = 1;
+    bool end = true;
     char *buf;
     Stack_t stack[32];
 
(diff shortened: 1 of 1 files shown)


siteshwar:
Lines changed in this patch are [not covered](http://situ.im/ast/coveragereport/index..._src_lib_libast_string_fmtre.c.html) by tests, can we add tests for them ?

src/lib/libast/string/fmtre.c
```diff
@@ -66,20 +67,24 @@ char *fmtre(const char *as) {
     } else {
         s++;
     }
-    while (1) {
+
+    while (*s) {
         c = *s++;
+        if (c == '*' && !*s) {
+            end = false;
+            break;
```
siteshwar: Was not this statement breaking out of `switch` statement ?

krader1961: Yes, but placed in the switch it relied on the `break;` at the bottom of the loop to break out of the loop. It also hid the fact that this was a special-case. Similar to how the `case 0: break;` obscured the fact that the loop is terminated when the end of the string is reached. Which is why I changed `while (1)` to `while (*s)` and dropped that case.

krader1961:
> Lines changed in this patch are not covered by tests, can we add tests for them ?

Actually the coverage is much better than reflected in your report because you built with `-Dbuild-api-tests=false build` which excludes the *src/lib/libast/tests/string/fmtre.c* API test of this function. Having said that the coverage is still unacceptably low so I'll try to improve it.

I think we should remove that option from the directions for producing a coverage report.

krader1961:
I added more tests and we now have 100% line and branch coverage of `fmtre()`. But note I wrote the tests to match the implementation rather than based on the documentation.

PR #1352, opened 2019-07-10
Remove math function `fpclassify()`.

### LINKED ISSUE #1346: Remove `fpclassify()` math function
I was adding some unit tests to improve test coverage of *src/cmd/ksh93/bltins/math.c* and noticed the `fpclassify()` function. That math function is completely unusable, at least in any portable fashion, because it returns small ints that correspond to symbols like `FP_NAN` in the *math.h* system header and those symbols are not available to a ksh script. So, for example, `print $(( fpclassify(1.0) ))` outputs `4` which, on my system, corresponds to `FP_NORMAL`. This is an even better example than those I documented in issue #88 as to why so many of the libc math functions should not be automatically exposed to ksh scripts just because doing so is trivial.

Cleaning this up by eliminating the truly useless math functions should be done after the next stable release. By "truly useless" I mean in the context of a ksh script not that these functions aren't useful to a high performance numeric computation app written in C/C++.

krader1961:
The math function `fpclassify()` is not portable since the values it
returns are not portable without exposing C symbols like `FP_NAN` from
the math.h header file.

This is technically a non-backward compatible change that should only
be done in a major release. However, the `fpclassify()` is unusable as
currently implemented; at least in any portable fashion. The probability
that any ksh script depends on `fpclassify()` is zero, or so close to
it, that removing it is a non-issue. Therefore removing this "feature"
should be harmless.

Fixes #1346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ was subsequently created to document the work being done.
 
 ## Notable non-backward compatible changes
 
+- The nonportable, unusable, `fpclassify` math function has been removed
+  (issue #1346).
 - The broken math functions `nextforward` and `nexttoward` have been removed
   (issue #1346).
 - Support for binary plugins written for ksh93u+ or earlier releases has been
--- a/src/cmd/ksh93/bltins/math.c
+++ b/src/cmd/ksh93/bltins/math.c
@@ -26,10 +26,6 @@ static int local_finite(Sfdouble_t a1) {
 
 static Sfdouble_t local_float(Sfdouble_t a1) { return a1; }
 
-static int local_fpclassify(Sfdouble_t a1) {
-    return fpclassify(a1);  //!OCLINT(constant conditional operator)
-}
-
 static Sfdouble_t local_int(Sfdouble_t a1) {
     if (a1 < LLONG_MIN || a1 > ULLONG_MAX) return 0.0;
     if (a1 < 0) return (Sfdouble_t)((Sflong_t)a1);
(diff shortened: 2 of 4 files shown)


krader1961:
@siteshwar Yes, this is a somewhat gratuitous change since there could be ksh programs that use math function `fpclassify()`. But they can't do so in portable manner. And given the coercion of ksh vars to type `Sfdouble_t` the result of calling this function is unreliable. That is, it is fundamentally, broken. Any ksh program that uses it is already broken or tied to a specific platform.

siteshwar:
I don't have a strong opinion about this change. May be it will be better to ask @cococlyde about it.

krader1961:
I would love to see feedback, even a thumbs down, from @cococlyde or anyone else with more historical context for why these bogosities exist. Sadly, that seems very unlikely. The bottom line is that when the old AST team decided (3+ decades ago) to implement wrappers around libm/libc functions like `sin()` they also wrapped all the other math functions. Seemingly without regard to whether they a) would be useful to a ksh program, and b) whether they could be meaningfully used given the ad-hoc casts between types like `float`, `double` and `long double` that the implementation performs and which is not under the control of the person writing a ksh program. Consider that a floating point value that is subnormal as a `long double` might become "normal" when case to a `double` and zero when cast to a `float`. These math functions are simply unusable in any meaningful sense in a ksh script. It looks like they implemented these functions simply because it looked trivial to do so given they had already created a framework for other functions such as `sin()` and `log()` that can be easily justified.

krader1961:
Merging since there has been adequate time for anyone with objections to voice them.

PR #1360, opened 2019-07-18
Add `glibc-all-langpacks` as a dependency for Fedora

siteshwar:
Recently Travis builds for Fedora broke because `glibc-all-langpacks`
was not installed on Fedora docker instance. Explicitly add it as a
dependency to build on Fedora.


krader1961:
LGTM

PR #1362, opened 2019-07-20
Check and link with musl-fts library for fts(3)

DanySpin97:


features/meson.build
```diff
@@ -52,6 +52,10 @@ libdl_dep = cc.find_library('dl', required: false, dirs: lib_dirs)
 # on others they are found in libexecinfo.
 libexecinfo_dep = cc.find_library('execinfo', required: false, dirs: lib_dirs)
 
+# On some platforms `fts` functions are found in libc,
+# on others they are found in musl-fts.
+libfts_dep = cc.find_library('fts', required: false, dirs: lib_dirs)
```
siteshwar: This library is actually required on Alpine linux, so `required: false` seems incorrect. We should check if we are running the build on Alpine linux. Probably check for presence of `/etc/alpine-release` file, or find a way to test for `musl` based systems.

DanySpin97: I think this is exactly the same as libexecinfo, as it is also required on musl. I would say that adding a header check (`fts.h` in this case) should work as intended.

krader1961: Agreed, @DanySpin97. I haven't used any distros that use Musl rather that GNU or BSD libc. I'm assuming there is no chance on such a system for glibc and musl to be installed at the same time. Correct? In other words, how likely is it that we'll need to add more feature tests related to the use of musl?

krader1961:
Merged

features/meson.build
DanySpin97: > I'm assuming there is no chance on such a system for glibc and musl to be installed at the same time. Correct?

Yea, only one can be installed at the same time.

> In other words, how likely is it that we'll need to add more feature tests related to the use of musl?

It seems that the other features required on musl are already there. I'd report back if everything else is needed for musl (especially if some test fail on musl and succeeded on glibc).

PR #1363, opened 2019-07-21
Check for and report I/O error in output builtins

### LINKED ISSUE #1093: Output builtins quietly ignore I/O errors
**Description of problem:**

The `echo`, `print` and `printf` builtins in ksh93 completely ignore I/O errors.

One ramification of this is that there is no way to recover if SIGPIPE is ignored and a script is piped into something like `head`; the script may end up in an infinite loop with no way to know that output is no longer possible.

**Ksh version:**

All ksh93 versions.

**Steps to reproduce:**

    (trap "" PIPE; exec ksh -c 'while :; do echo hi || exit; done')|head -n 3

**Actual results:**

After the first three greetings, ksh quietly prints 'hi' into the great void forever.

**Expected results:**

ksh exits (due to the `|| exit` check) after `echo hi` throws an I/O error. For instance, dash does:

```
hi
hi
hi
dash: 1: echo: echo: I/O error
```

and bash does:

```
hi
hi
hi
bash: line 0: echo: write error: Broken pipe
```

**Additional info:**

The ksh93 behaviour is not shared by any other shell that I know of; on bash, (d)ash, yash, zsh, pdksh/mksh, and related variants, the builtins check for I/O errors.

POSIX specifies that a message should be printed to standard error and the exit status should be > 0 if an error occurs in `echo` or `printf`. Surely a failure to write output is an error.

McDutchie:
This allows scripts to check for a nonzero exit status on the `print`, `printf` and `echo` builtins and prevent possible infinite loops if `SIGPIPE` is ignored.

`sfsync()` was already returning a negative value on I/O error, so all we need to do is add a check. The stream buffer will need to be filled before an I/O error can be detected, but this is the same on
other shells. See manual page: `src/lib/libast/man/sfio.3`

Fixes #1093

`src/cmd/ksh93/bltins/print.c`: `b_print()`:
- Make sure an error result from `sfsync()` is reflected in the output builtin's exit status (`exitval`).
- Write an I/O error message (`e_io`) if the exit status is nonzero.

`src/cmd/ksh93/data/msg.c`, `src/cmd/ksh93/include/io.h`:
- Add the `e_io[]` error message.

`src/cmd/ksh93/tests/b_print.sh`:
- Add I/O error regression test, checking for the correct error message and exit status. All three output builtins use the same `b_print()` function so we only need to test one.

### DIFF
--- a/src/cmd/ksh93/bltins/print.c
+++ b/src/cmd/ksh93/bltins/print.c
@@ -375,9 +375,10 @@ printv:
         // with syncing history if -s and -f are used together. History is synced
         // later with histflush() function.
         // https://github.com/att/ast/issues/425
-        if (!sflag && sffileno(outfile) != sffileno(sfstderr)) sfsync(outfile);
+        if (!sflag && sffileno(outfile) != sffileno(sfstderr))
+            if (sfsync(outfile) < 0) exitval = 1;
         sfpool(sfstderr, pool, SF_WRITE);
-        exitval = pdata.err;
+        if (pdata.err) exitval = 1;
     } else if (vflag) {
         while (*argv) {
             fmtbase64(shp, outfile, *argv++, fmttype, vflag == 'C');
@@ -386,7 +387,7 @@ printv:
     } else {
         // Echo style print.
         if (nflag && !argv[0]) {
-            sfsync(NULL);
+            if (sfsync(NULL) < 0) exitval = 1;
         } else if (sh_echolist(shp, outfile, rflag, argv) && !nflag) {
             sfputc(outfile, '\n');
         }
@@ -399,8 +400,9 @@ printv:
         sh_offstate(shp, SH_HISTORY);
     } else if (n & SF_SHARE) {
         sfset(outfile, SF_SHARE | SF_PUBLIC, 1);
-        sfsync(outfile);
+        if (sfsync(outfile) < 0) exitval = 1;
     }
+    if (exitval) errormsg(SH_DICT, 2, e_io);
     return exitval;
 }
 
--- a/src/cmd/ksh93/data/msg.c
+++ b/src/cmd/ksh93/data/msg.c
@@ -76,6 +76,7 @@ const char e_access[] = "permission denied";
 const char e_direct[] = "bad directory";
 const char e_file[] = "%s: bad file unit number";
 const char e_redirect[] = "redirection failed";
+const char e_io[] = "I/O error";
 const char e_trap[] = "%s: bad trap";
 const char e_readonly[] = "%s: is read only";
 const char e_badfield[] = "%d: negative field size";
--- a/src/cmd/ksh93/tests/b_print.sh
+++ b/src/cmd/ksh93/tests/b_print.sh
@@ -104,3 +104,24 @@ then
 else
     log_error 'print -s -f fails'
 fi
+
+# ======
+# Check that I/O errors are detected <https://github.com/att/ast/issues/1093>
+actual=$(
+    {
+        (
+            trap "" PIPE
+            for ((i = SECONDS + 1; SECONDS < i; )); do
+                print hi || {
+                    print $? >&2
+                    exit
+                }
+            done
+        ) | true
+    } 2>&1
+)
+expect=$': print: I/O error\n1'
+if [[ $actual != *"$expect" ]]
+then
+    log_error 'I/O error not detected' "$expect" "$actual"
+fi
--- a/src/cmd/ksh93/include/io.h
+++ b/src/cmd/ksh93/include/io.h
@@ -85,6 +85,7 @@ extern const char e_tmpcreate[];
 extern const char e_exists[];
 extern const char e_file[];
 extern const char e_redirect[];
+extern const char e_io[];
 extern const char e_formspec[];
 extern const char e_badregexp[];
 extern const char e_open[];


McDutchie:
Huh, why did the Travis CI build fail my regression test on macOS (and only macOS)? It works fine on my own Mac.

Could that build system be broken? It's got a warning about an outdated Xcode version.

krader1961:
Your test also succeeds on my macOS system and all my other VMs, including OpenIndiana (Solaris). The first thing to do is include the actual output received. At the end of the test:

```
actual=$(...)
expect=$': print: I/O error\n1'
[[ $actual == *"$expect" ]] || log_error 'I/O error not detected' "$expect" "$actual"
```

When I did that and pushed my branch the Travis execution succeeded. Which implies a race condition. The Travis macOS VMs have always been hideously oversubscribed and overloaded.

krader1961:
Looking at your test I can see it is inherently racy and thus likely to fail on an overloaded CI environment. I restarted your failing test and it passed. I would have written that test differently to avoid the race while still bounding the run time. But the simplest fix is to insert a `sleep 0.01` before the print in the inner loop. That should ensure the `head -n1` has exited when the first print is done.

krader1961:
If you change

```
) | head -n 1 >/dev/null
```

to

```
) | { sleep 1; head -n 1; } >/dev/null
```

the test will fail 100% of the time.

src/cmd/ksh93/bltins/print.c
```diff
@@ -375,9 +375,10 @@ int b_print(int argc, char *argv[], Shbltin_t *context) {
         // with syncing history if -s and -f are used together. History is synced
         // later with histflush() function.
         // https://github.com/att/ast/issues/425
-        if (!sflag && sffileno(outfile) != sffileno(sfstderr)) sfsync(outfile);
+        if (!sflag && sffileno(outfile) != sffileno(sfstderr))
+            exitval |= (sfsync(outfile) < 0);
         sfpool(sfstderr, pool, SF_WRITE);
-        exitval = pdata.err;
+        exitval |= pdata.err;
```
krader1961: I dislike this pattern since `exitval` is not a bitmask. Please change all such statements to the equivalent of `if (pdata.err) exitval = 1;`. This is the type of thing that might shave a few microseconds off the run time but cannot be justified. It makes bugs more likely. Not to mention reasoning about the code far harder. If this were a performance critical path I would allow it but that isn't the case.

McDutchie:
So it does. Yes, a race condition… that makes sense.

I wouldn't know how to write the test in a completely different way. I'm open to suggestions.

Unfortunately `sleep 0.01` is not portable (POSIX only specifies integers for `sleep`) and `sleep 1` just takes too long, but I think I found another solution. See 4a8d936.

src/cmd/ksh93/bltins/print.c
McDutchie: Very well. Done.

krader1961:
@McDutchie, You also wrote

> All three output builtins use the same  b_print() function so we only need to test one.

That is a reasonable conclusion in this instance. However, consider that unit tests tread a fine line between enough knowledge of the implementation to maximize their effectiveness and so little knowledge (other than project documentation or standards like POSIX) that they are wholly independent of the implementation. Whether applying this test to all three commands is justified is debatable.

You wrote

> Unfortunately sleep 0.01 is not portable

Yes, but in this context it's testing behavior unique to the ksh `sleep` builtin so should be acceptable. If this can be written in a POSIX compatible fashion that is also fine.

krader1961:
> I wouldn't know how to write the test in a completely different way. I'm open to suggestions.

Consider that ksh provides `$SECONDS`. If you assume it works as advertised then implementing an infinite print loop that exits if `$SECONDS` has rolled over is probably good enough for this situation.

McDutchie:
Oh. But now I've already pushed a commit with the `SECONDS` method. Let me know if you still think it's worth adding the `sleep`.

Weren't you going to get rid of all the extra builtins and just concentrate on ksh93, though? That's what README.md seems to suggest anyway. If `sleep` might not be a builtin in the future then portability is still a concern.

McDutchie:
Well, I'll be `#$$#@`. Travis CI macOS regress tests failed *again*.

This time I don't think it's my fault though. Timeout in `261/283 signal/shcomp`.

src/cmd/ksh93/bltins/print.c
```diff
@@ -375,9 +375,12 @@ int b_print(int argc, char *argv[], Shbltin_t *context) {
         // with syncing history if -s and -f are used together. History is synced
         // later with histflush() function.
         // https://github.com/att/ast/issues/425
-        if (!sflag && sffileno(outfile) != sffileno(sfstderr)) sfsync(outfile);
+        if (!sflag && sffileno(outfile) != sffileno(sfstderr))
+            if (sfsync(outfile) < 0)
```
siteshwar: I think you need to run `bin/style all` on your changes.

src/cmd/ksh93/data/msg.c
```diff
@@ -76,6 +76,7 @@ const char e_access[] = "permission denied";
 const char e_direct[] = "bad directory";
 const char e_file[] = "%s: bad file unit number";
 const char e_redirect[] = "redirection failed";
+const char e_io[] = "I/O error";
```
siteshwar: Can this message be made more descriptive to help the user ? Is it only shown on some specific type of I/O error ?

siteshwar:
@McDutchie You do not need to add separate commits for each review. Just amend the previous commit and do a force push. Ideally this pull request should be 1-2 commits only.

src/cmd/ksh93/tests/b_print.sh
```diff
@@ -104,3 +104,24 @@ then
 else
     log_error 'print -s -f fails'
 fi
+
+# ======
+# Check that I/O errors are detected (bug #1093)
```
siteshwar: We have several channels to track bugs (Red hat bugzilla, GitHub etc.). Also, bug tracking mechanisms may change over time. While mentioning bug id, it's good to have reference to actual platform (in this case GitHub).

src/cmd/ksh93/data/msg.c
McDutchie: It is shown when `sfsync()` returns an error exit status. The sfio.3 error page says `sfsync()` "returns a negative value for failure and 0 for success." It doesn't say how to get more specific error information out of it, so I have no idea how to make the error more specific.

McDutchie:
Sorry for all the commits – I always read that amending after pushing was basically the mortal sin of Git. Guess pull requests are an exception. I'll just merge everything into the first commit then.

siteshwar:
> Sorry for all the commits – I always read that amending after pushing was basically the mortal sin of Git. Guess pull requests are an exception. I'll just merge everything into the first commit then.

We never do forced pushes on `master` or any release branches. Those are protected. But it's fine to do forced pushes on pull requests to keep git history clean.

src/cmd/ksh93/bltins/print.c
McDutchie: Done.

src/cmd/ksh93/tests/b_print.sh
McDutchie: Done.

McDutchie:
> If you change
> 
> ```
> ) | head -n 1 >/dev/null
> ```
> 
> to
> 
> ```
> ) | { sleep 1; head -n 1; } >/dev/null
> ```
> 
> the test will fail 100% of the time.

So it does. But I now tried

```
) | { sleep 10; true; }
```

and that succeeds, at least on my system. Looks like using a builtin (and not forking a subshell) avoids the race condition.

krader1961:
FWIW, Whether you squash the commits and force push or just push when you add a commit to the branch is a judgement call and depends on which makes it easier to review the PR. When the PR is merged the person doing the merge will decide whether or not to squash the commits together if there is more than one.

krader1961:
Merged. Thx, @McDutchie.

PR #1364, opened 2019-07-21
Fix dynamic library for musl-fts

DanySpin97:
Upon further inspection after pull request #1362 , I've added the feature test for `fts.h` and fixed the dynamic library compilation.

### DIFF
--- a/src/lib/libast/meson.build
+++ b/src/lib/libast/meson.build
@@ -18,7 +18,7 @@ libast = library('ast', libast_files,
                  include_directories: [configuration_incdir, libast_incdir],
                  c_args: libast_c_args,
                  dependencies: [libm_dep, libiconv_dep, libcatgets_dep, libexecinfo_dep, libdl_dep,
-                                libsocket_dep, libnsl_dep, libfts_dep],
+                                libsocket_dep, libnsl_dep],
                  install: get_option('default_library') == 'shared')
 
 # This library exists solely to support libast unit tests so that `sh_getenv()` and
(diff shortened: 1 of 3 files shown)


features/meson.build
```diff
@@ -71,6 +71,7 @@ feature_data.set10('_hdr_sys_ldr', cc.has_header('sys/ldr.h', args: feature_test
 feature_data.set10('_hdr_sys_pstat', cc.has_header('sys/pstat.h', args: feature_test_args))
 feature_data.set10('_hdr_sys_ptem', cc.has_header('sys/ptem.h', args: feature_test_args))
 feature_data.set10('_hdr_sys_stream', cc.has_header('stream.h', args: feature_test_args))
+feature_data.set10('_hdr_fts', cc.has_header('fts.h', args: feature_test_args))
```
krader1961: You're not actually using this symbol so why is it needed? Also, we already `#include <fts.h>` where it is needed.

DanySpin97: Yea, this isn't needed, my bad. Should I add a header check for required include `fts.h`? Otherwise I'll leave only the dynamic library linking fix.

krader1961:
> Should I add a header check for required include fts.h? 

Probably not. We only need feature tests is the feature is optional, we provide a fallback if the platform doesn't provide it (e.g., `strlcpy()`), or where different platforms have different names for the same thing and we need to figure out which to use. None of those are applicable to *fts.h*.

DanySpin97:
Ah, I see, thanks for the explanation.

If there isn't anything else to fix, this PR is ready to be merged.

krader1961:
Merged

PR #1365, opened 2019-07-23
Install libksh.so when building shared library

DanySpin97:
This is the latest fix required to make ksh work when building shared library. Prior to this commit, the linking phase succeeded but at runtime it complains about libksh.so not being installed.

### DIFF
--- a/src/cmd/ksh93/meson.build
+++ b/src/cmd/ksh93/meson.build
@@ -24,7 +24,7 @@ libksh = library('ksh', ksh93_files,
                  include_directories: [configuration_incdir, ksh93_incdir],
                  c_args: shared_c_args,
                  dependencies: [libm_dep, libexecinfo_dep, libdl_dep, libsocket_dep, libnsl_dep],
-                 install: false)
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/meson.build
```diff
@@ -24,7 +24,7 @@ libksh = library('ksh', ksh93_files,
                  include_directories: [configuration_incdir, ksh93_incdir],
                  c_args: shared_c_args,
                  dependencies: [libm_dep, libexecinfo_dep, libdl_dep, libsocket_dep, libnsl_dep],
-                 install: false)
+                 install: true)
```
siteshwar: It should be built and linked statically. Are you building it as shared library ?

siteshwar: Sorry, I missed the commit message. This line should be written as:

```
install: get_option('default_library') == 'shared')
```

DanySpin97: Yea, you're right. Fixed.

siteshwar:
@DanySpin97 Thanks for the fix.

Issue #1367, opened 2019-07-28
functions fmtmatch() and fmtre() to convert between glob and ERE patterns are buggy


krader1961:
I recently added many more API tests of the `fmtre()` and `fmtmatch()` functions. Primarily to see if I could improve test coverage to acceptable levels. However, the API tests don't actually verify that the patterns are equivalent; only that the conversion produces the expected, but possibly incorrect, output. So I just added (not yet commited) tests to *src/cmd/ksh93/tests/sh_match.sh* to verify the patterns produce the same results when used in a shell script. Not surprisingly some patterns are incorrectly converted:

```
<E> sh_match[450]: pattern #13 ere |^x\!y$| and glob |x\!y| produce diff output
<E> sh_match[450]: pattern #20 ere |^x|y$| and glob |@(x|y)| produce diff output
<E> sh_match[450]: pattern #31 ere |^x\a|b$| and glob |@(xa|b)| produce diff output
<E> sh_match[493]: pattern #13 ere |^x\!y$| and glob |x!y| produce diff output
<E> sh_match[493]: pattern #21 ere |^x|y$| and glob |x|y| produce diff output
<E> sh_match[493]: pattern #34 ere |^z.?a$| and glob |~(E)z.?a| produce diff output
<E> sh_match[493]: pattern #38 ere |^x\a|b$| and glob |x\a|b| produce diff output
```

Notice that most of the failures involve patterns that use the alternation, `|`, operator. Consider the ERE `^x|y$`. That means match lines that begin with `x` or end with `y`. But the glob produced by `fmtmatch()` is `@(x|y)`. That means match lines that are one char in length where that char is `x` or `y`. In other words the generated glob is actually equivalent to the ERE `^(x|y)$`.
 
These functions are cool but I would never trust there output and therefore would never use them. If we were evaluating whether they should be added to the project I would have voted no. Sadly, even though it is highly unlikely any ksh script is using this feature we can't afford to break backward compatibility by removing it.

krader1961:
P.S., The way to use these in a script is to do something like this:

```
glob=$(print -f '%P' -- "$ere")
[[ "$line" = $glob ]] && print matches
ere=$(print -f '%R' -- "$glob")
[[ "$line" =~ $ere ]] && print matches
```

PR #1368, opened 2019-07-28
Test glob <=> ere conversion

krader1961:
When converting between globs and EREs it is a good idea to verify the
generated pattern matches the same lines as the original pattern.

Related #1367

### DIFF
--- a/src/cmd/ksh93/tests/sh_match.sh
+++ b/src/cmd/ksh93/tests/sh_match.sh
@@ -411,12 +411,121 @@ function test_nomatch
     [[ $got == "$expect" ]] || log_error 'unset submatches not handled correctly'
 }
 
-# run tests
+# ========
+# This is a regexp to extract pattern to pattern entries from related API unit test source. It emits
+# the patterns in a form this script can easily use; i.e., the two patterns separated by a tab.
+# See the next two test functions.
+p2p_sep=$'\t'
+p2p_re=$'/ tests\\[\\] =/,/NULL, NULL/s/^.*{"\\(.*\\)", "\\(.*\\)"},$/\\1\t\\2/p'
+# Generate a list of files from the source tree of the project. Then generate a file containing some
+# text from those files.
+find "$SRC_ROOT/src" -type f > glob2ere.files
+: > glob2ere.txt
+while read file
+do
+    sort --random-sort $file | head -5 >> glob2ere.txt
+done < glob2ere.files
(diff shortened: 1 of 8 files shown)


src/cmd/ksh93/tests/sh_match.sh
```diff
@@ -411,12 +411,131 @@ function test_nomatch
...
-# run tests
+# ========
+# This is a regexp to extract pattern to pattern entries from related API unit test source. It emits
+# the patterns in a form this script can easily use; i.e., the two patterns separated by a tab.
+# See the next two test functions.
+p2p_sep=$'\t'
+p2p_re=$'/ tests\\[\\] =/,/NULL, NULL/s/^.*{"\\(.*\\)", "\\(.*\\)"},$/\\1\t\\2/p'
+# These patterns are known not to be correctly converted to a glob. We still test them but only
+# issue a warning rather than failing the test. See https://github.com/att/ast/issues/1367.
+typeset -A ere2glob_blacklist
```
siteshwar: Definition of `ere2glob_blacklist` and `glob2ere_blacklist` can go in respective function definitions.

src/cmd/ksh93/tests/sh_match.sh
```diff
@@ -411,12 +411,131 @@ function test_nomatch
...
+glob2ere_blacklist['x\a|b']=1
+
+# ========
+# This is related to unit test API/string/fmtmatch. The difference is this uses ksh to convert the
+# pattern then uses ksh to verify both patterns match the same lines.
+function test_ere2glob {
+    # Generate a list of files from the source tree of the project.
+    # Then generate a file containing some text from those files.
+    find "$SRC_ROOT/src" -type f > ere2glob.files
+    : > ere2glob.txt
```
siteshwar: Why do you need a separate step to create this file ?

src/cmd/ksh93/tests/sh_match.sh
krader1961: I originally had them in their respective functions. I moved them here to emphasize how the two sets are related to each other. Bit of a coin toss as to which location is better.

src/cmd/ksh93/tests/sh_match.sh
krader1961: Because it's moderately expensive and both functions need some randomly generated text. I originally had this in each test function but when I noticed how expensive it is (especially on Cygwin) decided to hoist it to cut the cost in half.

PR #1374, opened 2019-08-04
Fix the coprocess.sh unit test

### LINKED ISSUE #1373: Bug in unit test coprocess.sh that manifests under code coverage
The first time I produced my own unit test code coverage report (just a few weeks ago) I noticed that the *coprocess.sh* unit test was failing with this error:

```
    coprocess.sh: line 237: process already exists
```

Which was perplexing because it doesn't fail under ASAN or non-coverage testing. Some judicious `DPRINTF()` and `set -x` plus `log_warning ...` statements showed the failure was here:

https://github.com/att/ast/blob/ed5c0f39e8ca62e5e7e8cda108b4194fbb150754/src/cmd/ksh93/tests/coprocess.sh#L93

The landmine that triggers that error is laid seven lines earlier in the unit test script:

https://github.com/att/ast/blob/ed5c0f39e8ca62e5e7e8cda108b4194fbb150754/src/cmd/ksh93/tests/coprocess.sh#L86

The problem is the script doesn't wait for the coprocess to exit. So, when the script tries to create the next coprocess, if the prior coprocess has not been reaped the attempt to create a new one fails. This is essentially a race condition that will succeed most of the time. The vast majority of the time the SIGCHLD sent to the coprocess will have caused it to exit and be reaped when the new coprocess is created. But when running with code coverage enabled things slow down enough that the coprocess has not been reaped when the next one is created.

krader1961:
The unit test had an inadvertent race condition involving reaping then
creating a coprocess. Fix that race. This also adds an explicit test of
the behavior that was causing random failures later in the unit test.

Fixes #1373


siteshwar:
LGTM

PR #1376, opened 2019-08-08
Clarify structure management in subshell.c

### LINKED ISSUE #1375: Comment about sizeof(struct Link) == sizeof(Dtlink_t) is wrong or the code is broken
While taking another look at issue #1002 I noticed this block of code:

https://github.com/att/ast/blob/8d91e8a2009ff9196454fea1362627fd1f921196/src/cmd/ksh93/sh/subshell.c#L61-L69

Note that, on systems where `sizeof(void *) == 8`, one struct is 32 bytes in size and the other is 16 bytes; i.e., four versus two pointers in size. This comment is also present in the source of the ksh93u+ release that is more than 12 years old. I would bet it goes back farther than that. So which is it? Do the two structures have to be the same size or is the comment bollocks?

krader1961:
There is much more to be done in this area. This is the bare mininum to
disentangle the allocation of `struct Link` and `Namval_t` structures in
`sh_assignok()` and downstream code. This makes the logic a lot more
obvious at the cost of an additional malloc()/free() pair of function
calls.

Related #1002
Fixes #1375


krader1961:
I've gone ahead and merged this.

PR #1384, opened 2019-08-24
Display the prompt on stdout rather than stderr

### LINKED ISSUE #1380: Interactive prompts should be written to stdout rather than stderr
While debugging a problem using an `expect` based unit test I noticed I couldn't redirect stderr without losing the prompt. That's because the prompt is written to stderr. This is problematic. Bash seems to have the same mistake but zsh does not. This means you can't do `exec 2>/tmp/stderr` in an interactive shell and still see a command prompt. This is problematic because debugging tools like the `DPRINTF()` macro I added long ago and mechanisms like ASAN write to stderr. You should be able to redirect stderr to a file to capture the output of such mechanisms while still having a usable interactive shell; i.e., a shell where the prompt appears on the stdout stream.

krader1961:
This makes it possible to redirect stderr and still have a usable
interactive shell. This is expected to be used rarely in the normal
course of using the shell, but is particularly valuable for unit
testing.

Fixes #1380

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ was subsequently created to document the work being done.
 
 ## Notable fixes and improvements
 
+- Interactive prompts are now written to stdout rather than stderr (issue #1380)
 - `declare` has been added as an alias for `typeset` (issue #220).
 - `local` has been added as a constrained alias for `typeset` when used inside
   a function (issue #220).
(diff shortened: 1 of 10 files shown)


siteshwar:
It should not be merged as it breaks POSIX compatibility ( https://github.com/att/ast/issues/1380#issuecomment-524862261 )

Issue #1391, opened 2019-09-01
Bare tilde expands to user name not home dir when $HOME is not set


floppym:
The 'builtins' test fails on Gentoo Linux.

[system-info.txt](https://github.com/att/ast/files/3563569/system-info.txt)

```
199/287 builtins                                FAIL     1.41 s (exit status 1)

--- command ---
TEST_ROOT='/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0-beta1/src/cmd/ksh93/tests' SHELL='/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/cmd/ksh93/ksh' LD_LIBRARY_PATH='/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/lib/libast:/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/lib/libcmd:/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/lib/libdll' LIBSAMPLE_PATH='/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/lib/libdll/libsample.so' SRC_ROOT='/x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0-beta1' /x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0_beta1-build/src/cmd/ksh93/ksh /x/portage/app-shells/ksh-2020.0.0_beta1/work/ksh-2020.0.0-beta1/src/cmd/ksh93/tests/util/run_test.sh builtins
--- stdout ---
<I> run_test[112]: TEST_DIR=/x/portage/app-shells/ksh-2020.0.0_beta1/temp/ksh.builtins.ntTgj6
--- stderr ---
<E> builtins[494]: cd with no arguments fails if HOME is unset
<W> builtins[-1]: error_count = 1
-------

200/287 builtins/shcomp                         FAIL     1.45 s (exit status 1)

--- command ---
[...]

PR #1396, opened 2019-09-04
Fix expansion of bare tildes when HOME is unset

floppym:
Use getpwuid to find the user name and home directory. Cache the user name for
future logins_tree lookups to avoid repeating the getpwuid call.

Fixes: https://github.com/att/ast/issues/1391

### DIFF
--- a/src/cmd/ksh93/sh/macro.c
+++ b/src/cmd/ksh93/sh/macro.c
@@ -2411,14 +2411,19 @@ static_fn void tilde_expand2(Shell_t *shp, int offset) {
 static_fn char *sh_tilde(Shell_t *shp, const char *string) {
     char *cp;
     int c;
-    struct passwd *pw;
+    struct passwd *pw = NULL;
     Namval_t *np = NULL;
     static Dt_t *logins_tree;
+    static char *username;
 
     if (*string++ != '~') return NULL;
     if ((c = *string) == 0) {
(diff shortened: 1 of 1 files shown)


floppym:
I do wonder if `getlogin()` is really the appropriate way to get the username here. Its manpage says:

>getlogin()  returns a pointer to a string containing the name of the user logged in on the controlling terminal of the process

If this is being used in a script, it's possible there is no controlling terminal, and we probably don't want that user in any case.

bash does this:

```
  entry = getpwuid (getuid ());
  if (entry)
    home_dir = savestring (entry->pw_dir);
```

Maybe ksh should use `getpwuid` to look up the username instead?

siteshwar:
A small note unrelated to this pull request: `variables` test on FreeBSD has been failing since we made a release last week. I will take a look at it.

floppym:
I force-pushed a new version that uses `getpwuid(getuid())` instead of `getlogin()`, and also caches the result.

floppym:
The travis failure seems completely unrelated.

krader1961:
LGTM. Do you feel comfortable adding a regression test? That is, a test that does something like `actual=$(env -u HOME $SHELL -c 'cd /; cd ~; pwd')` then compares the result to `$HOME`?

floppym:
Sure, I can work on that.

krader1961:
I added a unit test to avoid regressions and merged this PR.

PR #1404, opened 2019-09-17
run_test.sh: exit with status 77 when skipping a test

floppym:
meson test outputs "SKIP" in the log when it sees this status.

https://mesonbuild.com/Unit-tests.html#skipped-tests-and-hard-errors

### DIFF
--- a/src/cmd/ksh93/tests/util/postscript.sh
+++ b/src/cmd/ksh93/tests/util/postscript.sh
@@ -6,6 +6,6 @@ empty_fifos
 if (( $error_count != 0 ))
 then
     'log_warning' $((start_of_test_lineno - 1)) "error_count = $error_count"
-    exit $(( error_count < 125 ? error_count : 125 ))
+    exit $(( error_count < 50 ? error_count : 50 ))
 fi
 exit 0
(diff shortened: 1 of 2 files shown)


krader1961:
Please give it a symbolic name in the test preamble and use that rather than hardcoding a constant in the tests:

https://github.com/att/ast/blob/master/src/cmd/ksh93/tests/util/preamble.sh

Also, change the test postscript so that we don't inadvertently return that status:

https://github.com/att/ast/blob/master/src/cmd/ksh93/tests/util/postscript.sh

Capping the max error count to 50 should be fine.

floppym:
> Please give it a symbolic name in the test preamble and use that rather than hardcoding a constant in the tests:

preamble.sh isn't referenced until much later in run_test.sh. I don't think defining a symbolic name there would accomplish anything.

floppym:
Oh, I see. There are some specific test cases that also need to "skip". My patch only addresses those that exist early in run_test.sh itself.

krader1961:
Yeah, sorry for not being clearer. Skipping the io test used to be embedded in that test, not run_test.sh (not sure why it got moved). So rather than define it in preamble.sh make it an exported var in run_test.sh. Call it something like `MESON_SKIPPED_TEST`. Then we can use that symbol in run_test.sh and individual tests without causing people to go "Huh, what's with the magic 77".

floppym:
Force-pushed a new version.

krader1961:
Thanks, @floppym . Merged with the replacement of `exit 0` with `exit $MESON_SKIPPED_TEST` in the *directoryfd.sh* test.

PR #1413, opened 2019-10-09
Show version number through `ksh --version`

siteshwar:
Version number is stored in `SH_RELEASE` macro. Add it to the output of
`ksh --version`. This was a regression introduced by 44fb5d5a.


krader1961:
LGTM. I was just going to revert the five commits that externalized the builtin `optget()` strings since that work isn't essential for the next release but this is simpler.

PR #1415, opened 2019-10-09
Update changelog with stable version number

siteshwar:


krader1961:
Shouldn't this also update *meson_options.txt*? Otherwise, LGTM, although I'm ambivalent about adding a `-stable` suffix. Personally I think the lack of an explicit `alpha`, `beta`, or git commit ID suffix implies it is a stable release. Explicitly saying it is a stable version seems superfluous.

kdudka:
+1 for omitting the suffix on stable release(s)

siteshwar:
> +1 for omitting the suffix on stable release(s)

Fixed

siteshwar:
> Shouldn't this also update meson_options.txt? 

I will do that through a separate commit.

PR #1420, opened 2019-10-26
Fix `cd` function and make it automatically available

krader1961:
User Andras Farkas <deepbluemistake@gmail.com> brought this issue to our
attention via a message on the ksh mailing list. There are two problems.
First, autoloaded functions `cd` and `mcd` aren't available until the
user uses the `dirs` command. So put each function definition in its own
autoloaded file. Second, the `_cd` function behaves incorrectly if you
invoke it with no arguments.

I'm ambivalent about the `_cd` behavior of always printing the selected
directory on stdout. Personally I think that is just noise. If the user
wants to be reminded what the CWD is they can put it in their prompt.
But I'm keeping that behavior for the moment to see how others feel and
because that is what the legacy implementation does which implies some
people think the behavior is useful.

### DIFF
--- a/src/cmd/ksh93/data/config.ksh
+++ b/src/cmd/ksh93/data/config.ksh
@@ -4,6 +4,8 @@
 # for them to be loaded on demand. It should not do anything not suitable for every single new ksh
 # process. In particular it should be as fast as possible.
 #
+
+#
 # Arrange for standard autoloaded functions to be available. The test for whether or not FPATH is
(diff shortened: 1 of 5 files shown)


krader1961:
The `mcd` function is extremely primitive. It reminds me of the first version of that function I wrote for myself 30 years ago. I would like  to write, or see someone else write,  a version of that function much more like the `cdh` "fish" shell function I wrote.

PR #1424, opened 2019-11-02
Add missing declaration for argv_addrs in debug.c

DanySpin97:
I've added the same declaration that was in other parts of the file.

### DIFF
--- a/src/lib/libast/misc/debug.c
+++ b/src/lib/libast/misc/debug.c
@@ -329,6 +329,9 @@ static_fn char **addrs2info(int n_frames, void *addrs[]) {
     UNUSED(n_frames);
     UNUSED(addrs);
 
+    // Sixteen hex digits is enough for 64 bit addrs. The extra three chars are for the `0x` prefix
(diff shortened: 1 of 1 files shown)


krader1961:
Thanks, @DanySpin97, for reporting this issue and a proposed fix. This is a great example for why I dislike code predicated on preprocessor symbols; notwithstanding that I wrote the code in question. Such code tends to suffer bit rot. Your fix isn't correct but I've merged commit 11983a71, credited to you, with the proper fix since you brought it to my attention.

I'm curious why you noticed this broken code. Are you testing on a platform that does not have the `atos` or `addr2line` programs? Or simply testing on a platform where they are available but not installed?

DanySpin97:
> Your fix isn't correct but I've merged commit 11983a7, credited to you, with the proper fix since you brought it to my attention.

Thank you!

> I'm curious why you noticed this broken code. Are you testing on a platform that does not have the atos or addr2line programs? Or simply testing on a platform where they are available but not installed?

Due to an error in the binutils WIP package on my system, addr2line was not installed (and I've just discovered this error).

PR #1426, opened 2019-11-03
Fix for test failures when home directory is not available

aweeraman:
I've noticed that in some build environments (such as the Debian CI CD infra) that uses docker, the home directory is not available, and so some tests fail. I've been able to recreate this locally when using chroot-based environments for validating Debian package builds. In such cases 'cd' throws an error as there's no home directory to cd to, however, I think this is the desired behavior, so I've added a check for this in the tests to stop the tests from failing in those cases and show a warning. Please review this approach and see if this makes sense or if you see any issues with this. I've tested this on Linux with and without a valid home directory and tests pass. Thanks.

### DIFF
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -476,14 +476,15 @@ $SHELL <<- \EOF
     cd
     [[ $(pwd) == "$home" ]]
 EOF
-ret=$?
-if [[ ! -d ~$USER ]]
+if [[ $? != 0 ]]
 then
-    log_warning "skipping test of cd with no arguments if home directory does not exist"
-else
-    [[ $ret == 0 ]] || log_error 'cd with no arguments fails if HOME is unset'
+    if [[ -d ~$USER ]]
+    then
+        log_error 'cd with no arguments fails if HOME is unset'
+    else
+        log_warning "ignoring test of cd with no arguments if home directory does not exist"
+    fi
 fi
-unset ret
 
 cd "$TEST_DIR"
 if mkdir -p f1
--- a/src/cmd/ksh93/tests/builtins.sh
+++ b/src/cmd/ksh93/tests/builtins.sh
@@ -476,15 +476,14 @@ $SHELL <<- \EOF
     cd
     [[ $(pwd) == "$home" ]]
 EOF
-if [[ $? != 0 ]]
+ret=$?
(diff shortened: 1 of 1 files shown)


src/cmd/ksh93/tests/builtins.sh
```diff
@@ -476,7 +476,15 @@ $SHELL <<- \EOF
     cd
     [[ $(pwd) == "$home" ]]
 EOF
-[[ $? == 0 ]] || log_error 'cd with no arguments fails if HOME is unset'
+if [[ $? != 0 ]]
+then
+    if [[ ! -d $home ]]
```
krader1961: This is wrong. The `home` var isn't visible in this scope since it was defined in a subshell. Also, we generally want to skip a test if we know it is going to fail. Rather than deal with a failure by deciding if the test should have been skipped. I can't think of any other tests where we use this pattern.

I wonder if the real issue here is that `$USER` is undefined in your test environment. It's oftentimes useful to add a `set -x` at the top of the test script.

aweeraman: Yes, you're right, home shouldn't be visible here, let me rework this.

krader1961: I would like to see the output of `log_info "HOME=|$HOME| USER=|$USER|"` immediately prior to the failing test

aweeraman: <I> builtins[517]: HOME=|/tmp/ksh.builtins.Eflf2W/home| USER=|anuradha|

src/cmd/ksh93/tests/builtins.sh
```diff
@@ -476,15 +476,14 @@ $SHELL <<- \EOF
     cd
     [[ $(pwd) == "$home" ]]
 EOF
-if [[ $? != 0 ]]
+ret=$?
+if [[ ! -d ~$USER ]]
```
krader1961: Better but please invert the logic; i.e., only do this test if the previous command failed since there isn't any point in doing this if it succeeded. For example,

```
if [[ $? != 0 ]]
then
    if [[ -d ~$USER ]]
    then
        log_error 'cd with no arguments fails if HOME is unset'
    else
        log_warning "ignoring test of cd with no arguments if home directory does not exist"
    fi
fi
```

krader1961:
Squash merged. Thx.

PR #1431, opened 2019-11-05
Fix functionality of printf %H

kshfan:
The man page says %H is meant to produce text valid for use in both HTML and XML.  The current behavior is both invalid for HTML and invalid for XML. &nbsp; is invalid in XML.  &apos; is invalid in HTML (not including HTML5)
Thus, I eliminate the invalid entities here, making it so that the output of printf %H is valid in both HTML and XML. This is backwards-incompatible as for ksh, but it has to be, since it was implemented wrong to begin with.  Fortunately, this doesn't affect compatibility with POSIX printf. ;D

%#H is already in use for making valid URIs...  So I don't know what to do to make a %something which will also include &apos; for XML.  What flag could we use for printf to print in a more XML-oriented way? i.e. to be the same as this functionality post-diff, but with &apos; added back.

I've tested the behavior of ksh itself, but haven't tested the tests.  Please test the tests for me.

### DIFF
--- a/src/cmd/ksh93/bltins/print.c
+++ b/src/cmd/ksh93/bltins/print.c
@@ -431,12 +431,11 @@ static_fn char *fmthtml(Shell_t *shp, const char *string, int flags) {
                 sfputr(shp->stk, "&amp", ';');
             } else if (c == '"') {
                 sfputr(shp->stk, "&quot", ';');
+            /* Add this back when an XML-oriented output flag is added. */
+            /*
             } else if (c == '\'') {
                 sfputr(shp->stk, "&apos", ';');
-            } else if (c == ' ') {
-                sfputr(shp->stk, "&nbsp", ';');
-            } else if (!isprint(c) && c != '\n' && c != '\r') {
-                sfprintf(shp->stk, "&#%X;", c);
+            */
             } else {
                 sfputc(shp->stk, c);
             }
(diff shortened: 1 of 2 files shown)


kshfan:
Github changed my pull request text by eating the html/xml entities, so read the commit text instead.
Edit:
Specifically, read this:
https://github.com/att/ast/pull/1431/commits/309e759f67ec7a62527ec4d24004bec99117f606

krader1961:
Thanks for contributing, @kshfan. But I'm uncomfortable merging this PR. It's not clear your change is a net improvement in behavior. Please [open an issue](https://github.com/att/ast/issues/new/choose) including as much information as possible including links to the relevant standards and examples of how ksh currently behaves that you believe is incorrect.

kshfan:
@krader1961 
Thank you for the advice.  I made the issue:
https://github.com/att/ast/issues/1435

kshfan:
Can this PR be re-opened?

kshfan:
That way, more people might look at it and review it, since the verdict on this wasn't final yet.

krader1961:
If we knew this change was correct the statements would simply be deleted rather than hidden by `#if 0...#endif`.  What I recommend is adding a comment to the associated issue that has a permalink to the lines you believe should be removed.

kshfan:
@krader1961 
I don't know how to do that.

(also, only apos was commented out in my diff.  some lines were genuinely deleted)

krader1961:
https://help.github.com/en/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet

It's a great way to embed short blocks of code in Github issues. In this case link to the code you think should be removed and include a comment to that effect.

kshfan:
@krader1961 
Thanks :D

Issue #1449, opened 2019-11-22
serious performance breakdown of ksh2020 in comparison to ksh93u+


jghub:
**Description of problem:**

consider a stupid micro benchmark like
`
function repeatloop {
  nameref count=$1
  integer i=0 buf=0
  for ((i=0; i< $count; i++)); do
     ((buf++))
     :
  done
  print $buf
}

count=1000000
time repeatloop count
`

**Ksh version:**
on my machine (OSX 10.4.6, powerbook)

93u+:  1.01s
2020 release: 2.12s
head (42a580c5): 2.98s

**How reproducible:**
always

**Steps to reproduce:**
 run above script

**Actual results:**

100-200% slow down (i.e. by a factor 2-3!) of ksh2020 in comparison to 93u+. although ksh2020 thus remains faster than bash (11s) the "order of magnitude faster than bash" is no more . if this is not fixed one might consider to copy this line from the bash manpage: "It's too big and too slow."

I find a performance breakdown by such a massive factor disconcerting. it is a big step back from ksh93u+.

**Expected results:**

competitive performance

**Additional info:**

do people see this behaviour on other platforms, too? 

KornShell language is not only a command interpreter for interactive use. it also is a serious programming language. performance matters a *lot*.  ksh always had by far the best performance of the major shells.  current development in this ksh2020 project seems to severely damage ksh in this respect. I hope this can be fixed. are there performance benchmarks in the test suite at all?

krader1961:
> consider a stupid micro benchmark

LOL! Yes, a lot of micro benchmarks are stupid in as much as they are irrelevant with respect to any real world program or shell script. Having said that, I am disturbed by the factor of three change in run time that I am also seeing on my primary platform.

> KornShell language is not only a command interpreter for interactive use. it also is a serious programming language.

I disagree, notwithstanding being a fan of ksh as the first UNIX shell I considered usable. The korn shell can not be a "serious programming" language due to the POSIX shell standard. Having said that I agree this slow down is not acceptable.

kshfan:
> The korn shell can not be a "serious programming" language due to the POSIX shell standard.

Weird.  I use it as a serious programming language all the time.

jghub:
[...]

PR #1463, opened 2020-01-29
Change default build type to "minsize"

krader1961:
People (and distros) have been building ksh from this project using the
default "debug" build type since we switched from Nmake to Meson. That
results in a ksh program whose performance is far from optimal. So change
the default build type. If a debugging enabled binary is needed, such as
in a CI environment, it must be requested explicitly.

Related #1449

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,9 @@
-project('ksh93', 'c', default_options: [
+project('ksh9', 'c', default_options: [
+    # This is the optimal production build type for this project. If you need
+    # a debug build you must explicitly run `meson --buildtype=debug`. This
+    # helps ensure that each distro (or user) gets a performance optimized build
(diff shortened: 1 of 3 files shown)


siteshwar:
LGTM

krader1961:
Merged with a comment added to *CHANGELOG.md*. I also cherry-picked this into the 2020.0.1 branch so that future builds by distros will be appropriately optimized even if they don't explicitly use the `--buildtype=minsize` option.

meson.build
```diff
@@ -1,4 +1,9 @@
-project('ksh93', 'c', default_options: [
+project('ksh9', 'c', default_options: [
```
aweeraman: There seems to be a typo here on "ksh9"...

aweeraman: Just noticed that this is already fixed on trunk

PR #1482, opened 2020-06-30
change for gcc v10

lkoutsofios:
change to ps.c to remote pid size check
change to sfputr.c to disable use of memccpy
change to cmdarg.c to adjust argument length calc. due to failures in tw


kdudka:
I find it disturbing that you add line comments while changing code base that otherwise does not use them.

PR #1492, opened 2021-08-15
Repair support for UnixWare 7.x, preliminary modern Mac OS X support

lkujaw:
* src/cmd/ss/Makefile: Properly handle sys ownership of /dev/kmem.
* src/cmd/ss/Mamfile: Regenerate Mamfile.
* src/lib/libast/features/standards:
- Add heuristic (u_long availability) for systems that hide rather
  than reveal functionality in the presence of _POSIX_SOURCE, etc.
- Define _DARWIN_C_SOURCE, like _GNU_SOURCE, to enable the full range
  of definitions on Mac OS X systems.
- Due to the above, remove __MACH__ (Mac OS X)-specific hack.

### DIFF
--- a/src/cmd/ss/Makefile
+++ b/src/cmd/ss/Makefile
@@ -4,7 +4,19 @@ LICENSE = since=1990,author=gsf
 
 ss :: RELEASE ss.c -lcs
 
-$(LIBDIR) :INSTALLDIR: group=kmem mode=ug+s,+x preserve=1 ssd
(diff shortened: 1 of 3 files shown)


McDutchie:
Hi, I'm not sure why you bothered to post a PR here as this repo is inactive and AT&T will ignore it, but I'm glad you did, because the libast bit of this is applicable to [ksh 93u+m](/ksh93/ksh). :-)

Mliveon (approved):

PR #1506, opened 2025-04-10
Fix FILES section in the sh.1 manual

rjc:


McDutchie:
This AT&T repo has been inactive for a long time. Since ksh93 development continues at https://github.com/ksh93/ksh, please submit any PRs there. Although this particular fix was committed there back in 2020 (see https://github.com/ksh93/ksh/commit/c1c7a1bcaf325d35a047a2d132ac7fa595d28145).

rjc:
Hi @McDutchie,

Yes, I'm aware of that. Unfortunately, macOS still seems to be using this, instead of the `ksh93` repository and, subsequently, the bug is still present in the manual page there.

Also, given that there were still commits here in 2020, and the fact that the project is not archived, I assumed it could be fixed here, too.

Despite publishing the repositories they are using, Apple make it difficult to contribute - only PRs from contributors are allowed. Oh well, I guess I'll just have to live with it.

Ta!

